PR #7, opened 2021-04-16
Docs

hboisgon:
Migrate the core hydromt documentation


DirkEilander (approved):
The docs look good and build nicely at my laptop.
A small suggestion:  We should perhaps update some figures (e.g. introduction and sfincs before v04)

PR #26, opened 2021-05-07
Fiat devs

DirkEilander:
* add {'grid': path/to/raster.tif} to region options
* initiate placeholders earlier in model initialization
* improve docstring of build & update
* don't for epsg conversion for model.region and raster.box as not all CRS have an EPSG code.


hboisgon (approved):
Nice developments with a new 'grid' region options. 
Also good to have better docstrings for some important model api methods.

PR #53, opened 2021-07-14
improve raster mosaic

DirkEilander:
- mask option in merge.merge method for improved open_raster_from_tindex. The method reads and merges data from tiles with different projections summarized in a gdal tindex geometry within a bounding box. The destination crs and resolution are now determined based on the tile with the largest overlap with the destination bounding box, which basically keeps most of the data as is. 
- dropped dask version pins
- remove IOError in set_config (IOErrors should not be in set_ methods, only write_ methods)

### DIFF
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -9,7 +9,7 @@ dependencies:
   - black
   - bottleneck
   - cartopy # to run examples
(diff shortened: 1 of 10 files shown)


hboisgon (commented):
Looks good!
Nice that we are rid of the dask pin. It also should be removed from binder/environment.yml.

PR #55, opened 2021-07-21
Datatypes

xldeltares:
add as type and rearrange dataadaptor get_data function orders.

### DIFF
--- a/hydromt/data_adapter.py
+++ b/hydromt/data_adapter.py
@@ -1468,7 +1468,7 @@ class GeoDataFrameAdapter(DataAdapter):
         kwargs = self.kwargs.copy()
         _ = self.resolve_paths()  # throw nice error if data not found
 
-        #TODO: Ask Helene: could the order of functions be:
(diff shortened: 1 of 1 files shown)


DirkEilander:
@xldeltares I was wondering if this PR is still relevant and should be finalized or can be closed. If it needs to be finalized can you let me know what still needs to be done (and possibly when you could do this)?

xldeltares:
> @xldeltares I was wondering if this PR is still relevant and should be finalized or can be closed. If it needs to be finalized can you let me know what still needs to be done (and possibly when you could do this)?

Hi @DirkEilander! I don't think it is relevant anymore. The current data adaptor seems to be sufficient for the current development of the delft3dfm plugin (thanks to the sophisticated validation rules in hydrolib-core. Also for some special cases we deal with it in the code). So if no other plugins have such a request, I would agree to close it. :)

DirkEilander:
Thanks for the feedback @xldeltares. Will close for now. If there are new usecases let us know.

PR #58, opened 2021-07-23
bugfix get_geodataset with vector driver

DirkEilander:
- Fix passing timeseries and crs for get_geodataset with vector driver
- read-only check in write_config; dropped write_results
- added instersphinx mapping

The changes in `io.open_geodaset` and `data_adapter.DataCatalog.get_geodataset` only effect hydromt_sfincs (I couldn't find any usages in other plugins) and relates to issue https://github.com/Deltares/hydromt_sfincs/issues/27

To read a geodataset from a location and timeseries text (csv) files one can now use the following yml syntax:
```
waterlevels_gtsr:
  path: gtsr_stations.csv
  data_type: GeoDataset
  driver: vector
  crs: 4326
  kwargs:
    fn_data: gtsr_stations_data.csv
```


DirkEilander:
@hboisgon could you review this PR? 

I suggest to also make a new release (v0.4.2) if you agree? 
This version is required to fix an upstream issue in hydromt_sfincs and to be able to use new dask releases.

hboisgon (approved):
Good fix on open_geodataset and editing the tests.
I agree to remove the write_results method, it does not really make sense to do so.

I realize the newest states and results object have not been added to the docs. Something to do in this PR? Might be good to improve the docs on the model API

hboisgon:
I also agree for the new release!
Something I forgot, but is there any examples in the docs of the data adapter for such geodataset?
The example you mention above would be good to have there.

DirkEilander:
Thanks for the quick review! 
Will update the docs and ask you to quickly review that too before finalizing and merging the PR.

DirkEilander:
@hboisgon I think this is ready to be merged and released. 

I've restructured the data page in the docs using examples for various data formats and added intersphinx mapping to be able to refer to methods from e.g. xarray and geopandas.

Could you do a last check. And would you also have time to do the release?

PR #61, opened 2021-08-02
cli data_libs fix

DirkEilander:
* fixes use of  single path without list notation in data_libs in config [global] section. To behaviour is now also similar to --deltares-data which overwrites the setting in the ini file.
* small fix of erroneous version number. This should solve the issue with the wrong folder name in the gh-pages branch.

### DIFF
--- a/hydromt/__init__.py
+++ b/hydromt/__init__.py
@@ -1,6 +1,7 @@
 """HydroMT: Build and analyze models like a data-wizard!"""
 
-__version__ = "v0.4.3.dev"
+# version number without 'v' at start
+__version__ = "0.4.3.dev"
 
 import geopandas as gpd
 import warnings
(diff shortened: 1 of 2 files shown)


DirkEilander:
@hboisgon Just a very small PR. If all you agree it can be merged with main and the branch deleted.

DirkEilander:
after discussion with @hboisgon 

The following config notations with multiple and single files should both work. We need to double check if the first notation still works with this PR.

```
[global]
data_libs = ['relative/path/to/some_file.yml', '/absolute/path/to/another/file.yml']
```

```
[global]
data_libs = 'relative/path/to/some_file.yml'
```

DirkEilander:
I've checked and both notations below work now. 

> 
> ```
> [global]
> data_libs = ['relative/path/to/some_file.yml', '/absolute/path/to/another/file.yml']
> ```
> 
> ```
> [global]
> data_libs = 'relative/path/to/some_file.yml'
> ```

PS Would be good to create a dummy model in the conftest.py and add this to the entrypoints (using a monkeypatch?) such that it called from command line. That would make it possible to extent the tests on the cli interface and check if this behavior.

hboisgon (approved):
Thanks for the checks! And agree to give priority to the CLI instead of a mix of cli+config for data_libs options

PR #75, opened 2021-10-11
Basin

hboisgon:
Bugfix for basin_mask.get_basin_geometry method.

When using basin_index to do a read of the hydrography data and get the basin out of region, not all the very small basins are included in the basin_index data leading to a cut of very small basins that are still present in the user defined region.


DirkEilander:
thanks for this fix @hboisgon. On the long term it would be best to replace the current basin index with a pfafstetter based one where small coastal basins are merged with larger nearby basins, For now this is an important fix!

PR #79, opened 2021-11-11
Rivers

DirkEilander:
- gis_utils.nearest and gis_utils.nearest_merge methods to merge GeoDataFrame based on proximity
- river_width to estimate a segment average river width based on a river mask raster 
- river_depth to get segment average river depth estimates based bankfull discharge (requires pyflwdir v0.5.2 - to be realeased)
- flw.floodplain_elevation method which returns floodplain classification and hydrologically adjusted elevation

### DIFF
--- a/docs/api/api_methods.rst
+++ b/docs/api/api_methods.rst
@@ -251,6 +251,7 @@ visit the `pyflwdir docs. <https://deltares.github.io/pyflwdir/latest/>`_
    flw.gauge_map
    flw.outlet_map
    flw.clip_basins
+   flw.floodplain_elevation
 
 
 General GIS methods
--- a/docs/api/api_workflows.rst
+++ b/docs/api/api_workflows.rst
@@ -23,6 +23,16 @@ Basin mask
    workflows.basin_mask.get_basin_geometry
    workflows.basin_mask.parse_region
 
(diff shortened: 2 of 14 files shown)


hboisgon:
Nice changes! Is the idea that in the end we fully move the river.py workflow out of hydromt_wflow into hydromt core? If maybe good to transfer the functions now already?

hboisgon (approved):
Nice and useful to get river depth estimations! 
Very minor but could you also list the new functions in the docs? And add the new geopandas min version also in the environment.yml files? Just so that they all match.
Thanks a lot!

DirkEilander:
Thanks for the review @hboisgon. 

I've made the changes you requested to the docs and env files. Also bumped pyflwdir which was released on conda-forge this morning and is required for the flwdir.river_depth method.

I've put the general parts of the river methods here. Both hydromt_wflow and hydromt_sfincs will have slightly different implementations.

hboisgon:
Perfect! Sounds good to go then @DirkEilander

DirkEilander:
@hboisgon Added one more method in in hydromt.flw which requires a release of pyflwdir. Will merge after that release. Let me know if you have suggestions on that method.

PR #81, opened 2021-11-25
Add more statistics

hboisgon:
Add more efficiency coefficients in stats.py (eg kge_np)


DirkEilander:
Looks good. Just missing a brief description of the statistic in the docstring, specifically how the a specific KGE is different from the default KGE, now they say more or less the same.

hydromt/stats.py
```diff
@@ -187,6 +187,80 @@ def spearman_rank_correlation(sim, obs, dim="time"):
...
+
+    Returns
+    -------
+    xarray DataSet
+        Non Parametric Kling-Gupta Efficiency (2018) optimize for flood peaks
+        using pearson (see Pool et all., 2018) and with decomposed score
+    """
+    # cc = spearman_rank_correlation(sim, obs)
+    cc = pearson_correlation(sim, obs, dim=dim)
+    cc.name = "kge_pearson_coef"
```
DirkEilander, line 250: for consistent naming I suggest
`cc.name = "kge_np_flood_pearson_coef"`

hydromt/stats.py
```diff
@@ -187,6 +187,80 @@ def spearman_rank_correlation(sim, obs, dim="time"):
...
+    dim : str, optional
+        name of time dimension in sim and obs (the default is 'time')
+
+    Returns
+    -------
+    xarray DataSet
+        Non Parametric Kling-Gupta Efficiency (2018) and with decomposed score
+    """
+    cc = spearman_rank_correlation(sim, obs, dim=dim)
+    cc.name = "spearman_rank_correlation_coef"
```
DirkEilander, line 212: for consistent naming I suggest
cc.name = "kge_np_spearman_rank_coef"

hydromt/stats.py
```diff
@@ -295,12 +369,51 @@ def kge(sim, obs, dim="time"):
...
+        name of time dimension in sim and obs (the default is 'time')
+
+    Returns
+    -------
+    xarray DataSet
+        Kling-Gupta Efficiency (2012) and with decomposed scores
+    """
+    cc = pearson_correlation(sim, obs, dim=dim)
+    cc.name = "kge_2012_pearson_coef"
+    alpha = sim.std(dim=dim) / obs.std(dim=dim)
```
DirkEilander, line 402: I think these lines are redundant as alpha is calculated again later

```
alpha = sim.std(dim=dim) / obs.std(dim=dim)
alpha.name = "kge_2012_rel_var"
```

hydromt/stats.py
```diff
@@ -187,6 +187,80 @@ def spearman_rank_correlation(sim, obs, dim="time"):
...
+        simulations time series
+    obs : xarray DataArray
+        observations time series
+    dim : str, optional
+        name of time dimension in sim and obs (the default is 'time')
+
+    Returns
+    -------
+    xarray DataSet
+        Non Parametric Kling-Gupta Efficiency (2018) optimize for flood peaks
```
DirkEilander, line 245: I suggest moving this explanation to the start of the docstring

hydromt/stats.py
```diff
@@ -187,6 +187,80 @@ def spearman_rank_correlation(sim, obs, dim="time"):
...
+        simulations time series
+    obs : xarray DataArray
+        observations time series
+    dim : str, optional
+        name of time dimension in sim and obs (the default is 'time')
+
+    Returns
+    -------
+    xarray DataSet
+        Non Parametric Kling-Gupta Efficiency (2018) and with decomposed score
```
DirkEilander, line 209: I suggest moving this explanation to the start of the docstring

hydromt/stats.py
```diff
@@ -295,12 +369,51 @@ def kge(sim, obs, dim="time"):
...
+        simulations time series
+    obs : xarray DataArray
+        observations time series
+    dim : str, optional
+        name of time dimension in sim and obs (the default is 'time')
+
+    Returns
+    -------
+    xarray DataSet
+        Kling-Gupta Efficiency (2012) and with decomposed scores
```
DirkEilander, line 398: I suggest moving this explanation to the start of the docstring

tests/conftest.py
```diff
@@ -113,3 +113,29 @@ def hydds(flwda, flwdir):
     )
     ds.raster.set_crs(flwda.raster.crs)
     return ds
+
+
+@pytest.fixture
```
DirkEilander, line 118: I think we could only do with a single normalized time series instead of `obsda` en `simda`. Then in the tests one can create a second timeseries based on the first. Which also makes it directly clear how the two are related and thus what the statistic should be.

tests/test_stats.py
```diff
@@ -0,0 +1,27 @@
...
+
+import pytest
+import numpy as np
+import xarray as xr
+
+from hydromt import stats
+
+
+def test_geo(obsda, simda):
+    assert stats.bias(simda, obsda).values.round(2) == 5.0
```
DirkEilander, line 12: For these type of statistics two decimals is not a lot for comparison, small differences will not be noted. 
I suggest to use np.isclose with much more decimals or round with at least 5 decimals.

DirkEilander (changes requested):
I've left a few minor suggestions on the  code. 

Furthermore I'd like to move this script in a "stats" folder and call this script skill.py (or similar) to be able to add more stats (e.g. extreme value stats) in the near future. 
And finally we need to update the changelog.

DirkEilander (approved):
looks great!

PR #83, opened 2021-12-02
import rioxarray functionality

### LINKED ISSUE #82: xarray.open_rasterio will be deprecated
we will need to use rioxarray instead

DirkEilander:
Start with replacing some of the raster functionality with rioxarray functionality. 
The replacement of xr.open_rasterio is urgent as it will be deprecated. 
Other functionality should slowly also be ported to rioxarray functionality if possible. 
Also in the plugins I would advise to start import rioxarray directly and use this as much as possible.

- fixes #82


DirkEilander:
I think this step is ready to merge. Can you review the implementation @hboisgon? We can continue including more rioxarray functionality in future PRs.

DirkEilander:
@hboisgon Could you also review this branch? Perhaps test it with hydromt_delwaq? After which I suggest to make a new hydroMT release.

hboisgon:
I took a long time with reviewing so maybe that's why it seems there has been quite some changes... but the tests for hydromt_wflow were failing in a lot of different parts...

Seems that rioxarray became stricter on using a valid nodata value depending on the dataaaray dtype. While previously xarray would convert the dtype on the fly to match the new data value, rioxarray just does nothing...
I think we maybe should also be more strict about this in hydromt and use more common nodata values for each expected dtype. Especially in wflow, nodata values seem to be a bit arbitrary sometimes.

Maybe we need more time to review this. I'll check also with hydromt_delwaq but good to re-run the hydromt_sphincs tests as well.

DirkEilander:
TODO:
- fix #117

hboisgon (approved):

hboisgon:
@DirkEilander I think this is now ready to merge! Could you check the new gdal_compliant raster method and if you agree, merge the branch?

PR #89, opened 2022-01-19
rivers

laurenebouaziz:
Just made a branch with only one super tiny thing, but I wanted to add a few other small suggestions on docstring


DirkEilander:
Thanks for the suggestions @laurenebouaziz 

- I've added the  references and tried to clarified the docstrings.
- The distributed manning values based on stream order is actually something we do in `WflowModel.setup_rivers` and is used in `hydromt.workflows.river_depth.`

PR #90, opened 2022-01-20
general setup_region method

### LINKED ISSUE #73: Basin ID option in the docs
In the region docs, the example with correct syntax to build a model based on basin ID is missing.

### LINKED ISSUE #88: general Model.setup_region method
To open the door for more models which don't have a clear basemap component (eg network models) it would be good to make general `setup_region` method that takes the `region` argument and parses this into a geometry. The `Model.build` method will need to be relaxed to be able to pass the `region` argument, at least for the time being, to either `setup_region` or `setup_basemap` depending on the model implementation (similar for the `res` argument). 

In SFINCS we also have use cases for setting up the region based on eg MERIT Hydro, but than use a different (local) DEM for the sfincs.dep file, which is another argument to separate these methods, see https://github.com/Deltares/hydromt_sfincs/issues/44

DirkEilander:
- [x] relax Model.build method requirements for a setup_basemaps method
- [x] implement a setup_region method to parse the region CLI argument and return a region geometry, fixes #88
- [x] improve interbasin regions
- [x] improve region documentation, fixes #73

### DIFF
--- a/hydromt/flw.py
+++ b/hydromt/flw.py
@@ -257,7 +257,9 @@ def reproject_hydrography_like(
     elevation is used assuming these elevation values are <= 0 (i.e. offshore bathymetry).
 
     The upstream area on the reprojected grid is based on the new flow directions and
-    rivers entering the domain, defined by the minimum upstream area `river_upa` [km2].
+    rivers entering the domain, defined by the minimum upstream area `river_upa` [km2]
+    and a distance from river outlets `river_len` [m]. The latter is to avoid setting
+    boundary conditions at the downstream end / outflow of a river.
 
     NOTE: the resolution of `ds_hydro` should be similar or smaller than the resolution
     of `da_elv` for good results.
--- a/hydromt/flw.py
+++ b/hydromt/flw.py
@@ -324,7 +324,8 @@ def reproject_hydrography_like(
     if not kwargs:
         _msk = np.logical_and(_edge, elvsyn_reproj <= 0)
         _msk = np.logical_or(_msk, elvsyn_reproj == 0)
-        kwargs.update(idxs_pit=np.where(_msk.values.ravel())[0])
+        if np.any(_msk):  # False if all pits outside domain
+            kwargs.update(idxs_pit=np.where(_msk.values.ravel())[0])
     logger.info(f"Deriving flow direction from reprojected synthethic elevation.")
     da_flw1 = d8_from_dem(elvsyn_reproj, **kwargs)
     flwdir = flwdir_from_da(da_flw1, ftype="d8", mask=elv_mask)
--- a/hydromt/flw.py
+++ b/hydromt/flw.py
@@ -244,6 +244,7 @@ def reproject_hydrography_like(
     ds_hydro: xr.Dataset,
     da_elv: xr.DataArray,
     river_upa: float = 5.0,
+    river_len: float = 1e3,
     uparea_name: str = "uparea",
     flwdir_name: str = "flwdir",
     logger=logger,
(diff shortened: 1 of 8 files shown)


DirkEilander:
@hboisgon Could you let me know if you have time to review this? It's an important update to build SFINCS models in the FEWS FLiFS project in Australia. I will try to improve the test coverage in a follow up PR where we implement a dummy model class.

hydromt/flw.py
```diff
@@ -270,7 +271,9 @@ def reproject_hydrography_like(
     da_elv: xarray.DataArray
         DataArray with elevation on destination grid.
     river_upa: float, optional
-        Minimum upstream area threshold for inflowing rivers, by default 5 [km2]
+        Minimum upstream area threshold [km2] for inflowing rivers, by default 5 km2
+    river_len: float, optional
```
hboisgon, line 275: Went a bit deeper into the function this time and very nice one! 
It's quite complex and a lot of good things and exceptions are considered but a bit tough to understand what happens by just reading the docstring explnantion (maybe a bit more step by step?). The new addition of river_len is missing from the explanation.

hydromt/models/model_api.py
```diff
@@ -188,9 +197,7 @@ def update(self, model_out=None, write=True, opt=None):
 
         # check if model has a region
         if self.region is None:
-            raise ValueError(
-                'Model region not found, setup basemaps using "build" method first.'
-            )
+            raise ValueError("Model region not found, setup model using `build` first.")
 
         # remove setup_basemaps from options and throw warning
         if "setup_basemaps" in opt:
```
hboisgon, line 203: Shall we replace by setup_region ? Or both setup_basemaps+setup_region?

hydromt/models/model_api.py
```diff
@@ -211,6 +218,88 @@ def update(self, model_out=None, write=True, opt=None):
         if write:
             self.write()
 
+    ## general setup methods
+
+    def setup_region(
```
hboisgon, line 223: Super nice! Should replace the equivalent use in the different plugins (with a check on kind afterwards)

hydromt/models/model_api.py
```diff
@@ -32,6 +32,8 @@ class Model(object, metaclass=ABCMeta):
     _GEOMS = {"<general_hydromt_name>": "<model_name>"}
     _MAPS = {"<general_hydromt_name>": "<model_name>"}
     _FOLDERS = [""]
+    # tell hydroMT which methods should receive the res and region arguments
+    _CLI_ARGS = {"region": "setup_basemaps", "res": "setup_basemaps"}
```
hboisgon, line 36: I think this is good to avoid too much backward compatibility with existing plugins like wflow or sfincs etc.
But if we make a release with this change anyway, shouldn't we try to move away from setup_basemaps in the model api and here use setup_region instead ? res can be removed or stay with setup_basemaps for gridded models. I guess we may get new CLI args when the new model type come in (network, mesh).

Maybe in the end keep like this for now and discuss again during the design network/mesh sessions

hboisgon (approved):
Very nice PR @DirkEilander :)
I like the new way to pass the region and res options!
The new setup_region method is also good and we should use it in the already existing plugins (will avoid some code duplicates). I see no testing yet but I assume this is for when we work on replacing the DummyModel by a better TestModel class.

hydromt/models/model_api.py
DirkEilander: good one

hydromt/flw.py
DirkEilander: will add

hydromt/models/model_api.py
DirkEilander: yes! For hydromt_sfincs I'm already using it even without wrapping. For hydromt_wflow we could wrap it in the setup_basemaps function or split it

hydromt/models/model_api.py
DirkEilander: I indeed used "setup_basemaps" here to make sure all plugins keep running for now. We can decide again later to change this if that makes sense then.

DirkEilander: the "res" keyword is optional btw. If it's not in this dictionary it will not be used in any method.

PR #93, opened 2022-01-31
Climate wf

laurenebouaziz:
adding placeholders in the yml file now gives the opportunity to create multiple yml entries from a single entry, as suggested in the dicussion of 5171c3d2e2cb335cee9421d94a28418c0cb0fa81

### DIFF
--- a/hydromt/data_adapter.py
+++ b/hydromt/data_adapter.py
@@ -9,6 +9,7 @@ from os.path import join, isdir, dirname, basename, isfile, abspath
 from itertools import product
 import copy
 from pathlib import Path
+from unittest.mock import NonCallableMagicMock
 import numpy as np
 import xarray as xr
 import geopandas as gpd
@@ -392,6 +393,9 @@ class DataCatalog(object):
         buffer=0,
         align=None,
         variables=None,
+        models=None,
+        scenarios=None,
+        realizations=None,
         time_tuple=None,
         single_var_as_array=True,
         **kwargs,
@@ -403,6 +407,7 @@ class DataCatalog(object):
         To slice the data to the time period of interest, provide the `time_tuple` argument.
         To return only the dataset variables of interest and check their presence,
         provide the `variables` argument.
+
 
         NOTE: Unless `single_var_as_array` is set to False a single-varaible data source
         will be returned as :py:class:`xarray.DataArray` rather than :py:class:`xarray.Dataset`.
@@ -423,6 +428,12 @@ class DataCatalog(object):
         variables : list of str, optional.
             Names of RasterDataset variables to return. By default all dataset variables
             are returned.
+        models : list of a single str, optional.
+            Name of placeholder {model} in path of data catalog key.
+        scenarios : list of a single str, optional.
+            Name of placeholder {scenario} in path of data catalog key.
+        realizations : list of a single str, optional.
+            Name of placeholder {realization} in path of data catalog key.
         time_tuple : tuple of str, datetime, optional
             Start and end date of period of interest. By default the entire time period
             of the dataset is returned.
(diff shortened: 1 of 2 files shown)


laurenebouaziz:
Here is a copy of the yml entry that I used for testing:

```
root: p:/i1000365-007-blue-earth/ClimateChange
isimip3_{model}_{scenario}_{realization}:
  crs: 4326
  data_type: RasterDataset
  driver: netcdf
  kwargs:
    chunks:
      lat: 125
      lon: 120
      time: 50
    decode_times: true
    parallel: true
  meta:
    category: climate
    paper_doi: todo
    paper_ref: todo
    source_license: todo
    source_url: todo
    source_version: 1.0  
  placeholders:
    model: [gfdl, ipsl]
    scenario: [ssp126, historical]
    realization: [r1i1p1f1]
  path: ISIMIP3/{model}/{scenario}/{variable}/{model}*{realization}*{scenario}*{variable}_global_daily*.nc
  rename:
    pr: precip
    tas: temp
  unit_add:
    temp: -273.15
  unit_mult:
    precip: 86400
```

hydromt/data_adapter.py
```diff
@@ -711,6 +721,8 @@ def __init__(
         self.unit_add = unit_add
         # meta data
         self.meta = {k: v for k, v in meta.items() if v is not None}
+        # placeholders
+        self.placeholders = {k: v for k, v in placeholders.items() if v is not None}
```
DirkEilander, line 725: @laurenebouaziz: Is there a specific reason why you added this line? I think it can be removed.

tests/test_data_adapter.py
```diff
@@ -9,11 +9,71 @@
...
+    dd = {
+        "test": {
+            "data_type": "RasterDataset",
+            "path": "path/to/data.tif",
+        },
+        "test1": {"alias": "test"},
+    }
+    dd_out = _parse_data_dict(dd, root=root)
+    assert dd_out["test"].path == dd_out["test1"].path
+    # placeholder
```
DirkEilander, line 57: @laurenebouaziz I added unit test for this method. Can you make sure that the part on placeholders makes sense to you?

hydromt/data_adapter.py
```diff
@@ -602,48 +607,52 @@ def _parse_data_dict(data_dict, root=None, category=None):
...
         # lower kwargs for backwards compatability
+        # FIXME this could be problamatic if driver kwargs conflict DataAdapter arguments
         source.update(**source.pop("kwargs", {}))
         for opt in source:
             if "fn" in opt:  # get absolute paths for file names
                 source.update({opt: abs_path(root, source[opt])})
-        data[name] = adapter(path=path, meta=meta, **source)
+        if "placeholders" in source:
+            options = source["placeholders"]
+            for combination in itertools.product(*options.values()):
```
DirkEilander, line 646: @laurenebouaziz I've shortened this part a bit, but it should still function like  you intended it to function. Could you double check with your data?

DirkEilander:
Nice addition @laurenebouaziz ! Can you check my comments? Then it's ready to merge I think.

hydromt/data_adapter.py
laurenebouaziz: thanks @DirkEilander, looks better indeed and still works the same for my test case!

hydromt/data_adapter.py
laurenebouaziz: At some point the placeholders were added to *kwargs and this gave an error in open_mfdataset. So I made a few changes in the code to fix this, but I think you are right, this one is not needed. I will remove this line.

tests/test_data_adapter.py
laurenebouaziz: looks good to me!

laurenebouaziz:
Thanks @DirkEilander! I think it is ready to merge now.

PR #94, opened 2022-02-02
Description of hydromt.io.open_raster_from_tindex

couasnonanais:
The position of the tileindex argument was set under kwargs and not mosaic_kwargs. Documentation of this section modified accordingly


DirkEilander:
Thanks @couasnonanais for fixing this!

PR #99, opened 2022-02-17
Climate data

hboisgon:
Add some additional preprocess functions to open climate datasets:

- transpose_dims to reorder to (time, y_dim, x_dim)
- round_time to harmonize time array between different variables of the same data source

### DIFF
--- a/hydromt/data_adapter.py
+++ b/hydromt/data_adapter.py
@@ -694,7 +694,17 @@ def round_latlon(ds, decimals=5):
     return ds
 
 
-PREPROCESSORS = {"round_latlon": round_latlon}
+def transpose_dims(ds):
+    x_dim = ds.raster.x_dim
+    y_dim = ds.raster.y_dim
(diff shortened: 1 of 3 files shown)


hydromt/data_adapter.py
```diff
@@ -694,7 +694,17 @@ def round_latlon(ds, decimals=5):
     return ds
 
 
-PREPROCESSORS = {"round_latlon": round_latlon}
+def transpose_dims(ds):
+    x_dim = ds.raster.x_dim
+    y_dim = ds.raster.y_dim
+    ds = ds.transpose("time", y_dim, x_dim)
```
DirkEilander, line 700: I think we could do this in the RasterDataset.get_data(), but without specifying the name of the first dimension:

`ds = ds.transpose(..., y_dim, x_dim)`

hboisgon:
@DirkEilander I moved transpose_dims to be automatically done in the RasterDataset adapter. Both hydromt and hydromt_wflow tests ran without errors.

PR #102, opened 2022-02-21
Description of function DataCatalog.from_yml

couasnonanais:
I think the description of the function was not in line with the documentation for the options 'nodata', 'unit_add' and 'unit_mult' (<native_variable_name1> was used instead of <hydromt_variable_name1>).


DirkEilander:
@couasnonanais Thanks for catching this! I noticed we made the same mistake in from_dict which I also corrected.

DirkEilander (approved):

Issue #37, opened 2021-05-12
improve raster.interp_na


DirkEilander:
This function causes inconsistencies between OS, see https://github.com/Deltares/hydromt_wflow/issues/17 and also causes memory issues, see https://gitlab.com/deltares/wflow/hydromt/-/issues/60 

Potentially we could wrap a cKDTree based index method in xr.map_blocks  (with overlap) to  fix this.

DirkEilander:
see discussion in #108 and #105

PR #108, opened 2022-03-01
interpolate_na enhancement

DirkEilander:
Interpolate missing values based on D4 neighbors of missing value cells only. This largely improves the performance without loosing accuracy.

relates to #37

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,6 +13,7 @@ Added
 
 Changed
 ^^^^^^^
+- Interpolate missing values based on D4 neighbors of missing value cells only. This largely improves the performance without loosing accuracy.
 
 Fixed
 ^^^^^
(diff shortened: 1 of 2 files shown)


DirkEilander:
@hboisgon I've already tested this with the hydromt_wflow and hydromt_sfincs tests. Perhaps good to test with a couple more basins before we merge. We could extent the unit test to make sure we get the same result compared to the old implementation.

hboisgon:
When testing for a bigger basin (Wflow upper Danube) I found some differences. I think because we use the 'nearest' method when interpolating soil parameters, another neighbor value is now used in some cases. So this should not cause difference in quality of the interpolation but differences may be observed.

hboisgon:
When testing for a bigger basin (Wflow upper Danube) I found some differences. I think because we use the 'nearest' method when interpolating soil parameters, another neighbor value is now used in some cases. So this should not cause difference in quality of the interpolation but differences may be observed.

hboisgon (approved):
Differences can happen due to 'nearest' method but does not really impact quality of interpolation

PR #116, opened 2022-04-06
Small updates in get_data

hboisgon:
Some changes in get_data:

- variables also allowed to be a str instead of a list of str
- new DeprecationWarning for automatic renaming of single var array in get_data if variables is provided


DirkEilander (approved):
Nice work! Just updated the changelog to be a bit more specific about the changes.

PR #118, opened 2022-04-13
HydroMT API  (for HydroMT-dash)

DirkEilander:
API methods:
- get_model_components
- get_datasets
- get_predefined_catalogs
- Added entrypoints for new submodel classes


hboisgon (approved):
Nice additions!
Good to merge now to have the functions available for the dashboard but some things to remember for future improvements:

- _EXPECTED_TYPES in get_model_components: make sure the list is full (exemple list and tuple are not in)
- Types in get_model_components allow exceptions for Union but not for Optional. For Union 1st keyword is used but sometimes can be list of str or single str

PR #125, opened 2022-05-13
chirps & era5 scripts

DirkEilander:
adding scripts to download and update dynamic forcing datasets

todo: 
- [ ] extend script(s) with some check, e.g. plot at random single point for each variable starting before the update until the last available date


DirkEilander:
the docs build error is causing by this issue: https://github.com/Deltares/hydromt/issues/142

Issue #130, opened 2022-05-29
resolve_path not clear about expected format of {date} and {month}


couasnonanais:
File paths containing {year} or {month} will automatically be read as integer and converted to strings. This may be problematic if file paths were stored with two digits for the month (e.g. 01 for January) or two last digits of the year.

PR #131, opened 2022-05-29
{year} and {month} formatted to expect 4 and 2 digits. Ref #130

couasnonanais:
Added clarification to the docs and string formatting

### DIFF
--- a/hydromt/data_adapter.py
+++ b/hydromt/data_adapter.py
@@ -825,13 +825,15 @@ class DataAdapter(object, metaclass=ABCMeta):
         # Extract key formats
         format_keys = [i[2] for i in Formatter().parse(path) if i[1] is not None]
         # Store keys and keys format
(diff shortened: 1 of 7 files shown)


hydromt/data_adapter.py
```diff
@@ -1267,7 +1287,7 @@ def to_file(
         obj = self.get_data(
             bbox=bbox, time_tuple=time_tuple, variables=variables, logger=logger
         )
-        if obj.vector.index.size == 0 or ("time" in obj and obj.time.size == 0):
+        if obj.vector.index.size == 0 or ("time" in obj.coords and obj.time.size == 0):
```
couasnonanais, line 1290: @hboisgon : I think this is what needs to be fixed in the code

couasnonanais, line 1290: @hboisgon : I found back (I think) the change that makes the PR fail

PR #134, opened 2022-07-15
Small bugfix for {month} in DataAdapter

couasnonanais:
Reading GeoDataset with a time slice with month with one digits was failing depending on how the file was stored. (for eg. January 2012 could be stored as ("2012_01.nc" or "2012_1.nc").  data_adapter will now automatically check for both options to see which one exists. 

Documentation has been changed accordingly. 

Also modified another small bugfix based on an upgrade of xarray I think


DirkEilander:
another solution is proposed in #164

PR #139, opened 2022-07-22
Penman

laurenebouaziz:


hydromt/workflows/forcing.py
```diff
@@ -414,6 +477,153 @@ def pet_makkink(temp, press, k_in, timestep=86400, cp=1005.0):
...
+    kin,
+    wind,
+    var_for_avp,
+    elevtn,
+    doy,
+    lat_rad,
+    var_for_avp_name,
+):
+    """
+    todo
```
markhegnauer, line 494: This TODO is empty, please provide a TODO (I assume the TODO is to add a description of the function). Even better would be to already add the description.

hydromt/workflows/forcing.py
```diff
@@ -414,6 +477,153 @@ def pet_makkink(temp, press, k_in, timestep=86400, cp=1005.0):
...
+    # then net rad
+    # first net rad short
+    net_rad_s = pyeto.net_in_sol_rad(
+        kin * 86400 / 1e6, albedo=0.23
+    )  # in this formula kin should be in [MJ m-2 day-1]
+    net_rad = pyeto.net_rad(net_rad_s, long_out)
+
+    # now eto....
+    pet = pyeto.fao56_penman_monteith(
+        net_rad, pyeto.celsius2kelvin(temp), wind, svp, avp, delta_svp, psy
```
markhegnauer, line 545: As discussed, maybe nicer to have the .celcius2kelvin function outside of the .fao56_penman_monteith function.

markhegnauer, line 545: The function also has a shf parameter, default 0.0. Should we add this default parameter to the call?

markhegnauer (approved):
I think it looks good, some small comments, but not blocking before merging I would say.

hydromt/workflows/forcing.py
laurenebouaziz: good point, done in e3f5cc78d6a41ff9fd8a004b634d437671918440

hydromt/workflows/forcing.py
laurenebouaziz: agree, done in e3f5cc78d6a41ff9fd8a004b634d437671918440

laurenebouaziz: agree done in e3f5cc78d6a41ff9fd8a004b634d437671918440

laurenebouaziz:
@DirkEilander or @hboisgon, Mark did a first review, but probably good if one of you also has a look at it, does one of you have time? it is the penman branch in both hydromt and hydromt_wflow. thanks!

DirkEilander:
> @DirkEilander or @hboisgon, Mark did a first review, but probably good if one of you also has a look at it, does one of you have time? it is the penman branch in both hydromt and hydromt_wflow. thanks!

Thanks for the work @laurenebouaziz! One thing to still consider is whether we want to make pyeto a required dependency (in which case it should also be included in the pyproject.toml). Or make it optional and only import it when you are actually using it. I saw its a very small library with only native python so in that sense adding it as requirement is not a big issue. On the other hand there is no installation package available which makes me advice against it. We could reach out to them and make an issue about creating a pip and/or conda package.

laurenebouaziz:
thanks @DirkEilander! 
For me it is ok if we make it optional. I see that nothing has happened since 2015 in pyeto so I am not sure if issues are still handled.
yes let's have a look together at the unit test! (i ran it for Bhutan and it is very very very slow...)

DirkEilander:
left todo:
- [ ] update or create new data asset for tests
- [ ] create unit tests

DirkEilander:
@B-Dalmijn: Could you try fixing the failing test (add pyet to test and dev environments in envs folder) and the failing code analysis (workflows/forcing.py#L337 and `reproj_method` argument in `press` function). Then I can do a first code review and Laurene later a check on the formula's. Thanks!

Issue #140, opened 2022-07-26
add a version property to the data catalog yml files


DirkEilander:

DirkEilander:
the yml file can now have a `meta` section with amongst others version info:

```
meta:
   version: v2022.7            # we suggest calendar versioning
   root: <path>                 # optional
```

@IMarth can you add this to the deltares_data.yml

PR #146, opened 2022-07-27
restructure predefined data catalogs 

### LINKED ISSUE #62: move deltares_data.yml file
Currently the deltares data is read from https://github.com/DirkEilander/hydromt-artifacts/releases/download/v0.0.4/data_sources_deltares.yml

As we are developing the deltares data and artifacts separately at different rates it is no longer necessary to keep these files together in the hydromt-artifacts repos and it might make more sense to move it back into this repos. 

We could simply keep it in a data folder in the repos which can be accessed by hydromt through the following link (link does not yet exist). In this url _main_ can also be replaced by a tag to refer to previous tagged versions.
https://raw.githubusercontent.com/Deltares/hydromt/main/data/data_sources_deltares.yml

The advantage of keeping it in the repos but outside the package folders is that it is not shipped with the releases and can be updated more frequently. This approach (data catalog yaml files in a data folder) could further be generalized to include other open data sources (e.g. BE data, DD opendap, STAC data) in the future.

DirkEilander:
- predefined data catalogs through new yml file in data folder
- support for data archives
- new global "meta" section in yml file

fixes #62, #140

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,7 +14,7 @@ Added
 - Function ``to_datetimeindex`` in available preprocess functions for xr.open_dataset in the data adapter.
 - Function ``remove_duplicates`` in available preprocess functions for xr.open_dataset in the data adapter.
 - New ``DataCatalog.from_predefined_catalogs`` and ``DataCatalog.from_archive`` to support predefined data catalogs and archive 
-  in a generic way throught the data/predefined_catalogs.yml file.
+  in a generic way through the data/predefined_catalogs.yml file.
 
 Changed
 ^^^^^^^
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,6 +13,8 @@ Added
 - New raster method for adding gdal_compliant() attributes to xarray object.
 - Function ``to_datetimeindex`` in available preprocess functions for xr.open_dataset in the data adapter.
 - Function ``remove_duplicates`` in available preprocess functions for xr.open_dataset in the data adapter.
+- New ``DataCatalog.from_predefined_catalogs`` and ``DataCatalog.from_archive`` to support predefined data catalogs and archive 
+  in a generic way throught the data/predefined_catalogs.yml file.
 
 Changed
 ^^^^^^^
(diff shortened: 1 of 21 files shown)


JoostBuitink (approved):

hboisgon (changes requested):
The changes currently break the CLI interface:

- In artifact_keys, from CLI we always pass {'deltares_data': False} which would make that deltares_data from predefined catalogs is always used.
- data_libs from CLI returns a tuple instead of a list so merging in line 73 of data_catalog.py returns an error.
- when providing a catalog_name instead of path with -d, the CLI still automatically turns it into a path

DirkEilander (commented):
@hboisgon the PR is ready for a new round of reviews :)

hydromt/data_catalog.py
```diff
@@ -32,60 +36,76 @@
...
-        if data_libs is not None:
-            for path in np.atleast_1d(data_libs):
-                self.from_yml(path)
+            elif isinstance(version, str):
+                lib += f"={version}"
+            data_libs = [lib] + data_libs
+
+        # parse data catalogs; both user and pre-defined
+        for name_or_path in data_libs:
+            if name_or_path.endswith(".yml"):  # user defined path
```
hboisgon, line 81: Do we only support data_catalogs with .yml extension or should we also read in if .yaml ? Else we need to be very careful in the docs / error message.

docs/changelog.rst
```diff
@@ -13,6 +13,8 @@ Added
 - New raster method for adding gdal_compliant() attributes to xarray object.
 - Function ``to_datetimeindex`` in available preprocess functions for xr.open_dataset in the data adapter.
 - Function ``remove_duplicates`` in available preprocess functions for xr.open_dataset in the data adapter.
+- New ``DataCatalog.from_predefined_catalogs`` and ``DataCatalog.from_archive`` to support predefined data catalogs and archive 
+  in a generic way through the data/predefined_catalogs.yml file.
```
hboisgon, line 17: Maybe good to also add it the list of newly available predefined catalogs / versions?

hboisgon (approved):
Nice changes @DirkEilander ! I tested again and it all works again from CLI :) I have two small minor comments but I think this is now good to go!

docs/changelog.rst
DirkEilander: We'll keep a separate changelog file for the data catalogs as those are not connected to the hydromt versions, see #156

hydromt/data_catalog.py
DirkEilander: Good one, will make sure both extensions are recognized!

PR #153, opened 2022-07-29
add DataFrameAdapater for table and 1D timeseries data

JoostBuitink:


DirkEilander (changes requested):
Thanks for this PR @JoostBuitink 

It looks great in general. Good tests and docs!

I made some small changes in the docs:
- added the functions in api.rst
- made a reference for dataframes in data_types.rst L472

Two small suggestions:
- check the docstrings (especially to_file has some old references)
- create tests for to_file

DirkEilander (approved):

Issue #159, opened 2022-08-02
setup_response_unit: region and subbasins


bartvanosnabrugge:
create setup methods for LumpedModel Class. See also #151 .

- [ ] setup_region
- [ ] setup_subbasins
- [ ] setup_downstream_links
- [ ] setup_response_unit

**setup_region**
check that setup_region works for lumped/semi-distributed models, and set region to model as response_unit

**setup_subbasins**
wrap functions from here https://deltares.github.io/pyflwdir/latest/basins.html to divide basins into subbasins and add (overwrite) response_unit.
Note: this step could also derive stream links.

**setup_downstream_links**
Take the region and subbasins and determine the downstream subbasin of each basin
Note: this could / should be extended to include stream links.

**setup_response_unit**
combine above steps in one initial step.

DirkEilander:
This links to #497

DirkEilander:
this links to #149

alimeshgi:
Closing this issue. The VectorModel class was removed in the v1 refactoring, and PR #160 was closed without merge. The concepts of response_unit and downstream_link were never adopted. Low-level subbasin delineation exists in gis/flw.py and model/processes/region.py. Please reopen with a proposal targeting VectorComponent and the current process architecture if linked-subbasin support is still needed.

PR #160, opened 2022-08-02
setup_response_unit: region and subbasins 

bartvanosnabrugge:
See issue #159 

Informal test script available, needs help with formal testing.


savente93:
I'm closing this PR as it has not diverged from the main branch and it doesn't make sense to implement the relevant functionalities from here instead from the main branch. The PR will still remain available through GitHub so we can come back to the code later if necessary

PR #170, opened 2022-08-03
prepare release v0.5.0

DirkEilander:
- bumped version to v0.5.0
- matched dependencies in pyproj.toml with all environment yml files


binder/environment.yml
```diff
@@ -1,33 +1,40 @@
...
   - requests
+  - rioxarray
   - scipy
+  - toml
   - xarray
-  - zarr
\ No newline at end of file
+  - zarr
+  - pip:
+    - hydroengine
```
hboisgon, line 39: I see you added hydroengine, do we need it in hydromt core? I thought just hydromt_wflow

binder/environment.yml
```diff
@@ -1,33 +1,40 @@
...
   - affine
-  - click
   - black
   - bottleneck
   - cartopy # to run examples
+  - click
   - dask
   - descartes # to run examples
   - entrypoints
   - flit>=3.2
```
hboisgon, line 16: Shall we do flit 3.4 to be sure pip install works ?

hboisgon (approved):
Looks good! Just two small comments for flit minimum version and hydroengine dependency.

binder/environment.yml
DirkEilander: Good one. I've changed this in all environment files and the pyproj.toml

binder/environment.yml
DirkEilander: Removed hydroengine here and marked it optional in hydromt-dev.

PR #171, opened 2022-08-03
Remove unnecessary nodata value not found warnings

hboisgon:
Too many unnecessary warnings message of type "No numerical nodata value found, skipping set_nodata" were issued by HydroMT.

This happened in two places:

- calls to raster.geometry_mask (example in get_rasterdataset if geom is not None)
- calls to raster.nodata (calling a set_nodata even if both rio.nodata and rio.encoded_nodata are None already)


DirkEilander (approved):
Nice work!

Issue #167, opened 2022-08-03
extend raster.zonal_stats with count of classes (mode, percentage)


bartvanosnabrugge:
**use case**
For lumped models (geometries), derive the count, mode, or percentage that is covered by a certain land class.

**Current behavior**
zonal_stats can calculate several stats, but not counts (and related stats) related to categorical grids (i.e. the count of occurrence of a land class).

**simple solution**
wrapper around zonal_stats that runs zonal_stats('count') for each class in the grid, with all other values masked.
see [here](https://github.com/bartvanosnabrugge/summaWorkflow_public/blob/20781916e85b863d8407655175fe7b76b135b305/cwarhm/model_agnostic_processing/HRU.py#L208).

**elaborate solution**
add functionality to zonal_stats itself.

- check which classes exists (globally, or provide a list)
- use numpy.unique to gie both values and counts
- store counts (or other stats) in additional dimension 'classes' to zonal_stats xarray output

PR #175, opened 2022-08-04
Submodelclass zonalstatsperclass

bartvanosnabrugge:
#167 

This implements the 'simple solution' to cover the needed functionality, while envisioning how the results can be stored in a multidimensional xr.DataArray.

A dimension is added to store the count for each class, and the class itself is added as coordinate.

Would need help to merge into one zonal_stats function. That activity would be currently out of scope for the SemiDistributedModel project.

Attached test file shows how the method works and how it can be used to calculate basin fractional cover and basin mode.

[test_zonal_stats_per_class.py.txt](https://github.com/Deltares/hydromt/files/9259755/test_zonal_stats_per_class.py.txt)


bartvanosnabrugge:
only merge after submodelclass and submodelclass-bart is merged

bartvanosnabrugge:
@couasnonanais This pull request pulls in the latest corrections and edits to test the response_units introduced in submodelclass. This branch is not 'cleanly'only doing the intended... however this makes branch management quite easier as the latest additions all require the added zonal_stats_per_class function.

hydromt/models/model_lumped.py
```diff
@@ -1,3 +1,21 @@
+from multiprocessing.sharedctypes import Value
```
DirkEilander, line 1: please remove unused imports

hydromt/models/model_lumped.py
```diff
@@ -153,56 +183,67 @@ def __init__(
             data_libs=data_libs,
             logger=logger,
         )
-
-    def read(
```
DirkEilander, line 157: should read indeed be deleted?

hydromt/models/model_lumped.py
```diff
@@ -122,14 +145,21 @@ def write_response_units(
...
-            # write geometry
+            # write geometries
             ds = self._response_units
-            gdf = gpd.GeoDataFrame(geometry=ds["geometry"].values, crs=ds.rio.crs)
-            if not isdir(dirname(join(self.root, fn_geom))):
-                os.makedirs(dirname(join(self.root, fn_geom)))
-            gdf.to_file(join(self.root, fn_geom), driver="GeoJSON")
+            # get list of coordinates that hold geometries
+            corl = list(ds.coords)
+            geom_coords = [v for v in corl if 'geometry' in v]
```
DirkEilander, line 152: Nice addition to write multiple geometries! I wonder if we cannot better do a data type check rather than a name check. I.e. you could check if the first value of every 1D coordinate is an instance of a shapely BaseGeometry.

hydromt/models/model_lumped.py
```diff
@@ -153,56 +183,67 @@ def __init__(
...
+        ):
+        """
+        This component sets up the region and response units.
+        
         """
-        super().read(components=components)
 
-    def write(
+        if len(self.region) == 0:
+            raise ValueError("No region defined. Define :py:meth:`~hydromt.models.model_api.setup_region`")
```
DirkEilander, line 200: Links to methods don't work in error messages, this is sphinxs docstrings specific I believe.

hydromt/models/model_lumped.py
```diff
@@ -153,56 +183,67 @@ def __init__(
...
+        
+        if not split_regions:
+            self.set_response_units(self._geoms['region'])
```
DirkEilander, line 203: I'm wondering if this will work as we don't yet support GeoDataFrames in set_response_units. Also I would always use `self.region` instead of `self._geoms['region']`.

hydromt/raster.py
```diff
@@ -789,6 +789,51 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
         ds_out["index"] = xr.IndexVariable("index", gdf.index.values[np.array(idx)])
 
         return ds_out
+    
+    def zonal_stats_per_class(self, gpd_shape: gpd.GeoDataFrame ,class_list: list, stat: str, class_dim_name='class',all_touched=False):
```
DirkEilander, line 793: would be good to add a test to tests/test_raster.py

hydromt/workflows/to_response_units.py
```diff
@@ -0,0 +1,160 @@
...
+import geopandas as gpd
+import hydromt
+
+logger = logging.getLogger(__name__)
+
+__all__ = ["make_ds_mask", "hydrography_to_basins","create_response_unit_ds",
+           "ru_geometry_to_gpd", "fracs", "ds_class_mode"]
+
+
+def make_ds_mask(ds, gdf, col_name='value'):
```
DirkEilander, line 17: I think the `xr.Datasset.raster.geometry_mask` function is what you are looking for.
https://deltares.github.io/hydromt/latest/_generated/hydromt.DataArray.raster.geometry_mask.html

hydromt/workflows/to_response_units.py
```diff
@@ -0,0 +1,160 @@
...
+    if np.shape(geometries)==():
+        unit_df = pd.DataFrame(data=dict(geometry=geometries.item(),value=indexes.item()),
+                                index=[indexes.item()])
+    else:
+        unit_df = pd.DataFrame(data=dict(geometry=geometries,value=indexes),
+                                index=indexes)
+    unit_gpd = gpd.GeoDataFrame(unit_df, crs = ru_ds.rio.crs)
+    return unit_gpd
+
+def hydrography_to_basins(
```
DirkEilander, line 86: Nice function! I think this would fit well in flw.py with a test in tests/test_flw.py and could perhaps be renamed to `subbasins`

DirkEilander (commented):
Thanks for your work on this PR @bartvanosnabrugge 
I've already added a few comments that I hope are helpful at this stage.

hydromt/models/model_lumped.py
DirkEilander: idem for write

hydromt/models/model_lumped.py
bartvanosnabrugge: Done

hydromt/models/model_lumped.py
bartvanosnabrugge: I put read and write back and the region property to. Indeed to enthusiast removing these -> confusion with the write_response_units in the mixedin class. Also,it seems that the addition here is minimal, only a different list is passed on to the super function.

hydromt/models/model_lumped.py
bartvanosnabrugge: Yes that could be a check, likely to be more fail safe (in combination with a naming convention that involves the word geometry I would say, to keep that from geoDataframes.) I put it in as a TODO recommendation for now.

hydromt/models/model_lumped.py
bartvanosnabrugge: Changed to "No region defined. Define region first with setup_region."

hydromt/models/model_lumped.py
bartvanosnabrugge: Good point. I started the original by just parsing a geodataframe with geometries, then changed to derive the region based on points. Good example why multiple tests are needed.

On topic: I think parsing a geoDataFrame into the response unit is a good functionality to have. For example when reading external (preprocessed) geoDataFrames Either standalone, or in the set_response_units function.

hydromt/workflows/to_response_units.py
bartvanosnabrugge: Yes that would fit nicely. Something to consider is that currently it outputs a response_unit like dataset - it could also output a geoDataFrame with two geometry columns, which can then be parsed into a response_unit in the lumpedModel class. I think that would keep this function cleaner?

I find subbasins a bit of a too general name perhaps. (from that perspective hydrography_to_basins is also not too descriptive

hydromt/models/model_lumped.py
bartvanosnabrugge: This is now solved by adding the function gpd_to_response_units in workflows.to_response_units

hydromt/workflows/to_response_units.py
bartvanosnabrugge: I think this function is no longer used, will check and delete.

savente93:
I'm closing this PR as it has not diverged from the main branch and it doesn't make sense to implement the relevant functionalities from here instead from the main branch. The PR will still remain available through GitHub so we can come back to the code later if necessary

Issue #191, opened 2022-09-20
create docs for Model subclasses


couasnonanais:

DirkEilander:
fixed in #198

PR #196, opened 2022-09-21
Description of model classes in the docs. Ref #191

couasnonanais:
First implementation of updated docs to describe the submodelclass. 
Some points not fixed:

-  API reference is showng up properly. This is because the importing of the module is not done properly. But not sure how to fix that in the code. 
- Code snippets have been added but not tested
- Examples about how to clip a model with submodel class have not been added
- I realize now that it would be good to write somewhere that the postprocessing methods (after the model runs) have not been implemented (yet?) so those functionalities are only for plugins


couasnonanais:
Also forgot to mention:

- The model data components table in model_main.rst file should be updated for "maps" reference and not auxmaps in the API column
- One table is not showing up properly

couasnonanais:
Ready for review

DirkEilander:
@couasnonanais I did not see this branch before and already merged #198. Is this branch different or can we close it?

DirkEilander (approved):
Thanks @couasnonanais !

PR #198, opened 2022-09-22
Doc anais

couasnonanais:


DirkEilander (approved):
Thanks for this work @couasnonanais! Huge improvement! I've made some small final changes, but I think it's ready to go now.

@hboisgon: I've now added the Maps component to the Model class as discussed. I think that on the long term (once we have forgotten about staticmaps) this will be more logical and consistent with the geoms component.

Issue #201, opened 2022-09-30
expose submodelclasses in entrypoints for cli


DirkEilander:
TODO
- [x] create entrypoints for hydromt submodel classes bypassin the pyproj.toml file
- [x] remove the dynamic assignment of model plugins the the hydromt namespace
- [x] develop model catalog object

PR #202, opened 2022-09-30
add model catalog #201

DirkEilander:


hboisgon (approved):
Nice changes! I just found a bug if xugrid is not in the environment for mesh_model ep. Hope you agree with my fix

hboisgon:
Related to the HAS_XUGRID and mesh_model import I also found a bug in the previous way we loaded entrypoints as well and corrected it in this PR: https://github.com/Deltares/hydromt/pull/211
I thought this one was merged already so there may be some merge conflicts. @DirkEilander not sure which one you would prefer to merge first in that case?

DirkEilander:
Thanks @hboisgon! I made a small final change to fix you proposed and expanded the tests to cover this exception.

Issue #210, opened 2022-10-10
Bug: raster.dim0 and check_dimensions


hboisgon:
In a recent commit, we changed raster.check_dimensions to not add dim0: None to the attrs

https://github.com/Deltares/hydromt/blob/69f6419046101d1fb3b6f9c8d86f43d8ff5dbad2/hydromt/raster.py#L400
https://github.com/Deltares/hydromt/blob/69f6419046101d1fb3b6f9c8d86f43d8ff5dbad2/hydromt/raster.py#L401

In case a raster has an old dim0 attrs that should be removed, dim0 should still be remove with a pop instead of a None (this is the case for soil parameters in Wflow after averaging per soil layer, dim0 'layer' is still in the attrs when calling raster.interpolate_na):

https://github.com/Deltares/hydromt_wflow/blob/4f2639259f7ffc2e4489a991cd0cefc02a1f6227/hydromt_wflow/workflows/soilgrids.py#L547

PR #211, opened 2022-10-10
Bugfix #210

hboisgon:
Bugfix #210 dim0 removal in raster._check_dimensions

While testting I did not have xugrid and found a bug in ep.load in model_plugins.py if an exception is found while loading entrypoints.


DirkEilander (approved):
Good fix! I thought this was fixed before though...

PR #216, opened 2022-10-19
Example point data

IMarth:
- [x] disappears name in the path when having it online? 
- [x] please check the content
- [x] added nbsphinx_allow_errors=True to conf.py --> check whether it needs to be kept
- [x] remove data content from branch


hboisgon (approved):
Nice example @IMarth ! Thanks :) 
I updated a bit the last part with geodataset from csv that was not working for me and remove the allow_errors flag

DirkEilander (approved):
Thanks @IMarth and @hboisgon ! Very useful example. I've just changed the some terminology to be more consistent with the docs.

PR #217, opened 2022-10-19
notebook for tabular data

IMarth:
Work in progress


DirkEilander (approved):
Thanks for this PR @Tjalling-dejong and @IMarth !

I've just changed the xls example to a more common reclassify example.
Also note that you can add kwargs to the yml file to parse the tables upon reading (so that you don't have to do that in the code afterwards (see updated example).
Finally, note that you should always submit cleaned notebooks to avoid lot's of data on github.

Issue #218, opened 2022-10-24
notebook to export data


hboisgon:
Notebook to export data based on data catalog

PR #222, opened 2022-10-24
added export_data notebook

hboisgon:
See #218


DirkEilander (approved):
Nice notebook @hboisgon !

I've made a few minor changes:
- save output in tmp* folder which will automatically be ignored
- use consistent US English for data catalog
- remove the cli/api.py methods. These might still change and I prefer to focus on the DataCatalog methods

Issue #221, opened 2022-10-24
notebook to explore new submodel classes


hboisgon:
Notebook to explore new submodel classes.

For now quite general only maps and mesh have some setup_ methods but lumped not yet (PR in progress)

Can maybe later be based for the simple_hydrology model from hcwinsemius in https://github.com/hcwinsemius/deltares_snakemake

PR #229, opened 2022-10-25
Working with models example

hboisgon:
Added a new notebook on how to work with models in python using grid_model as an example.
See #221 

I also detailed all installation steps to install and run the notebooks


DirkEilander (approved):
Thanks for the notebook example @hboisgon. I've made a few changes. If you agree with the changes the PR is ready to merge.

- added a "working with models from CLI"  notebook
- added a `setup_maps_from_raster_reclass` example
- changed some python code where you look up available methods and docstrings to references in the docs, because I think that's a better way of working.
- removed the overview picture. While I think the picture is  great in a ppt where you can explain it, it's too complicated in an example without explaining the details.
- moved some explanation from the example to the docs user guide
- changed references to online url so that these also work if running locally

DirkEilander:
@hboisgon any clue why these tests are failing? I don't see how my changes would affect the tests.

hboisgon:
I guess this is the problem: https://pypi.org/project/pytest-benchmark/4.0.0/#history
pytest-benchmark was released today/yesterday and is the line causing the error

Issue #220, opened 2022-10-24
notebook for flw methods


hboisgon:
Notebook to discover methods in flw.py and pyflwdir library

PR #231, opened 2022-10-26
add flow direction example #220

DirkEilander:


hboisgon (approved):
Very nice notebook! Thanks @DirkEilander

Issue #219, opened 2022-10-24
notebook prep data catalog


hboisgon:
Notebook with examples ho to prepare (local) data catalog.
Can re-use example from hydromt-delwaq: https://deltares.github.io/hydromt_delwaq/latest/_examples/adding_local_emission.html

PR #232, opened 2022-10-26
Prepare data catalog notebook

hboisgon:
See #219


DirkEilander:
@hboisgon I don't  have time to finish this also, but I had a quick go at it. The setup looks great! I have tried to use the artifact_data as I think this is much cleaner. We can still show all we want with this data I think. If you agree we can delete the data you added in the data folder.

hboisgon:
@DirkEilander I finished and already merged the notebook so that people that may download hydromt this morning have it already. You already pre-checked the draft sohopefully also fits what you had in mind. The branch is still there so you can also easily re-open the PR if you want to modify still.

hboisgon:
Reopened in #234

PR #235, opened 2022-10-27
Fix path in setting up examples

JoostBuitink:
Updated the `cd` argument to open the examples directory inside the hydromt folder


DirkEilander (approved):
Thanks for checking all the notebooks @JoostBuitink !

PR #250, opened 2022-11-29
Draft: support CMIP6 from Google Cloud

hboisgon:
This PR set different steps to support CMIP6 data from Google Cloud Storage:

- [x] new predefined catalog for remote data: ``cmip6_data``
- [x] support for remote data with a yml meta flag for ``remote`` and a per source new ``filesystem`` attribute.
- [x] use fsspec and upath in ``resolve_path``. Additionnal dependency to upath and gcsfs for Google Cloud
- [x] direct support for CMIP6 dataset requires a new ``harmonise_dims`` preprocess function
- [x] RasterDataset zarr driver: possiblility to read from several zarr stores. Enable PREPROCESSORS before merging
- [x] add test
- [x] instead of remote flag try fixing abs_path method in data_catalog (use uri_validator) or at from_dict level before abs_path
- [x] for gs: fix in glob maybe look at https://github.com/fsspec/universal_pathlib (s3fs has the same problem)
- [x] add docs
- [ ] optional: add example --> when less experimental

**[edit] general comments**
I tested for s3 to see if the gcs implementation is robust enough.
For s3, authentication information are need even if anonymous information for connection to public cloud buckets.
So I did some research to see how to pass authentication information within that PR.
Findings:

- ``pandas`` supports to have "storage_options" that are passed directly to fsspec for all the readers (csv, excel..).
- ``xarray`` supports "storage_options" for ``open_zarr`` only and else storage_options can be put into backend_kwargs in open_dataset but this only works with engine="zarr" see [link](https://docs.xarray.dev/en/stable/user-guide/io.html#cloud-storage-buckets)
- ``xarray`` ``open_dataset`` support file-like object as so we can pass a fs.open(filename) with the authentication credits instantiated in the fsspec filesystem object fs. ``open_mfdataset`` does not support this`.
- geopandas read_file uses fiona.open() in the background. Not fully sure if there is a way to pass "storage_options" to it. But it does accept file-like object like open_dataset.


hboisgon:
@DirkEilander : I implemented some new steps to be able to directly access CMIP6 data from Google Cloud Storage using the pangeo example you sent.
It seems to work now but before I go further with testing and documentation, I would really like to have your first comments on it

hboisgon:
About universal_pathlib: I tried using UPath only in resolve_path method (commented out in commit https://github.com/Deltares/hydromt/pull/250/commits/e5e756923e89007c8b9318c7abd9903aec22a52e). I think the main issue is that with some filesystems, the first part of the path is not understood correctly (pathlib root/driver/anchor are empty for gcs). Going around is what I found:

- compared to fsspec (which is used in the background) is that it uses ``pathlib.Path.glob`` method which is of type path.glob("*") instead of glob(path). Which means the path needs to be properly split.
- If '*' are at the directory levels and not filename (eg: "gs://cmip6/CMIP6/CMIP/{model}/historical/{member}/{timestep}/{variable}/*/*"), there is no easy way to get the last part of the path ("*/*") than using ``pathlib.Path.parent`` and ``pathlib.Path.parts`` and recollating together (splitting str with * causes problems as well as parts of the path can become incomplete if "/tas*/variable_*.py"). 
- When using ``pathlib.Path.parts`` to recreate the last part of the path, problem is then that the filesystem is lost and goes back to os by default (would get *\\* instead of */* on windows which results in no file found).
- You can access the specific ``sep`` of a filesystem in pathlib with ``Path._flavour.sep`` but this is not in the official API...
- Pathlib method like ``Path.relative_to`` would strangely give "gs://cmip6/*/*"" instead of just "*/*".

In the end I could still use UPath and ``joinpath`` in order to correct that the prefix "gs://" is lost when using fsspec glob method.

DirkEilander:
Thanks for the update @hboisgon! 

Some thoughts/ suggestions:
- I'd be nice if we could directly include other filesystems like s3 in this PR
- Ideally the specific filesystem packages (e.g. gssfs) should not be a required dependency. We could add a list of optional 'cloud' dependencies. But than we need to make sure that hydromt runs without these dependencies as well. I can do some recoding to make this work if you want me to.
- The fix with UPath you applied, would that work with normal Path as well?

Other interesting open datasets for our reference:
- ERA5 (GS) https://cloud.google.com/storage/docs/public-datasets/era5
- Copernicus DEM (S3) https://registry.opendata.aws/copernicus-dem/

hboisgon:
Hi @DirkEilander !

Agreed we can use this PR to also include other filesystems and a bit more data. 
I also thought it would be nice to have gcsfs, s3fs as optional dependencies. We could add some steps in the try/except part of resolve path where we import the different packages.

The fix with UPath should work for other filesystems as well. It basically replaces the first path of the path which is incomplete after going through glob. I assume also what may happen with s3 but didn't test it yet.

Does not really matter if you or I continue with this PR, whatever you prefer :)

hboisgon:
Note: also tested compatibility with CST scripts to derive climate statistics and with hydromt_wflow to use to prepare precip and temp_pet forcing.

hboisgon:
I think this PR is ready and in the changelog I flagged remote data as experimental. I throw error messages for all non supported or tested drivers (except DataFrame but this one should really work out of the box).
I tested the two datasets you mentioned Dirk:

- Copernicus DEM (S3): no *.vrt files is available and individual tiles are stored in separate folders. This makes finding the data and reading very slow. So I skipped for now.
- ESA Worldcover (S3): this one does have a *.vrt file for version v100/2020 so can be used directly :) it is missing for v200/2021 so same problem as Copernicus DEM
- ERA5 (GCS): very easy to read but dimension are (time, values) with longitude and latitude in the coordinates. So either needs regridding or maybe the rotated grid PR to be useful.

data/predefined_catalogs.yml
```diff
@@ -13,6 +13,16 @@ deltares_data:
      v2022.7: b4542a6882d4c0709591be11adbef50ba4b832c8 
      v2022.5: d88cc47bd4ecc83de38c00aa554a7d48ad23ec23
    notes: This data is only accessible when connected to the Deltares network.
+cmip6_data:
```
DirkEilander, line 16: perhaps rename to "gcs_cmip6_data". Also the actual filename etc.

hydromt/data_adapter/dataframe.py
```diff
@@ -143,8 +149,18 @@ def get_data(
...
 
+        # Extract storage_options from kwargs to instantiate fsspec object correctly
+        if "storage_options" in self.kwargs:
+            kwargs = self.kwargs["storage_options"]
+            # For s3, anonymous connection still requires --no-sign-request profile to read the data
+            # setting environment variable works
+            if "anon" in kwargs:
+                os.environ["AWS_NO_SIGN_REQUEST"] = "YES"
+            else:
+                os.environ["AWS_NO_SIGN_REQUEST"] = "NO"
```
DirkEilander, line 160: perhaps this could eventually be generalized in storage_options method of the DataAdapter class

DirkEilander (approved):
Nice contribution @hboisgon !
I have just renamed the cmip6_data and edited some docstrings.

Going forward it would be good to evaluate how to better structure these different filesystems in the data adapter.

PR #255, opened 2022-12-12
update data_artifacts to v0.0.7

DirkEilander:
add era5 zarr data with more variables

see artifact data:
PR https://github.com/DirkEilander/hydromt-artifacts/pull/5 
release https://github.com/DirkEilander/hydromt-artifacts/releases/tag/v0.0.7


DirkEilander:
@hboisgon I've only added new data here (era5 zarr with more variables) and double checked that the plugin test stil work with the new version. In this case the dates in the wflow toml file are modified when using the zarr instead of the nc data, so i've decided to also keep the nc data. It would be good to update the tests (and examples!) to work with the zarr data.

PR #258, opened 2022-12-14
make pygeos an optional dependency

### LINKED ISSUE #228: test with new shapely v2 instead of pygeos
With geopandas v0.12, the new shapely v2 is supported and pygeos can be removed.
I suggest to create different test environments, one with shapely v1.x and pygeos and one with geopandas>=v0.12 and shapely>-2

see [geopandas changelog](https://github.com/geopandas/geopandas/blob/main/CHANGELOG.md), [shapely migration docs](https://shapely.readthedocs.io/en/latest/migration.html) and [shapely changlog](https://shapely.readthedocs.io/en/latest/release/2.x.html#version-2-0-0-in-progress)

DirkEilander:
Pygeos is still required from some xugrid functionality we use. For now I have added it to the list op optional dependencies.

closes #228


DirkEilander:
This looks good Brendan! Thanks for the work!

One small (side) note. I prefer not to have [the same fix](https://github.com/Deltares/hydromt/pull/258/commits/c12b77cd826a0bb6f65198db25a0e8e1cd779127) committed twice in different branch to avoid merge conflicts later. In that case it's better to merge the other branch with the fix.

PR #266, opened 2023-01-12
Penman

DirkEilander:
restarted https://github.com/Deltares/hydromt/pull/139 to ask @laurenebouaziz for a review. Could you let us know when you have time to work on this. Ideally we include it in the next release which we have planned for this month.


hydromt/workflows/forcing.py
```diff
@@ -248,29 +330,95 @@ def pet(
...
+                altitude=wind_altitude,
+                altitude_correction=wind_correction,
+            )
+        if method == "penman-monteith_rh_simple":
+            pet_out = pm_fao56(
+                temp["temp"],
+                temp["temp_max"],
+                temp["temp_min"],
+                ds["press"] / 10,
+                ds["kin"],
```
laurenebouaziz, line 395: unit conversion is done at this level for pressure but inside the function for kin, please check if this should be changed

tests/test_forcing.py
```diff
@@ -35,3 +35,14 @@ def test_precip():
...
+
+def test_pet():
+    cat = DataCatalog()
+    et_data = cat.get_rasterdataset("era5_daily_zarr")
+    dem = cat.get_rasterdataset("era5_orography").squeeze("time").drop("time")
+
+    peto = pet(et_data, et_data, dem, method="penman-monteith_tdew")
+
+    assert peto.raster.shape == dem.raster.shape
+    np.testing.assert_almost_equal(peto.mean(), 0.57746, decimal=4)
```
laurenebouaziz, line 48: when I run the test I get the following value: 0.57739717
still very close to what you have but not exactly the same.

hydromt/workflows/forcing.py
```diff
@@ -414,6 +562,105 @@ def pet_makkink(temp, press, k_in, timestep=86400, cp=1005.0):
...
+    )
+    nr = swr - lwr  # daily net radiation
+
+    # Penman Monteith FAO-56
+    gamma = pyet.calc_psy(press)  # psychrometric constant
+    dlt = pyet.calc_vpc(
+        temp
+    )  # Slope of saturation vapour pressure curve at air Temperature [kPa °C-1].
+
+    gamma1 = gamma * (1 + 0.34 * wind)
```
laurenebouaziz, line 653: can we add more description about the variables / units used in this function?

hydromt/workflows/forcing.py
```diff
@@ -248,29 +330,95 @@ def pet(
...
+                ds["rh"],
+                dem_model,
+                "rh",
+            )
+        elif method == "penman-monteith_tdew":
+            pet_out = pm_fao56(
+                temp["temp"],
+                temp["temp_max"],
+                temp["temp_min"],
+                ds["press"] / 10,
```
laurenebouaziz, line 406: temp variable contains the downscaled temperature

laurenebouaziz (changes requested):
Thanks, nice work of using the pyet package now!

I tested the workflow for the review and pushed the following changes:

- I added more explanation on the variables in the pm_fao56 function. Please check if more details can be added (units etc.) 
- some of the variables (wind and dewpoint temperature) were not yet renamed in the functions, this is now changed. 
- I made some changes to make sure that if downscale temperature is applied, the downscaled temperature is also used in the pet calculation. 
- I made some changes in the test and in the artifact yml file to make sure renaming and unit conversions are not required in the test. 
- I also made changes in the wflow.py file within hydromt_wflow to make sure setup forcing works with the latest updates of pyet. Perhaps good if this is reviewed/checked. 
- when I compare the pet that I calculated for Bhutan with the previous workflow, I see some small differences. however, I was not able to find the cause of this. Not sure if this should be investigated further. 

@B-Dalmijn , please check the changes I made and last small comments and also the hydromt_wflow penman branch. After this I think it is ready to merge.

hydromt/workflows/forcing.py
B-Dalmijn: This is correct, hpa vs kpa

tests/test_forcing.py
B-Dalmijn: ??? no clue

hydromt/workflows/forcing.py
B-Dalmijn: We probably should. I'll look at it

hydromt/workflows/forcing.py
B-Dalmijn: Yes.

PR #275, opened 2023-01-24
fix eval setup_config 

### LINKED ISSUE #274: values in [setup_config] section of ini files not evaluated
This bug originates in PR https://github.com/Deltares/hydromt/pull/246

The required behaviour is that only paths in setup_config are not evaluate and changed into absolute paths by default. Other python objects should be evaluated. 

Suggested solution: Distinguish between the `skip_eval_sections` (`[]` by default) and a new `skip_abspath_sections` optional argument (`['setup_config']` by default) in configread.

this fixes https://github.com/Deltares/hydromt_wflow/issues/151

DirkEilander:
fixes #274


B-Dalmijn (approved):
Seems to be fine!

PR #277, opened 2023-01-30
CLI revision (i.e. region & res)

B-Dalmijn:
TODO:

- [ ] remove res from cli build, make region optional 
- [ ] idem in model_api build functions
- [ ] add nice warning message if old -r is used & add deprecation in help
- [ ] add extra option for file overwriting, plus additional error and warning


DirkEilander:
TODO:

- [ ] update user guide - building a model. Include a NOTE on the changes from v0.7.0

PR #286, opened 2023-02-13
add vrt tile caching functionality

DirkEilander:
This branch adds caching for tiled raster datasets.

@hboisgon can you review this part and then merge everything and release? Could you especially advise on the used xml parser to read vrt tiles? I used xml2dict, mostly for convenience (simple interface) and because I'm only interested in reading but I haven't check the performance against other parsers. If you think another parser is better / more in line with the FEWS branch I'm also happy to change / feel free to change that code.


hboisgon (approved):
Thanks Dirk for the changes! For the xml library at the time we looked into declxml but chose to keep using ET (did not fit all xml options that FEWS uses) and does not have a conda release I think. So we stuck to etree which has a lot of flexibility but is not so easy to use. I didnt know about xmltodict but looks great for what you want to do with it!

Issue #284, opened 2023-02-10
Make arguments in raster.clip_geom and raster.clip_mask consistent


DirkEilander:
```
def clip_mask(self, mask: xr.DataArray):
  """
  Arguments
  ---------
  mask : xarray.DataArray
      Mask array.
  """
```

should change into 

```
def clip_mask(self, da_mask: xr.DataArray, mask: bool = False):
  """
  Arguments
  ---------
  da_mask : xarray.DataArray
      Mask array.
  mask: bool, optional
      Mask values outside geometry with the raster nodata value
  """
``"

PR #290, opened 2023-02-22
ENH: consistent raster.clip_mask and clip_geom methods

Tjalling-dejong:


Tjalling-dejong:
See issue #284

DirkEilander (approved):
Looks good @Tjalling-dejong 
I think we should however wait with merging until we have released hydromt_wflow to avoid issues with the current version.

PR #291, opened 2023-02-23
ENH: cache predefined data catalogs file to be able to work offline

DirkEilander:
This PR makes it possible to work with artifact data (required for many tests!) when offline. Currently it relies on the online predefined data catalog yaml file. With this change is caches that file to ~/.hydromt_data and uses the cached version if now internet connection is available.

A small modification to the _copyfile method was required to be able to download text data (this didn't work with shutil.copyfileobj in combination with the request.get().raw attribute). I've tested this still works with other data formats as well.

### DIFF
--- a/hydromt/data_adapter/caching.py
+++ b/hydromt/data_adapter/caching.py
@@ -26,7 +26,7 @@ def _uri_validator(uri: str) -> bool:
         return False
 
 
-def _copyfile(src, dst):
+def _copyfile(src, dst, chunk_size=1024):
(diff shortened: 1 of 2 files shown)


DirkEilander:
@B-Dalmijn Could you review this?

B-Dalmijn:
I'll try to get around it next week.

alimeshgi:
@B-Dalmijn , did you manage to review this pull request?

Tjalling-dejong (approved):
LGTM 👍

PR #292, opened 2023-02-24
ENH: support YAML configuration files

Tjalling-dejong:
TODO
- [x] add toml parser to config.py. Note that we will need a short postprocessing for absolute paths (see also configread method)
- [x] in cli/main.py we should decide whether to use the toml parser or configread method. Also throw a PendingDeprecation warning if ini is used.
- [x] update the documentation and examples.


Tjalling-dejong:
@DirkEilander Do we want to have ini replaced everywhere in the docs and the examples or keep both in for now?

Tjalling-dejong:
- Removed toml support from config.py
- Added yaml parser
- Updated configread to use yaml parser while keeping ini available
- Updated config.py tests

Tjalling-dejong:
TODO
[] - Update documentation

Tjalling-dejong:
@DirkEilander Yes, the  eval is now only used on .ini files

config.toml
DirkEilander, line 1: Be careful with the files you commit.

hydromt/config.py
```diff
@@ -3,15 +3,16 @@
...
+from ast import literal_eval
+import codecs
+import abc
+from configparser import ConfigParser
+from warnings import warn
+
+import yaml
 
 
 def configread(
```
DirkEilander, line 18: This function has become a bit hard to read. I've splitted the elements into multiple functions to make it more clear what happens.

DirkEilander (approved):
Thanks @Tjalling-dejong. In general the commit looks good. I've splitted some functions to make it a bit more readable. Please have a last look and merge if you're fine with my edits.

hydromt/config.py
Tjalling-dejong: Thanks! Looks good to me.

PR #294, opened 2023-03-02
BUG: fix error if crs property of Model.grid is None

### LINKED ISSUE #293: BUG: gridmodel CRS gives error if None
If the model grid crs is not set the crs property of the model gives a unreadible error, but should return None.
Also, the crs property should trigger reading the grid if not yet defined.

DirkEilander:
closes #293


Tjalling-dejong (approved):
Looks good to me

PR #304, opened 2023-03-08
Test Python 3.11

Hofer-Julian:


Hofer-Julian:
I've tried to remove all limitations, but unfortunately it still fails.
I'll close this PR for now.

PR #305, opened 2023-03-15
Improved `flw.d8_from_dem` method

DirkEilander:


savente93:
@DirkEilander are you still interested in this functionality? If so I'd urge you to try to get this merged because the further we get with v1 the harder it will be to integrate down the line

DirkEilander:
> @DirkEilander are you still interested in this functionality? If so I'd urge you to try to get this merged because the further we get with v1 the harder it will be to integrate down the line

Thanks for the reminder. Yes, this functionality should land in HydroMT. I'll try to squeeze it in some time soon

hydromt/flw.py
```diff
@@ -89,10 +89,13 @@ def flwdir_from_da(
 
 def d8_from_dem(
     da_elv: xr.DataArray,
-    gdf_stream: Optional[gpd.GeoDataFrame] = None,
     max_depth: float = -1.0,
-    outlets: str = "edge",
+    outlets: Literal["edge", "min"] = "edge",
```
hboisgon, line 93:
```suggestion
    outlets: Literal["edge", "min", "idxs_pit"] = "edge",
```
In pyflwdir, the docs for outlets is still "edge" or "min" and not "idxs_pit". Seems in the code though only "min" is checked so I guess in a way this works

hboisgon (approved):
Nice that we can use something else than upstream area to burn rivers in the DEM! 
PR looks good, just one small check if ``outlets`` = ``idxs_pit`` that this indeed works with pyflwdir but else can be merged :)

DirkEilander:
> Nice that we can use something else than upstream area to burn rivers in the DEM! PR looks good, just one small check if `outlets` = `idxs_pit` that this indeed works with pyflwdir but else can be merged :)

I wanted to make it more explicit here what option is selected. I should do the same in pyflwdir. But it already works because in pyflwdir I only check for the outlets="min" case.

PR #311, opened 2023-03-30
FIX: force writing crs and nodata attributes in interpolate_na 

### LINKED ISSUE #309: BUG: interpolate_na function loses crs information
### HydroMT version checks

### Reproducible Example

when running an update of setup_lulc in hydromt_wflow using a landuse map with a utm coordinate system, I noticed that the crs was lost in the step interpolate_na of landuse. 

I initially fixed it in hydromt_wflow 

```
    da_crs = da.raster.crs
    da = da.raster.interpolate_na(method="nearest")
    da.raster.set_crs(da_crs)
```

but @hboisgon correctly pointed out that this should be fixed in interpolate_na of hydromt-core. 
it is probably related by the creation of an empty array which then loses the crs.

### Issue Description

see above. not sure if the landuse map is needed to fix the error, but I can provide.

### Installed Versions

sorry not the last version of hydromt-core, but just made sure the code has not changed for interpolate_na and this is the case. 
the version I used to test is still:  0.6.1.dev0

### Additional context

_No response_

DirkEilander:
fixes #309

### DIFF
--- a/hydromt/raster.py
+++ b/hydromt/raster.py
@@ -1850,6 +1850,8 @@ class RasterDataArray(XRasterBase):
             data=interp_data,
             attrs=self._obj.attrs,
         )
+        interp_array.raster.set_nodata(self.nodata)
+        interp_array.raster.set_crs(self.crs)
         return interp_array
 
     def to_xyz_tiles(


DirkEilander:
@laurenebouaziz can you check if this solves your issue and review the PR? From the code and without a reproducible example I don't really understand why this happens.

DirkEilander:
ping @laurenebouaziz

DirkEilander:
@hboisgon or @laurenebouaziz could you review this?

hboisgon (approved):
I also cannot reproduce easily without the data. My guess is this happened with a data for which crs info was stored in GEO_MAP_COORD rather than in attrs. If that is the case then this should work. Looking at the fix @laurenebouaziz previously implemented in hydromt_wflow [PR#163](https://github.com/Deltares/hydromt_wflow/pull/163/commits/641a75794902c93b051ac1574c448c3754530864) then I think this should indeed solve it in her case.

PR #313, opened 2023-04-03
ENH: support data like objects in DataCatalog.get methods

### LINKED ISSUE #312: ENH: enable in-memory data to be passed to data source arguments in setup methods
### Feature Type

- [ ] Adding new functionality to HydroMT

- [X] Changing existing functionality in HydroMT

### Enhancement Description

To be able to use the setup methods with data that has been created/modified in memory instead of only file based data. 

### Feature Description

Currently the *_fn arguments in setup methods require a data catalog source name or a file path which are then parsed by the specific DataAdapter to obtain an uniformized data object (e.g., xarray dataset).  By adding the option to pass actual in memory data of the same type as expected we can open up all setup methods for usage with in memory data. This would simply be an addition check at the level where we now distinguish between data catalog source names and file paths.

PS. I know I've been opposed to this idea before, but if we can fix it centrally in the DataAdapter I think there are many benefits which outweigh the ugly multitude of possible data types.

### Additional Context

This would largely benefit e.g. HydroMT-SFINCS where the option to pass in-memory data to the setup methods is required. Currently we are solving this with a second create function which raises confusion with  users and requires a lot of duplicate docstrings.

DirkEilander:
- fixes #312


DirkEilander:
@B-Dalmijn or @Tjalling-dejong could one of you review this PR? See attached issue for more background.

Tjalling-dejong (commented):
LGTM 👍

Tjalling-dejong (approved):

Tjalling-dejong (approved):

DirkEilander:
failing tests are the result of pandas 2.0 in combination with xarray. This shoud be resolved soon at xarray see https://github.com/pydata/xarray/issues/7716

PR #333, opened 2023-05-01
ENH: add generic grid method

hboisgon:
Add generic grid methods:

- setup_grid
- setup_grid_from_constant
- setup_grid_from_raster
- setup_grid_from_raster_reclass
- setup_grid_from vector

For vector added a possibility to rasterize the fraction of area of grid cell intersecting with grid rather than rasterizing a column of vector via a new function raster.rasterize_geometry
Update notebook example on grid model with some of the new setup methods

TODO:

- [x] make more clear from the dosctrings these are generic method
- [x] check if we can cancel inheritance of some of the methods (setup_grid) - or just overwrite in the plugin or error message?
- [x] optional: in setup_grid use align option to align with the resolution (bool) + if res is None then use res/coords of hydrography directly
- [x] setup_grid: use da_like.raster.box instead of vector_grid.total_bounds
- [x] variable grid from setup_grid renamed to mask (or add option to add or not the mask variable)
- [x] move some steps inside the generic setup methods to workflows so that they can be re-used by plugins
- [x] rename rmdict arg to rename


hydromt/raster.py
```diff
@@ -1175,6 +1175,97 @@ def rasterize(
         da_out.raster.set_attrs(**self.attrs)
         return da_out
 
+    def rasterize_geometry(
```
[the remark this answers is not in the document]
hboisgon: Thanks for your comments! I can change the name of the function indeed to make it more explicit.
I'm not sure I fully understand your comment on ``vector_grid`` method. I guess for computing polygon area, what matters is the intersection with the grid so should be fine if grid is a polygon or line type. For fraction this is a little more delicate as I need to make sure fraction <1 so the way cell area is calculated should match calculating the intersected polygon area. Maybe ``raster.area_grid`` would still work without needing the polygons?
In any case seems like it would be nice to commit your changes to this branch so that I can test and update if my new ``rasterize_polygon_area`` method still work with your changes!

hboisgon:
On TODO item:  check if we can cancel inheritance of some of the methods (setup_grid) - or just overwrite in the plugin or error message?

I don't think this is really possible with python class inheritance. I think actually the right way would be to overwrite this method in the subclasses (eg for Wflow change setup_basemaps to setup_grid when switching to GridModel)

hydromt/raster.py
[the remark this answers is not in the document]
DirkEilander: I've pushed the improved vector_grid method to #351

hboisgon:
Hi @DirkEilander, this PR is ready to be reviewed with all the new changes implemented. I struggled way to much only to find that I should use np.linspace rather than np.arange to create a grid based on bbox...

hydromt/workflows/grid.py
```diff
@@ -0,0 +1,309 @@
...
+    grid_like: Union[xr.DataArray, xr.Dataset],
+    ds: xr.Dataset,
+    variables: Optional[List] = None,
+    fill_method: Optional[str] = None,
+    reproject_method: Optional[Union[List, str]] = "nearest",
+    mask_name: Optional[str] = "mask",
+    rename: Optional[Dict] = dict(),
+) -> xr.Dataset:
+    """
+    Prepares data from resampled to grid_like object.
```
DirkEilander, line 81:
```suggestion
    Prepares data by resampling ds to grid_like.
```

hydromt/models/model_grid.py
```diff
@@ -129,6 +393,193 @@ def __init__(
...
+                region.update(basin_index=self.data_catalog[basin_index_fn])
+            # get basin geometry
+            geom, xy = workflows.get_basin_geometry(
+                ds=ds_hyd,
+                kind=kind,
+                logger=self.logger,
+                **region,
+            )
+            # get ds_hyd again but clipped to geom, one variable is enough
+            ds_hyd = self.data_catalog.get_rasterdataset(
```
DirkEilander, line 512: For clarity rename `ds_hyd` to `da_hyd` as this is a DataArray

hydromt/models/model_grid.py
```diff
@@ -129,6 +393,193 @@ def __init__(
...
+            crs=geom.crs,
+            lazy=False,
+        )
+        # Create geometry_mask with geom
+        grid = grid.raster.geometry_mask(geom, all_touched=True)
+        grid.name = "mask"
+        # Remove mask variable mask from grid if not add_mask
+        if not add_mask:
+            grid = grid.to_dataset()
+            grid = grid.drop_vars("mask")
```
DirkEilander, line 574: Can you not skip the geometry mask step if `add_mask == False`?

DirkEilander (commented):
This is a really nice peace of work @hboisgon 

I have made a couple of suggestions to the code, especially in setup_grid method. One more general comment is on the names of the methods like grid_from_raster. I wonder how clear this is, see also suggestions.

hboisgon:
Merging after addressing all comments/suggestions from Dirk's review

PR #339, opened 2023-05-09
ENH: Add pre-commit hooks and add their fixes for all files

### LINKED ISSUE #172: setup pre-commit hooks
https://pre-commit.com/#quick-start

savente93:
This PR adds some pre-commit hooks and applies them to all the files. The diff looks very large, but there are no functional changes, it's all just formatting. I went with a few pre-commit hooks that seemed sensible, but we can add or remove as is necessary. I wanted to also add a linter for the docstrings, but that generated so many warnings that I think it's best to address that in a separate PR, that's why that particular one is commented out in the pre-commit yaml file. 
Resolves #172


DirkEilander (commented):
Thanks for adding this @savente93, this one was long due!

A few questions/remarks:
- Could you add some documentation in docs/dev/contributing.rst about how this works with links to the pre-commit docs.
- I'm not so happy with the default yml pre-commit hook as it is different from the default visual studio code. There seem to be options to configure this, see [readme](https://github.com/jumanjihouse/pre-commit-hook-yamlfmt#configure-pre-commit). Could you change those to match the current yml layout we use.

savente93:
I've taken care of the first two, but the last request I will need some more instructions.
-  There is no default YAML formatting in VSCode as you have to install a formatted as an extension on a new vscode install. I suppose you are probably referring to the Red Hat Yaml LSP as that's the first option when you search for a yaml extention can you conform this is the correct the one, or clarify which one you mean? 
- I could not find any documentation on the used formatting rules of either that extension or any where in the repository. Could you be a bit more specific which rules you want?

savente93:
sorry, I thought I'd done that, must have gotten lost in the force push. I've added it again, so should be good now!

DirkEilander (approved):
Nice addition @savente93! It would be good to also communicate this via email/teams with the team with a link to your explanation in the docs.

savente93:
Good suggestion, thanks!

PR #341, opened 2023-05-10
ENH: support relative paths for Model.read_config

### LINKED ISSUE #316: BUG: ``read_config`` fails silently when providing an incorrect path for ``config_fn`` in reading mode
### HydroMT version checks

### Reproducible Example

```python
from hydromt import Model

root = Path(r"c:\Users\herman\some\path")

mod = Model(root=root, mode="r", data_libs="artifact_data")

config_fn = r"example_imod.ini"
mod.read_config(config_fn=config_fn)

assert mod.config != {}
```

Throws an ``AssertionError``

### Issue Description

``read_config`` fails silently when providing an incorrect path for ``config_fn`` in reading mode.

I didn't notice I had to provide the full path instead of the filename.
Issue is fixed with:

```python
mod.read_config(config_fn= root / config_fn)
```

### Installed Versions

0.7.0

### Additional context

_No response_

savente93:
resolves #316 
Preference order is: 
1. check if file exists at exact path
2. try to convert `config_fn` to an absolute path and check if file exits there
3. check if model root is set and if so if the file is there

I was wondering if we should silent fail if no config can be read as mentioned in the issue, but that should probably be discussed since I'm not sure if that can have other consequences.

### DIFF
--- a/hydromt/models/model_api.py
+++ b/hydromt/models/model_api.py
@@ -672,6 +672,12 @@ class Model(object, metaclass=ABCMeta):
             if isfile(config_fn):
                 cfdict = self._configread(config_fn)
                 self.logger.debug(f"{prefix} config read from {config_fn}")
+            elif isfile(abspath(config_fn)):
+                cfdict = self._configread(abspath(config_fn))
+                self.logger.debug(f"{prefix} config read from {abspath(config_fn)}")
+            elif self.root is not None and isfile(join(self.root,config_fn)):
+                cfdict = self._configread(join(self.root,config_fn))
+                self.logger.debug(f"{prefix} config read from {join(self.root,config_fn)}")
             elif not self._read and prefix != "Default":  # skip for missing default
                 self.logger.error(f"{prefix} config file not found at {config_fn}")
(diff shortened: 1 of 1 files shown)


DirkEilander (changes requested):
Nice first PR @savente93 !
I would like to reverse 2 (abspath) and 3 (model root) in your order of checks as 3 will be the more common one.
For option 3 we could also add a check on wether config_fn is not an absolute path to avoid possible errors downstream in os.path.join if we give two absolute paths (not sure what is does in that case)
Finally, the last elif can be an else-statement

DirkEilander (changes requested):
Nice first PR @savente93 !
I would like to reverse 2 (abspath) and 3 (model root) in your order of checks as 3 will be the more common one.
For option 3 we could also add a check on wether config_fn is not an absolute path to avoid possible errors downstream in os.path.join if we give two absolute paths (not sure what is does in that case)
Finally, the last elif can be an else-statement

DirkEilander (approved):
Great first PR @savente93. It's ready to merge!

PR #342, opened 2023-05-10
ENH: Add Dockerfile 

### LINKED ISSUE #248: Deploy HydroMT docker image to dockerhub

savente93:
Doesn't quite resolve  #248 but makes a significant step. Once this gets merged we can star looking at publishing the image. Wasn't quite sure what the use case was going to be (although it's sure to be useful for cloud integration) so for now it's just a minimal image. We can always add more parts if necessary


DirkEilander:
@Jaapel Could you review this?
It would also be nice to explore how we can use this file to deploy a docker image to docker hub (could be done in a different PR).

savente93:
I've added a github action to deploy the docker image to dockerhub. not sure whether we need to configure credentials for the deltares dockerhub account, but baring that it should work. I tested it with my own dockerhub account and that went fine.

Jaapel:
Nice work @savente93.
What I would add is a mechanism to consistently tag the image with the HydroMT version.
I think the tag is `deltares/hydromt`, which I believe will push to latest. We run this workflow when releasing, so it is best to tag is with the release tag.

savente93:
Like so?

PR #351, opened 2023-05-16
ENH: improve raster.vector_grid

### LINKED ISSUE #347: ENH: improve speed of raster.vector_grid method.
### Feature Type

- Changing existing functionality in HydroMT

### Enhancement Description

The current vector grid implementation is slow. This is used to visualize the grid as well as for some GIS methods, hence speed is important.

### Feature Description

This can be enhanced by avoiding looping over each cell. In addition, instead of a polygon per cell only the lines of the cell edges could be returned. 

### Additional Context

_No response_

DirkEilander:
fixes #347


hboisgon:
Hi @DirkEilander can you still check why the tests are failing before I review? Seems to be linked to your changes.

hboisgon:
Tests are failing because the order of the generated grid polygon has changed compared to previous method

hboisgon (approved):
Thanks @DirkEilander I tried with my rasterize_geometry method in PR#333 and it really is much faster (minutes even depending how big the grid is!)

PR #354, opened 2023-05-22
Fix deprecated numpy function

JoostBuitink:
Since numpy 1.20.0, aliases of builtin types is deprecated. In most cases, this was not a problem, but one line in `write_map` used np.bool. This functionality is now removed from numpy, and is failing the hydromt_wflow tests. This PR fixes this issue, by changing `np.bool` with `bool` and `np.bool_`. See https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations


hboisgon (approved):
Thanks @JoostBuitink !

PR #369, opened 2023-05-30
Fix warnings in test suite

### LINKED ISSUE #357: BUG: Test suite uses depricated geopandas dataset module
### HydroMT version checks

### Reproducible Example

```
pytest
```

### Issue Description

The testsuite uses a deprecated part of geopandas raising the following issue: 
```
FutureWarning: The geopandas.dataset module is deprecated and will be removed in GeoPandas 1.0. You can get the original 'naturalearth_lowres' data from https://www.naturalearthdata.com/downloads/110m-cultural-vectors/.
```
This should be fixed, as we should not depend on depricated software. 
The data should have moved to https://github.com/geopandas/geodatasets so all that really needs to be done is figure out how to pull it from there. Can also be integrated with other data caching work, but is not necessary

### Installed Versions

0.7.2.dev0

### Additional context

_No response_

savente93:
Running the test suite throws a lot of errors, which makes it easy to miss genuine problems and [footguns](https://en.wiktionary.org/wiki/footgun). This PR addresses _most_ of the errors in the test suite, including the geopandas deprecation warning (i.e. Fixes #357 ) Since the files were only a couple 100 KB total, I though it was fine to just add then to the repo. If people have a problem with this let me know and we'll find a solution. There are two warnings that remain unsolved (relating to `Converting a CFTimeIndex` and `Variable(s) referenced in cell_measures not in variables` respectively) but these will require more work, so I thought it'd be best to address those in their own PR down the line.


hydromt/io.py
```diff
@@ -389,11 +389,14 @@ def open_timeseries_from_table(
     """
     kwargs0 = dict(index_col=0, parse_dates=True)
     kwargs0.update(**kwargs)
+    if "date_format" not in kwargs0.keys():
+        kwargs0["date_format"] = "%Y-%m-%d"
```
hboisgon, line 393: This is a bit risky. What are the inconvenients of letting pandas try to guess format in pd.to_datetime()?

hydromt/raster.py
```diff
@@ -1697,8 +1707,9 @@ def set_nodata(self, nodata=None, logger=logger):
                 nodata = self._obj.rio.encoded_nodata
         # Only numerical nodata values are supported
         if np.issubdtype(type(nodata), np.number):
-            self._obj.rio.set_nodata(nodata, inplace=True)
-            self._obj.rio.write_nodata(nodata, inplace=True)
+            # cast to float since using int causes inconsistent casting
```
hboisgon, line 1710: what do you mean by this? Some dataarray are of type type or uint and the nodata value should be same type as array values type.

hydromt/raster.py
```diff
@@ -1257,6 +1261,10 @@ def rasterize_geometry(
         nodata : int or float, optional
             Used as fill value for all areas not covered by input geometries.
             By default -1.
+        keep_geom_type : bool
+            Only maintain geometries of the same time if true, otherwise
```
hboisgon, line 1265:
```suggestion
            Only maintain geometries of the same type if true, otherwise
```

hydromt/raster.py
```diff
@@ -1240,6 +1243,7 @@ def rasterize_geometry(
         mask_name: Optional[str] = None,
         name: Optional[str] = None,
         nodata: Optional[Union[int, float]] = -1,
+        keep_geom_type=False,
```
hboisgon, line 1246:
```suggestion
        keep_geom_type: Optional[bool] = False,
```

pyproject.toml
```diff
@@ -31,6 +31,7 @@ dependencies = [
     "rasterio",
     "requests",
     "rioxarray",
+    "ruff",
```
hboisgon, line 34: ruff is now twice in the list.
I didn't review this but aren't ruff and pre-commit more developpers dependencies? So to exclude from a conda-forge install?
I also see there is no real 'developper' or dev install with basically all+test+doc

hboisgon (commented):
Nice work at @savente93 ! And yes good to be able to easily track relevant warnings.
I added some comments / questions.
For naturalearth data from geopandas, I guess fine for now having the data here but ideally we can access this data via data catalog later. Or else move it to the data artifact folder of hydromt: https://github.com/DirkEilander/hydromt-artifacts
Maybe could you save it as geojson rather than shapefile? Then just it's just the one file that is added/tracked

pyproject.toml
savente93: You're right about it being in there twice, that was my mistake. As for the second part of your comment I think you're correct as well. Previously there wasn't really such an option. I'll try to restructure them a bit to account for this

hydromt/raster.py
savente93: Well, rioarray generates this warning:

```
tests/test_model.py::test_gridmodel_setup
  /home/sam/.micromamba/envs/hydromt/lib/python3.11/site-packages/rioxarray/raster_writer.py:130: UserWarning: The nodata value (0.01) has been automatically
changed to (0.009999999776482582) to match the dtype of the data.
``` 
So the conversion is already happening so I think we might as well be explicit about it. I think it also has to do with the fact that `float` has built in for `inf` and `nan` whereas `int` doesn't which is causing these implicit conversions in a lot of unexpected places I think (see below) 

```Python
Python 3.11.3 | packaged by conda-forge | (main, Apr  6 2023, 08:57:19) [GCC 11.3.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.14.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: import numpy as np

In [2]: type(np.nan)
Out[2]: float

In [3]: np.float64('inf')
Out[3]: inf

In [4]: np.float64('nan')
Out[4]: nan

In [5]: np.int32('inf')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[5], line 1
----> 1 np.int32('inf')

ValueError: invalid literal for int() with base 10: 'inf'

In [6]: np.int32('nan')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 np.int32('nan')

ValueError: invalid literal for int() with base 10: 'nan'

In [7]:
```

hydromt/io.py
savente93: Letting Pandas guess at the format generates the following warning: 
```
UserWarning: Could not infer format, so each element will be parsed individually, falling back to `dateutil`. To ensure parsing is consistent and as-expected, please specify a format.
``` 
Aside from being very bad for performance, this means that you could end up with inconsistent datatype in your column, which is obviously not something you want. In addition this could be a problem for reproducibility, since changes in how pandas makes these guesses can change between minor versions leading to different behavior. Generally I think it is much better to let these things fail explicitly rather than silently because bugs that are caused by this sort of behavior are very difficult to track down. It might be somewhat inconvenient for users, but it is a problem they only have to solve once, and it would protect us from major headache down the line, therefore I think this is worth it.

savente93:
> For naturalearth data from geopandas, I guess fine for now having the data here but ideally we can access this data via data catalog later.
Absolutely, this is something I want to work on down the line, for now I think this is an acceptable short term solution. 

> Maybe could you save it as geojson rather than shapefile? Then just it's just the one file that is added/tracked
I just took it as is since I'm not very familiar with these file types, but indeed having at one file would be better. I'll see if I can make the conversion, but I might ask for your help on this.

hydromt/io.py
hboisgon: I agree in principle but then I would at least throw an explicit warning with it and try to already correct the warning by updating date_format upstream (data catalog or test data).

Something else which is quite ugly and on our TODO list is that the user currently has no way to pass **kwargs to that function... It is is usually used through io.open_geodataset which only supports so far direct filepath rather than data catalog entries in which you could specify kwargs for pandas and the timeseries table.... so unless we allow the user to pass kwargs I would not implement that change yet.
Not too sure how much work it is though so this might require its own PR.

hydromt/raster.py
hboisgon: You can convert for float but I prefer to have the error for int. Some models have strict requirements on data types so I prefer the user to have a clear error message and set a correct nodata value rather than converting the dtype of the data.

savente93: After giving it some more thought, it seems that it's mostly the native python types that are not being very nice, so I've added some checks to convert those into the appropriate numpy types and otherwise just leave it as is, I think that should be a good compromise

hydromt/io.py
savente93: Do we have an issue for this?

savente93: As you say this might require a bit more than is appropriate for this PR. Especially given that there are some other warnings I've not touched I think it's fine to leave this like this for now. I've reverted this bit back to the original version. I'll make an issue about this

hboisgon (approved):

Issue #366, opened 2023-05-26
Maintenence: Unify redundant enviroments


savente93:
There are many places where environment specs are kept in HydroMT, at best this leads to more maintenance, at worst it introduces bugs constantly. A solution should be found that allows for a single source of truth

PR #370, opened 2023-05-30
Unify enviroments 

savente93:
The different environments for the documentation, testing, developing, and docker are difficult and error prone to maintain. Therefore this PR introduces a way to generate all of the necessary environments from the `pyproject.toml`. This way that is the only thing that needs to be updated, and will be a single source of truth when it comes to dependencies. This will also make docker support vastly easier and more maintainable to the point where I consider it a requirement. I've also updated the actions to use the new script which all seems to go pretty well.


savente93:
Fixes #366

savente93:
All is not a target, the options are  `{full,test,doc,min}` if it's about `min` I think it's fine, if it's about `full` then we should add them. Which one are you talking about? 

I think the `all` table in the pyproject.toml refers to the fact that it should be added to all environments, not that it includes all dependencies. At leas that's how I read it. What do you thtink?

Jaapel:
I really like using the pyproject.toml as a main requirements file!

savente93:
You were right, the full spec was indeed missing two packages. I've added them and locally all tests pass now so once the GHA are done we should be ready to roll

Jaapel:
I was talking about full, for me that implies that all possible dependencies are included.
I will check running tests with `python make_env.py test` to see if no tests are skipped.

Jaapel (approved):

Issue #321, opened 2023-04-17
DOC: update user stories and publications 


DirkEilander:
### HydroMT version checks

- [X] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/hydromt)

### Location of the documentation

https://deltares.github.io/hydromt/latest/getting_started/user_stories
https://deltares.github.io/hydromt/latest/getting_started/publications.html

### Documentation problem

The figures of the compound flood model case are outdated and the publication list does not contain our JOSS paper and refers to the preprint version of the Eilander (2022) paper.

### Suggested fix for documentation

The outdated figures and references and should be replaced by https://nhess.copernicus.org/articles/23/823/2023/
The JOSS  paper is linked in the readme and should be added to the publications.
Potentially there are more publications that can be added to the list.

Issue #300, opened 2023-03-07
DOC: table layout HydroMT data catalog unreadible


sibrenloos:
Layout of data catalog on wiki (https://deltares.github.io/hydromt/latest/user_guide/data_existing_cat.html) is no longer readible:

![TableLayout_unreadible](https://user-images.githubusercontent.com/75011151/223392256-9d50c71f-4c3d-4f91-ba23-dfa9297fb51b.png)

savente93:
This seems to no longer be the case (when you follow the link) should this issue be closed?

savente93:
ah my bad, I didn't think about that (it does show correctly on firefox). please disregard my previous comment

DirkEilander:
I think this could be solved by removing the path argument, or only showing the filename without the path (which isn't relevant info anyway). The table is created in docs/conf.py where this issue should be addressed.

Issue #299, opened 2023-03-07
DOC: DataFrame doc is missing from the preparing a data catalog page


hboisgon:
On page: https://deltares.github.io/hydromt/latest/user_guide/data_prepare_cat.html
DataFrame and the different driver are not listed in the available options for data_type and driver.

Issue #251, opened 2022-12-02
update logo to match deltares blue 


DirkEilander:
See attached logo kindly updated by @JoerivanEngelen 
![HydroMT_logo](https://user-images.githubusercontent.com/15379728/205268599-742c48d1-c3c2-4849-8de0-ff38d91eb912.svg)

PR #373, opened 2023-06-02
Fix some documentation warnings and formatting issues

### LINKED ISSUE #371: DOC: fix small mistakes in developper doc code format
### HydroMT version checks

https://deltares.github.io/hydromt/latest/dev/contributing.html#code-format

### Documentation problem

Link to pre-commit and first code block

image

Fix ink with _ at the end and fix code block indent

savente93:
Fixes #371 along with some other warnings generated when generating the docs that should help keep the docs clean. There are a few other warnings that still get generated that I haven't managed to address yet, which I would like to get back to at some point, but I figured it's better to just solve part of these things and move on to more urgent matters. Sorry for the huge number of files changed, but they are docstring changes only so hopefully it won't take too much time to review.


savente93:
also Fixes #244

savente93:
should also Fix #300

savente93:
fixes #251

hboisgon (changes requested):
Thanks for the changes @savente93 I corrected a few mistakes in a commit and added all developers in the team.

There is still one thing that is not fully fixed. In the pre-defined data catalog table now the indent looks good but the '/' in the path are now missing:
image

hboisgon:
@savente93 do you think you could also adress #299 and #321 here?
They are also very small doc issues.

savente93:
Something very strange is going on here. It is all generated correctly on my machine (see screenshot) 
image
Did you generate them locally? if so were there any relevant warnings?

savente93:
Fixes #299 
Fixes #321

hboisgon:
> Something very strange is going on here. It is all generated correctly on my machine (see screenshot) image Did you generate them locally? if so were there any relevant warnings?

Strange indeed. I generated the docs locally yes but don't remember any particular warning. Maybe Ubuntu vs Windows?

hboisgon (commented):
Nice! I'll update the pictures and then just left is to figure out if we can why the data catalog tables prints differently. We generate the docs using Linux github actions so maybe not too urgent

hboisgon:
Well also not fully clear to me which figures and I guess text modifications are needed. I'll let @DirkEilander change this directly

savente93:
I've generated the docs both on linux and on windows and still cannot reproduce the problem. Can you pull the latest version and regenerate it in a new environment? sometimes code from older versions can make it's way into the docs because it was installed in conda

PR #380, opened 2023-06-02
Dockerfile cleanup

### LINKED ISSUE #379: BUG: Dockercontainer doesn't run tests before publishing
### HydroMT version checks

### Reproducible Example

.

### Issue Description

the current GHA that publishes the docker contianer doesn't actually run the tests to make sure they all work, which could lead us to publishing a faulty container. This should be adressed

### Installed Versions

current

### Additional context

This was mentioned by @DirkEilander during a conversation. I don't think we ever made an issue for it at the time, so I'm doing this retroactively so I can link it in the PR addressing it.

savente93:
I've cleaned up the docker file a bit, make use of the new environment generation to make sure that it is up to date with the rest of the project, and I've restructured the github actions ever so slightly so they can be tested locally using [act](https://github.com/nektos/act) without fear of actually publishing something. Fixes #379


savente93:
Note to self: This PR is spiraling a little bit out of control scope wise. Trim it back to just fix the docker testing before publishing. Things like binder, different versions and ci caching should happen in seperate PRs

.github/workflows/publish-docker.yml
```diff
@@ -11,29 +11,41 @@ on:
...
         uses: docker/login-action@v2
         with:
           username: ${{ secrets.DOCKERHUB_USERNAME }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
-      - name: Docker meta
-        id: meta
-        uses: docker/metadata-action@v4
+
+      - name: Build and test
+        uses: docker/build-push-action@v4
```
Jaapel, line 33: Why two times the build and push action?

Dockerfile
```diff
@@ -1,15 +1,57 @@
...
+ && find /opt/conda/ -follow -type f -name '*.a' -delete \
+ && find /opt/conda/ -follow -type f -name '*.pyc' -delete \
+ && find /opt/conda/ -follow -type f -name '*.js.map' -delete
+
+
+FROM  mambaorg/micromamba:1.4-jammy AS cli
+ARG NB_USER=hydromt
+ARG NB_UID=1000
+COPY --from=dev /opt /opt/
+COPY --from=dev /home/hydromt /home/hydromt/
```
Jaapel, line 44: great use of multi-stage builds!

.github/workflows/publish-docker.yml
savente93: yes. The first is to build, then we run the test tuite and the second is to push it once the test suite has passed. The building should be cached between the two actions. See example here https://docs.docker.com/build/ci/github-actions/test-before-push/

Jaapel: Ah so if you do not add push in the `with` options, it just builds, good to know

PR #381, opened 2023-06-02
Create CONTRIBUTING.md

savente93:
Might not seem like much, but often people interested in getting involved will look for this, in addition to some automated tools, so it's easy bonus points.


hboisgon:
Nice small addition I guess! Is it possible to make it a rst file like the rest of our docs and the README?

savente93:
Good point!

alimeshgi (approved):

Issue #383, opened 2023-06-02
Previous CI flows should be canceled on new commit


savente93:
The CI workflows could be better optimised, improving on cose as well as developer time. Optimisations that come to mind now: 

1. canceling current workflows if they are rerun
2. ~~using common worklows that can be cached for things like installing~~
3. ~~Not running all CIs at once, (e.g. don't run the tests if the linting doesn't pass.)~~

savente93:
Upon further reflection, I think all of these points should be in their own issue. This issue will focus on just the first one, I'll make seperate issues for the other ones.

PR #386, opened 2023-06-07
Introduce makefile for easier build management

### LINKED ISSUE #387: Use Makefile more extensively for build management
### Kind of request

Adding new functionality

### Enhancement Description

Currently Hydromt has quite a few build processes that need to be managed separately, including:
1. environment installation
2. building documentation
3. building & pushing docker image
4. building & publishing the pypi package

Some of these processes have their logic defined in the GitHub actions only, such as the building of the PyPi artifacts which makes local testing and debugging hard. 

There is already a `Makefile` in the docs folder that manages the building of the documentation, but this could be extended to help manage the other build systems as well. This would provide a more unified entry point into hydromt which would be beneficial for users and maintainers alike. 

### Use case

installing the required environment would be done by simply running `make env`. Docs could be build by running `make html` or `make docs` the pypi artifacts could be built with `make pypi` and the docker images could be build with `make docker`. There would also be a `make clean` to clean up all residual artifacts. 

### Additional Context

This will also help significantly for the coming cloud infrastructure.

savente93:
## Issue addressed
Fixes #387 

## Explanation
I took the makefile from the docs folder and extended it so that it will work from the root folder, in addition to adding the features mentioned in the issue. For this Initial PR I decided not to change anything, but simply to move current logic into the makefile to make for an easier review. If/when this gets merged, I'll open additional PRs afterwards to improve on the logic where necessary. 

## Checklist

This will be extremely useful for optimizing the CI workflows as well. see #383 for additional info


Jaapel (approved):
Good that you also added Make installation to the docs, including windows install. This makes local testing of the workflow on github actions easier as well.

PR #399, opened 2023-06-13
Fix binder integration

### LINKED ISSUE #391: Binder is broken
### HydroMT version checks

### Reproducible Example

:

### Current behaviour

Currently binder is broken because of changes that got merged in #370 This should be restored. If possible it would be good to integrate it with the docker image so it can just use that and reduce the startup time. 

### Desired behaviour

binder works again. 

### Additional context

_No response_

savente93:
## Issue addressed
Fixes #391 

## Explanation
1. I've split the first Dockerfile in two: one for just hydromt and one for the binder to build, that just imports the main hydromt and adds the requirements on top. This is better for caching as well as keeping the main Dockerfile much cleaner. 
2. I've added a workflow that tests the docker image in the suite as well as with the mybinder integration, This will help to make sure we(/I 👼 ) don't break this functionality again. 

## Checklist

Dockerfile setup is beter but could still do with a bit of clean up, this should happen after the environments are restructured so it can pick what it needs more effectively.


.binder/Dockerfile
```diff
@@ -0,0 +1,25 @@
+ARG NB_USER=hydromt
+ARG NB_UID=1000
+
+# currently we haven't published an official hydromt image yet, so for now
+# I import from an export I've put on my own account. this should be moved
+# to the official image as soon as it's published.
+FROM savente/hydromt:current as binder
```
DirkEilander, line 7: what is the current plan for publishing hydromt on the deltares dockerhub account?

DirkEilander (approved):
nice work @savente93!
Good to see the binder image is tested and curious to see what the gained speedup is with binder.

savente93:
@sam https://forum.orfeo-toolbox.org/t/zonal-statistics-corrupted-size-vs-prev-size/672/4 see link, might be useful for the corruption error?

Issue #406, opened 2023-06-15
Very minor spelling erros in documentation


Schuddebuikje:
### HydroMT version checks

- [X] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/hydromt)

### Kind of issue

Docs are wrong

### Documentation problem

[https://deltares.github.io/hydromt/latest/user_guide/intro.html](url)
Under "Methods and workflow" there is a broken reference

[https://deltares.github.io/hydromt/latest/_examples/prep_data_catalog.html#GeoDataset-from-a-netcdf-file](url)
Under "GeoDataset from a netcdf file", "open" should be "opened"

[https://deltares.github.io/hydromt/latest/_examples/export_data.html#Explore-the-current-data-catalog](url)
Under "Explore the current data catalog", "with" should be "which"

[https://deltares.github.io/hydromt/latest/_examples/reading_raster_data.html#Raster-driver](url)
Under "Raster driver", "of" should be "or"

[https://deltares.github.io/hydromt/latest/_examples/working_with_tiled_raster_data.html](url)
Multiple times on this page, "a xarray" should be "an xarray"

[https://deltares.github.io/hydromt/latest/_examples/working_with_tiled_raster_data.html#reading-tiled-raster-data-with-zoom-levels](url)
Under "reading tiled raster data with zoom levels", "retreived" should be "retrieved"

[https://deltares.github.io/hydromt/latest/user_guide/model_overview.html#high-level-functionality](url)
Under "High level functionality", "websites" should be "website"

Issue #403, opened 2023-06-14
Error in hyperlink to a figure under "Methods and workflow" of "User Guide"


alimeshgi:
### HydroMT version checks

- [X] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/hydromt)

### Kind of issue

Docs are wrong

### Location of the documentation

https://deltares.github.io/hydromt/latest/user_guide/intro.html

### Documentation problem

There is an error in hyperlink to a figure under "Methods and workflow" of "User Guide".  It is written as follows: Most of the heavy work in HydroMT is done by Methods and workflows, indicated by the gear wheels in autoref{fig:schematic}. 

### Suggested fix for documentation

In the above sentence, autoref{fig:schematic} needs to be replaced by a proper hyperlink to the figure.

PR #407, opened 2023-06-16
Restructure envs

### LINKED ISSUE #394: Environment should be restructured.
### HydroMT version checks

### Reproducible Example

:

### Current behaviour

Currently the environment specifications (e.g. all/doc/test) are not structured very well and this needs a bit of an overhaul so other applications can be more targeted in what they need. Some specs that would probably be useful: 
- tests
- examples/jupyter/binder
- cli/dev
- env
- min
- mesh

Feel free to edit as necessary

### Desired behaviour

The environments are better specified so they have fewer overlap and systems can target more specifically what they need. 

### Additional context

Been running into this a lot when trying to restructure the Dockerfile for the various use-cases.

savente93:
## Issue addressed
Fixes #394 

## Explanation
This is a fairly big one so the explanation is going to be a bit longer. Since it also touches on multiple areas, I'll request reviews from multiple people though not all of them have to review all of it. I'll mark each section for the appropriate reviewer.

### Structure of the environments [Dirk/Helene]
There are two concepts here that I've used to structure the dependencies: 
1. groups are "logical" clusters of dependencies that are necessary for each other to function. They are meant to be modular, optional and independent. the ones I've landed on are: 
   a. `io` Reading and writing various formats like excel but also  cloud file systems
   b. `extra` Couldn't think of a better name for this one, but it has some extra for ET and mesh calculations
   c. `dev` everything you need to develop and publish HydroMT
   d. `test` What you need to run the test suite. Test suite should be setup that only tests that use the dependencies that are installed are run, so this should always pass no matter what other dependencies you may or may not have installed.
   e. `doc` generate the docs
   f. `jupyter` Run Jupyter notebooks and run the examples. Going to use this for binder support mostly.
   g. `deprecated`  dependencies that we hope to remove soon, but aren't quite ready to yet. 
2. "flavors" are more or less just collections of one or more groups designed for common use cases. The flavors are: 
  a. `min` no optional dependencies. mostly as a base to build your DIY stack on. 
  b. `slim` Just the operational bits, what most people will probably want and what the cloud will most likely use 
  c. `full`  absolutely everything, useful for developing. 

Of all the combinations I tried I think this strikes the best balance between being fully featured without being bloated for each use case. This was particularly important for the docker images which I'll talk about below. 

### Env management [Jaap/Dirk/Helene]
To manage everything I've started making more use of the make file to keep it all somewhat easy. The docker bits are mostly for me to make sure I (and the CI) don't accidentally build the docker images and the corresponding environments. What might be more useful for users is the `make env` or `make dev` receipts. the `dev` receipt is meant to be an easy way to set up everything you need to get coding.  ( I also use this frequently as well). The `make env` gives a more fine grained access to building the correct environments and can be used to address the groups directly like so `make env OPT_DEPS="extra,io,docs"`. By default it uses micromamba, but you can also adjust that like so `make env PY_ENV_MANAGER='mamba'`. If you do this frequently, you can set it as an environment variable in your shell and it should just pick up on it as well. 

### Docker images [Jaap]

Took quite some finagling, and I went through many different setups, but I managed to get all of the images around 2 GB which I'm really pleased by. I've made docker images for each of the "flavours" (see above). I think these will be good for most use cases. In particular this (in addition to the docs cleanup I did) should make binder much faster to startup. The statements are laid-out to maximize cache hits when only code changes, which will be useful as I intend to use these images to speedup the testing CI as well, but that's a story for another PR. 

(I'll add all of this to the docs as well) 
 

## Checklist


savente93:
Fixes #403

savente93:
Fixes #406

savente93:
I've also taken the opportunity to fix a few small doc mistakes why I was at it.

savente93:
Thanks! If I remember correctly, they were optional dependencies in a category called "all" which is why they are still optional now. I just didn't want to call it call because I thought that would be confusing with the `full` flavour. I'm happy to add them to the main dependencies, and just get rid of the extra category if you want?

savente93:
Given that Jaap is very busy at the moment with GWW I'm going to just merge so we can move on. I want to use this in a couple of other PRs. if anything turns up we'll adjust down the line

DirkEilander:
> Thanks! If I remember correctly, they were optional dependencies in a category called "all" which is why they are still optional now. I just didn't want to call it call because I thought that would be confusing with the `full` flavour. I'm happy to add them to the main dependencies, and just get rid of the extra category if you want?

There's one other thing to note. pyet is currently not distributed as a conda package, making it impossible to include it as a dependency in [our conda-forge recipe](https://github.com/conda-forge/hydromt-feedstock/blob/main/recipe/meta.yaml). Hence, for now I'd like to keep pyet as an optional dependency. The best solution here would be to contact the pyet developers teams and check I've they are open to distribute their package through e.g. conda-forge. I briefly discussed this with one of the main developers at EGU23 and he seemed open to it.

PR #412, opened 2023-06-19
Mesh model

### LINKED ISSUE #326: ENH: update MeshModel.mesh property based on recent xugrid developements
### Feature Type

- Adding new functionality to HydroMT
- Changing existing functionality in HydroMT

### Enhancement Description

Previously xugrid would only allow one grid to be defined in a UgridDataset object and in our first definiton of MeshModel.mesh we assumed one 2D grid definition.
xugrid can now support several grids to be defined in one UgridDataset object so we can update and enhance MeshModel.mesh to also support multi grids (1D/2D/3D).
A decision needs to be taken how to store 1D/2D/3D meshes in the MeshModel.mesh property. 

### Feature Description

**MeshModel.mesh**
We have more or less two options on how to define MeshModel.mesh:

- Option 1 (preferred): ``UgridDataset`` containing all grid definition and variables
- Option 2: ``Dictionnary`` of single grid UgridDataset

For the high-level attributes:
``MeshModel.region`` would be the UgridDatasetAccessor.total_bounds and ``MeshModel.crs`` UgridDatasetAccessor.crs
Problem is that this supposes that all grids in MeshModel.mesh are in the same CRS. We could do a check and else give an error. Not sure how frequent this would be so plugins could decide to re-write these properties if they need to support meshes with grids in different CRS? 

**Selecting a specific grid in MeshModel.mesh**
In order to add new data variable (eg DEM) to MeshModel.mesh, we need to be able to select a specific grid (1D/2D/3D) for data processing. The assumption is that it would be possible to have multiple 1D or 2D or 3D grids so shortcut properties like ``MeshModel.mesh1d`` / ``MeshModel.mesh2d`` and ``MeshModel.mesh3d`` cannot really be added in the core.

The core should then be really flexible on grid selection based on grid_name and maybe a shortcut to easily check if the grid is 1D/2D/3D. In case of option 1, grid name should be able to be derived from the grid topology (grid_name_edge eg mesh1d_edge...) and in case of option 2 this would be the dictionary key. Maybe define a new ``MeshModel.get_mesh``(grid_name) to easily select a specific grid and related data variables.

Conventions and shortcuts on grid_name could be defined in specific MeshModel plugins (eg for Delft3D conventional grid names could be mesh1d, network1d, mesh2d, mesh3d).

**Adding data variables to MeshModel.mesh via setup_ methods**
We can still define generic method to add and resample data (eg DEM) for MeshModel.mesh. The generic method will however require the user to provide a grid_name in order to select the correct grid for resampling.
The names of the method will be updated in case the grid is 1D / 2D or 3D as the GIS methods will differ.

Eg of new/updated setup methods:
``setup_mesh2d_from_raster``(raster_fn, grid_name, ...) would (needs refinement):

- read raster_fn regular gridded external data
- extract from MeshModel.mesh the 2D grid_name
- interpolate raster_fn to grid_name
- add the new data to MeshModel.mesh using MeshModel.set_mesh(interpolated data)

``setup_mesh1d_from_pointvector``(vector_fn, grid_name) would (needs refinement):

- read point vector data
- extract from MeshModel.mesh the 1D grid_name and convert to geopandas.GeoDataFrame
- intersects the points with the 1D grid and add the required data attributes from point to grid
- go back from the GeoDataFrame to 1D UgridDataset
- add the new data to MeshModel.mesh using MeshModel.set_mesh(point_data)

**Adding or updating a grid to MeshModel.mesh**
Compared to a classic GridModel.grid, MeshModel.mesh definition cannot be easily done in one ``setup_mesh`` method. New grids can be added (eg add a 1D grid, a 2D grid) or grid definition could even be updated (eg extend the current 1D grid with additional lines like rivers/pipes etc). 
We need methods to be able to deal with this.

Usually such method would be called ``setup_mesh1d``, ``setup_mesh2d``, ``setup_mesh3d`` and would have a grid_name argument.

- Prepare a 1D/2D/3D grid schematization based on the method 'region' argument
- Call a method to add/update the new grid with name grid_name. 

Classically the last step would be calling MeshModel.set_mesh but I feel we may need a different name as this step would be quite different than just adding data variables (``MeshModel.update_mesh`` instead of ``MeshModel.set_mesh``?):

- In case of adding (new) grid_name, we add the new grid definition and potentially update the links of that grid to other (eg updating link1d2d in Delft3D-FM). So far, updating the links between several grids seems to be plugin specific so can be left out of the core.
- In case of updating an existing grid_name, we then need to replace the grid_name definition in MeshModel.mesh. An additional step here compared to adding is that if data variables for grid_name were already present, then the data is not 'valid' anymore and may need to be re-interpolated to the updated grid. Not fully sure how and if we could support it? A possibility is that when calling update_mesh(UgridDataset, grid_name), if UgridDataset contains all data variables that were already present in MeshModel.mesh for grid_name then we allow overwritting the data variables and else we throw an error. This would leave interpolating the existing data variables to the plugin (eg is setup_rivers, setup_pipes in Delft3D-FM where attributes of the 1D network like roughness and material are also added at the same time as the 1D lines).

### Additional Context

@hboisgon will discuss the best implementation with stakeholders and afterwards refine the issue

hboisgon:
## Issue addressed
Fixes #326

## Explanation
Supports that self.mesh contains several topologies by implementing changes discussed in #326 

## Checklist


hydromt/models/model_mesh.py
```diff
@@ -253,13 +293,120 @@ def set_mesh(
...
+            for grid in self._mesh.ugrid.grids:
+                grid.set_crs(crs)
+
+        # update related geoms if necessary: region
+        if overwrite_grid or new_grid:
+            # add / updates region
+            self._geoms.pop("region", None)
+            self.region
+
+    def get_mesh(
```
xldeltares, line 367: Reminder to further think about the behaviour of this function. 
Is get_mesh a get method wrapper around mesh_grids and mesh_datasets?
If so, why get_mesh(name), given mesh is None or name does not exist, raises error? 
Instead, if a user calls mesh_grids.get(name) or mesh_grids.get(name) they would be return with none by default. 
Is the idea to harmonize the behaviour or distinct (so that user are forced with use the mesh_grids.get() and meshd_datasets.get())?

hydromt/models/model_mesh.py
```diff
@@ -304,17 +451,69 @@ def write_mesh(self, fn: str = "mesh/mesh.nc", **kwargs) -> None:
             os.makedirs(dirname(_fn))
         self.logger.debug(f"Writing file {fn}")
```
xldeltares, line 452: Reminder: allow optional_attributes as argument? or use ugriddataset already in self.mesh?

xldeltares:
Hi @hboisgon, I committed the updates from our discussion. Please have a look and make adaptations where needed.

I believe now the review goes back to you, which might be because of my commits...

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
...
+                        uds[var] = self.mesh[var]
+                # additionnal topology properties
+                elif var.startswith(grid_name):
+                    uds[var] = self.mesh[var]
+                # else is global property (not grid specific)
+
+            return uds
+
+        else:
+            return self.mesh_grids[grid_name]
```
xldeltares, line 414: Hi @hboisgon, I slitely modified this function to return only xu.UgridDataset. Before it could return `Ugrid1d`/`Ugrid2d` (if `include_data` is True, or `UgridDataset` if `include_data` is False). I find that a little inconsistant. Considering when we call self.mesh it returns a `UgridDataset`, I think `get_mesh` returning `UgridDataset` make sense. Then the downstream actions for getting the grid `UgridDataset.grid` stays the same with or without data.  Could you review this and check if this does what you would like it to? 

```suggestion

    def get_mesh(self, grid_name: str, include_data: bool = False) -> xu.UgridDataset:
        """
        Return a specific ugrid dataset from mesh based on grid_name.

        If include_data is True, the data variables (including optional attribtues)
        for that specific grid are also included.

        Parameters
        ----------
        grid_name : str
            Name of the grid to return.
        include_data : bool, optional
            If True, also include data variables, by default False.

        Returns
        -------
        uds: xu.UgridDataset
            UGrid dataset with or without data variables.
        """
        if self.mesh is None:
            raise ValueError("Mesh is not set, please use set_mesh first.")
        if grid_name not in self.mesh_names:
            raise ValueError(f"Grid {grid_name} not found in mesh.")
        if include_data:
            grid = self.mesh_grids[grid_name]
            # FIXME xugrid #138 #140
            uds = xu.UgridDataset(grid.to_dataset(optional_attributes=True))
            uds.ugrid.grid.set_crs(grid.crs)
            # Look for data_vars that are defined on grid_name
            for var in self.mesh.data_vars:
                if hasattr(self.mesh[var], "ugrid"):
                    if self.mesh[var].ugrid.grid.name == grid_name:
                        uds[var] = self.mesh[var]
                # additionnal topology properties
                elif var.startswith(grid_name):
                    uds[var] = self.mesh[var]
                # else is global property (not grid specific)
        else:
            uds = xu.UgridDataset(self.mesh_grids[grid_name].to_dataset(optional_attributes = True))

        return uds
```

hydromt/models/model_mesh.py
```diff
@@ -304,17 +474,76 @@ def write_mesh(self, fn: str = "mesh/mesh.nc", **kwargs) -> None:
...
-        if self.mesh.ugrid.grid.crs is not None:
+        if self.crs is not None:
             # save crs to spatial_ref coordinate
-            ds_out = ds_out.rio.write_crs(self.mesh.ugrid.grid.crs)
+            ds_out = ds_out.rio.write_crs(self.crs)
         ds_out.to_netcdf(_fn, **kwargs)
 
+    # Other mesh properties
+    @property
+    def mesh_grids(self) -> Dict:
```
xldeltares, line 484: Add type hinting for the dict
```suggestion
    def mesh_grids(self) -> Dict[str, Union[xu.Ugrid1d, xu.Ugrid2d]]:
```

hydromt/models/model_mesh.py
```diff
@@ -304,17 +474,76 @@ def write_mesh(self, fn: str = "mesh/mesh.nc", **kwargs) -> None:
...
+                    elif grid.topology_dimension == 2:
+                        dim = grid.face_dimension
+                    v[f"{grid.name}_node_id"] = xr.DataArray(
+                        v[dim].values.astype(str), dims=dim
+                    )
+                gdf = v.ugrid.to_geodataframe()
+                # FIXME xugrid crs not passed on correctly: https://github.com/Deltares/xugrid/issues/138
+                mesh_gdf[k] = gdf.set_crs(v.ugrid.crs[k])
+
+        return mesh_gdf
```
xldeltares, line 535: Hi @hboisgon, when testing the building and updating of mesh2d, I found that the `to_geodataframe()` function does not always work if 1) there is no data variable (I guess thas why you first assigned a new node_id data variable) 2) there are optiaonl attributes besides face coordinates. Since this is just to return the feometry of the mesh, would it suffice if we skip 'to_geodataframe()' func and create a simple one?
```suggestion
    def mesh_gdf(self) -> Dict:
        """Returns dict of geometry of grids in mesh as a gpd.GeoDataFrame."""
        mesh_gdf = dict()
        if self._mesh is not None:
            for k, v in self.mesh_grids.items():
                # works better on a Dataarray / Dataset
                # (need to add dummy variable if empty)
                if len(v.data_vars) == 0:
                    grid = v.ugrid.grid
                    if grid.topology_dimension == 1:
                        dim = grid.edge_dimension
                    elif grid.topology_dimension == 2:
                        dim = grid.face_dimension
                gdf = gpd.GeoDataFrame(
                    index=v[dim].values.astype(str),
                    geometry=grid.to_shapely(dim),
                )
                mesh_gdf[k] = gdf.set_crs(v.ugrid.crs[k])

        return mesh_gdf
```

hydromt/models/model_mesh.py
```diff
@@ -304,17 +474,76 @@ def write_mesh(self, fn: str = "mesh/mesh.nc", **kwargs) -> None:
...
+                datasets[grid.name] = self.get_mesh(
+                    grid_name=grid.name, include_data=True
+                )
+
+        return datasets
+
+    @property
+    def mesh_names(self) -> List[str]:
+        """List of grid names in mesh."""
+        # FIXME would be nice if xugrid supports it: https://github.com/Deltares/xugrid/issues/141
```
xldeltares, line 508:
```suggestion
        # FIXME to be replaced by uds.ugrid.name/uds.ugrid.name: https://github.com/Deltares/xugrid/issues/141
```

hydromt/models/model_mesh.py
```diff
@@ -304,17 +474,76 @@ def write_mesh(self, fn: str = "mesh/mesh.nc", **kwargs) -> None:
...
             # save crs to spatial_ref coordinate
-            ds_out = ds_out.rio.write_crs(self.mesh.ugrid.grid.crs)
+            ds_out = ds_out.rio.write_crs(self.crs)
         ds_out.to_netcdf(_fn, **kwargs)
 
+    # Other mesh properties
+    @property
+    def mesh_grids(self) -> Dict:
+        """Dictionnary of grid names and Ugrid topologies in mesh."""
+        grids = dict()
```
xldeltares, line 486:
```suggestion
        # FIXME to be replaced by uds.ugrid.topology. See https://github.com/Deltares/xugrid/issues/141
```

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
...
+
+        # update related geoms if necessary: region
+        if overwrite_grid or new_grid:
+            # add / updates region
+            self._geoms.pop("region", None)
+            self.region
+
+    def get_mesh(
+        self, grid_name: str, include_data: bool = False
+    ) -> Union[xu.UgridDataArray, xu.UgridDataset]:
```
xldeltares, line 372: Hi @hboisgon, I slitely modified this function to return only xu.UgridDataset. Before it could return Ugrid1d/Ugrid2d (if include_data is True, or UgridDataset if include_data is False). I find that a little inconsistant. Considering when we call self.mesh it returns a UgridDataset, I think get_mesh returning UgridDataset make sense. Then the downstream actions for getting the grid UgridDataset.grid stays the same with or without data. Could you review this and check if this does what you would like it to?

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
...
+        """
+        Return a specific grid topology from mesh based on grid_name.
+
+        If include_data is True, the data variables for that specific
+        grid are also included.
+
```
xldeltares, line 378:
```suggestion
        Return a specific ugrid dataset from mesh based on grid_name.

        If include_data is True, the data variables (including optional attribtues)
        for that specific grid are also included.

```

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
...
+            Name of the grid to return.
+        include_data : bool, optional
+            If True, also include data variables, by default False.
 
-    def read_mesh(self, fn: str = "mesh/mesh.nc", **kwargs) -> None:
+        Returns
+        -------
+        uds: Union[xu.UgridDataArray, xu.UgridDataset]
+            Grid topology with or without data variables.
+        """
```
xldeltares, line 390:
```suggestion
        uds: xu.UgridDataset
            UGrid dataset with or without data variables.
        """
```

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
...
+        if self.mesh is None:
+            raise ValueError("Mesh is not set, please use set_mesh first.")
+        if grid_name not in self.mesh_names:
+            raise ValueError(f"Grid {grid_name} not found in mesh.")
+        if include_data:
+            grid = self.mesh_grids[grid_name]
+            uds = xu.UgridDataset(
+                grid.to_dataset(optional_attributes=True)
+            )  # FIXME: would be nice if always write all attributes https://github.com/Deltares/xugrid/issues/140
+            uds.ugrid.grid.set_crs(grid.crs)
```
xldeltares, line 400:
```suggestion
            # FIXME xugrid #138 #140
            uds = xu.UgridDataset(grid.to_dataset(optional_attributes=True))
            uds.ugrid.grid.set_crs(grid.crs)
```

hydromt/models/model_mesh.py
xldeltares, line 414: Hi @hboisgon, I slitely modified this function to return only xu.UgridDataset. Before it could return Ugrid1d/Ugrid2d (if include_data is True, or UgridDataset if include_data is False). I find that a little inconsistant. Considering when we call self.mesh it returns a UgridDataset, I think get_mesh returning UgridDataset make sense. Then the downstream actions for getting the grid UgridDataset.grid stays the same with or without data. Could you review this and check if this does what you would like it to?

```suggestion
    def get_mesh(self, grid_name: str, include_data: bool = False) -> xu.UgridDataset:
        """
        Return a specific ugrid dataset from mesh based on grid_name.

        If include_data is True, the data variables (including optional attribtues)
        for that specific grid are also included.

        Parameters
        ----------
        grid_name : str
            Name of the grid to return.
        include_data : bool, optional
            If True, also include data variables, by default False.

        Returns
        -------
        uds: xu.UgridDataset
            UGrid dataset with or without data variables.
        """
        if self.mesh is None:
            raise ValueError("Mesh is not set, please use set_mesh first.")
        if grid_name not in self.mesh_names:
            raise ValueError(f"Grid {grid_name} not found in mesh.")
        if include_data:
            grid = self.mesh_grids[grid_name]
            # FIXME xugrid #138 #140
            uds = xu.UgridDataset(grid.to_dataset(optional_attributes=True))
            uds.ugrid.grid.set_crs(grid.crs)
            # Look for data_vars that are defined on grid_name
            for var in self.mesh.data_vars:
                if hasattr(self.mesh[var], "ugrid"):
                    if self.mesh[var].ugrid.grid.name == grid_name:
                        uds[var] = self.mesh[var]
                # additionnal topology properties
                elif var.startswith(grid_name):
                    uds[var] = self.mesh[var]
                # else is global property (not grid specific)
        else:
            uds = xu.UgridDataset(self.mesh_grids[grid_name].to_dataset())

        return uds
```

hydromt/models/model_mesh.py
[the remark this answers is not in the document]
xldeltares:
```
    def get_mesh(self, grid_name: str, include_data: bool = False) -> xu.UgridDataset:
        """
        Return a specific ugrid dataset from mesh based on grid_name.

        If include_data is True, the data variables (including optional attribtues)
        for that specific grid are also included.

        Parameters
        ----------
        grid_name : str
            Name of the grid to return.
        include_data : bool, optional
            If True, also include data variables, by default False.

        Returns
        -------
        uds: xu.UgridDataset
            UGrid dataset with or without data variables.
        """
        if self.mesh is None:
            raise ValueError("Mesh is not set, please use set_mesh first.")
        if grid_name not in self.mesh_names:
            raise ValueError(f"Grid {grid_name} not found in mesh.")
        if include_data:
            grid = self.mesh_grids[grid_name]
            # FIXME xugrid #138 #140
            uds = xu.UgridDataset(grid.to_dataset(optional_attributes=True))
            uds.ugrid.grid.set_crs(grid.crs)
            # Look for data_vars that are defined on grid_name
            for var in self.mesh.data_vars:
                if hasattr(self.mesh[var], "ugrid"):
                    if self.mesh[var].ugrid.grid.name == grid_name:
                        uds[var] = self.mesh[var]
                # additionnal topology properties
                elif var.startswith(grid_name):
                    uds[var] = self.mesh[var]
                # else is global property (not grid specific)
        else:
            uds = xu.UgridDataset(self.mesh_grids[grid_name].to_dataset())

        return uds
```

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
                     f"Cannot set mesh from {str(type(data).__name__)} without a name."
                 )
             data = data.to_dataset()
+
+        # Checks on grid topology
+        # TODO: check if we support setting multiple grids at once. For now just one
+        if len(data.ugrid.grids) > 1:
+            raise ValueError(
+                "set_mesh methods only supports adding data to one grid at a time."
```
xldeltares, line 301: When setting a mesh, this convertion will lose the optional attributes (accoding to xugrid and ugrid convention) that are *not* optional for plugin.
```suggestion
            data = data.to_dataset(optional_attributes = True)
```

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
...
+                            " data variables in mesh."
+                        )
+                        grids = [
+                            self.mesh_datasets[g].ugrid.to_dataset()
+                            for g in self.mesh_names
+                            if g != grid_name
+                        ]
+                        # Re-define _mesh
+                        grids = xr.merge(grids)
+                        self._mesh = xu.UgridDataset(grids)
```
xldeltares, line 348: Should also preserve optional attributes
```suggestion
                            self.mesh_datasets[g].ugrid.to_dataset(optional_attributes=True)
```

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
...
+            else:
+                # We are potentially adding a new grid without any data variables
+                self._mesh = xu.UgridDataset(
+                    xr.merge([self.mesh.ugrid.to_dataset(), data.ugrid.to_dataset()])
+                )
+            # Restore crs
+            for grid in self._mesh.ugrid.grids:
+                grid.set_crs(crs)
+
+        # update related geoms if necessary: region
```
xldeltares, line 364: same above
```suggestion
                    xr.merge([self.mesh.ugrid.to_dataset(optional_attributes=True), data.ugrid.to_dataset(optional_attributes=True)])
```

hydromt/models/model_mesh.py
```diff
@@ -253,33 +293,163 @@ def set_mesh(
...
-                self._mesh[dvar] = data[dvar]
+            # Check on crs
+            if not data.ugrid.grid.crs == self.crs:
+                raise ValueError("Data and self.mesh should have the same CRS.")
+            # Save crs as it will be lost when converting to xarray
+            crs = self.crs
+            # Check on new grid topology
+            if grid_name in self.mesh_names:
+                # check if the two grids are the same
+                if (
```
xldeltares, line 323: I think because here is just checking, the optional_attributes can be False.

xldeltares, line 323: I think here optional attributes are not needed.

xldeltares:
HI @hboisgon , I have finished review by testing building and updating mesh2d. This time I tried the **make suggestions** for review, hoping that it is a bit more clear for communication (also for my remembering where and why I changed something). Could you please review the review (..:D)? 

attention points:

- [ ] updated `mesh_gdf` to write only the geometry of the mesh (without any data, therefore not using `to_geodataframe` anymore)
- [ ] updated `get_mesh` and hence `mesh_datasets` to return UgridDataSet always (mimicking `self.mesh`)
- [ ] adding `optional_attributes = True` where needed, mainly when converting grid topology to grid dataset or when transferring information between dataset (context: https://github.com/Deltares/xugrid/issues/141) 
- [ ] finally, xugrid will have `uds.ugrid.names` and `uds.ugrid.topology` available next release (context:https://github.com/Deltares/xugrid/issues/141)

Thanks and feel free to book a meeting in my agenda for discussion!

hydromt/models/model_mesh.py
hboisgon: I guess you did some commits after the review so the line tracking from git is not working anymore (change is applied to another line...). But I will change and apply your suggestion in a local commit :)

hydromt/models/model_mesh.py
hboisgon: Same as above

hydromt/models/model_mesh.py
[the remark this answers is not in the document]
hboisgon: Looks good and indeed makes more sense!

hydromt/models/model_mesh.py
hboisgon: Sounds good for now and avoid the error yes. Maybe then if it's useful we could add a flag to also add data variables to it, but let's put that on the wish list. Maybe still good to report these bugs to xugrid though as they may want to look into this as well.

hboisgon:
> HI @hboisgon , I have finished review by testing building and updating mesh2d. This time I tried the **make suggestions** for review, hoping that it is a bit more clear for communication (also for my remembering where and why I changed something). Could you please review the review (..:D)?
> 
> attention points:
> 
> * [x]  updated `mesh_gdf` to write only the geometry of the mesh (without any data, therefore not using `to_geodataframe` anymore)
> * [x]  updated `get_mesh` and hence `mesh_datasets` to return UgridDataSet always (mimicking `self.mesh`)
> * [ ]  adding `optional_attributes = True` where needed, mainly when converting grid topology to grid dataset or when transferring information between dataset (context: [[help wanted] understand the optional_attributes  xugrid#140](https://github.com/Deltares/xugrid/issues/140))
> * [ ]  finally, xugrid will have `uds.ugrid.names` and `uds.ugrid.topology` available next release (context:[support get mesh/grid names  xugrid#141](https://github.com/Deltares/xugrid/issues/141))
> 
> Thanks and feel free to book a meeting in my agenda for discussion!

For optional_attributes discussion, I checked the reply in the linked issue and from this I think it's important we keep during data processing within hydromt but then maybe we make it optional to actually write these to netcdf in the end? Then if some software require them or should not have them, they can decide to write them out or not.

hboisgon:
Hi @xldeltares , thanks a lot for the very detailed and useful review :) I think we are there now to finally merge this in core. Could you do a final check? Then I will ask someone else from core to also have a look before merging. Many thanks!

xldeltares (approved):
Hi @hboisgon , I have reviewed the changes and they look good. hence approving the pull request for the next step in the PR pipeline.

hydromt/models/model_mesh.py
```diff
@@ -252,16 +292,135 @@ def set_mesh(
...
+            else:
+                # We are potentially adding a new grid without any data variables
+                self._mesh = xu.UgridDataset(
+                    xr.merge(
+                        [
+                            self.mesh.ugrid.to_dataset(optional_attributes=True),
+                            data.ugrid.to_dataset(optional_attributes=True),
+                        ]
+                    )
+                )
```
savente93, line 366: Should we check here that the meshes have the same spacial coordinates like it says in the doc string or is this guaranteed?

hydromt/models/model_mesh.py
```diff
@@ -252,16 +292,135 @@ def set_mesh(
...
+            uds.ugrid.grid.set_crs(grid.crs)
+            # Look for data_vars that are defined on grid_name
+            for var in self.mesh.data_vars:
+                if hasattr(self.mesh[var], "ugrid"):
+                    if self.mesh[var].ugrid.grid.name == grid_name:
+                        uds[var] = self.mesh[var]
+                # additionnal topology properties
+                elif var.startswith(grid_name):
+                    uds[var] = self.mesh[var]
+                # else is global property (not grid specific)
```
savente93, line 414: do we need an else branch here? (not sure if the comment implies that we should or shouldn't)

hydromt/workflows/mesh.py
```diff
@@ -0,0 +1,237 @@
...
+
+    Returns
+    -------
+    mesh2d : xu.UgridDataset
+        Generated mesh2d.
+    """  # noqa: E501
+    kind, region = parse_region(region, logger=logger)
+    if kind != "mesh":
+        if not isinstance(res, (int, float)):
+            raise ValueError("res argument required for kind 'bbox', 'geom'")
```
savente93, line 76: for clarity I think it'd be good to show the type that was actually passed.

savente93 (approved):
Looking good, a few small comments, but nothing blocking. Well done!

hydromt/workflows/basin_mask.py
```diff
@@ -135,6 +141,17 @@ def parse_region(region, logger=logger):
             kwargs = dict(grid=open_raster(value0, **kwargs))
         elif isinstance(value0, (xr.Dataset, xr.DataArray)):
             kwargs = dict(grid=value0)
+    elif kind == "mesh":
+        if _compat.HAS_XUGRID:
+            import xugrid as xu
```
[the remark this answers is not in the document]
hboisgon: Check

hydromt/models/model_mesh.py
hboisgon: In this part, a new grid (1D or 2D or 3D) is added so the spatial coordinates can be different, the check is done only if new data is added on an existing grid definition in mesh

hydromt/models/model_mesh.py
hboisgon: No no need for an else but I added the comment to kind of explain or be reminded why we don't need an else statement

hydromt/workflows/mesh.py
hboisgon: True. but here I expect more that people forget to set the argument and the default None is returned, so I could also have put the check to if res is not None. So I think the error message should be okay or clearer this way.

PR #423, opened 2023-06-21
Fix haning test

### LINKED ISSUE #408: One of the data adapter tests sometimes hangs
### HydroMT version checks

### Reproducible Example

pytest

### Current behaviour

Sometimes the test suite doesn't complete because one of the tests is hanging. 

### Desired behaviour

The tests always either fail or succeed but don't run indefinitely

### Additional context

Not sure yet what causes this, more triage will be needed, but given that it usually seems to be waiting on a lock when this happens and I kill it I suspect it's a concurrency deadlock of some description. As I said, needs more investigation

savente93:
## Issue addressed
Fixes #408 

## Explanation
The problem seemed to be that Dask by default uses a light weight thread scheduler for it's workloads. This is lightweight but threads are flaky in python at best, and to the best of my understanding this caused the test to occasionally hang because at least two Dask threads were getting into a deadlock (each thread waiting on the other to complete something so they can continue their own work). Because this is quite a hard to reproduce issue and is arguably an upstream issue, I decided to go for the easy fix and just use the synchronous scheduler in Dask which should not run into this problem. If we run into the issue in more places we can start thinking of a more fundamental solution. 

Because the issue is quite hard to reproduce I can't be 100% certain that this solves the issue, but I ran the solution in a loop on my laptop in a while and the hang didn't occur once where without the fix, so I assume this is an adequate solution for this. 

## Checklist

async code is hard, especially in python :(

### DIFF
--- a/tests/test_data_adapter.py
+++ b/tests/test_data_adapter.py
@@ -120,6 +120,11 @@ def test_rasterdataset_zoomlevels(rioda_large, tmpdir):
 
 
 def test_geodataset(geoda, geodf, ts, tmpdir):
+    # this test can sometimes hang because of threading issues therefore
+    # the synchronous scheduler here is necessary
+    from dask import config as dask_config
+
+    dask_config.set(scheduler="synchronous")
     fn_nc = str(tmpdir.join("test.nc"))
     fn_gdf = str(tmpdir.join("test.geojson"))
     fn_csv = str(tmpdir.join("test.csv"))


hboisgon:
Hey @savente93, nice finding. I'm just a little bit worried if this only happens in test environment or if that means it could also happens on the user side? I remember for python 11 I already had to change that test or well delete part of it that did not pass anymore and I had the feeling it was also linked to the TempDirectory call:

``` with tempfile.TemporaryDirectory() as td ```

Could you check if we can replace this and if it may be a second reason why tests hang specifically here? Else I'm also good with the dask fix, sounds also like this could be the issue indeed :)

savente93:
So I had a suspiscion that it had something to do with the temporary file directive. It seems to happen when there is some kind of contention around file descriptors (my intuition I don't have anything to back this up). I think what causes this is that in this test a number of files is written to in "rapid succession". I'm not totally sure, but that doesn't sound like a very common use case, so I don't think that it would be much of an issue for the user. However, because issues like these are very difficult to diagnose/ resolve, I have no idea how much more time I'd need to dive deeper into this problem so I would suggest we leave it here, see if it works, and deal with it if it comes up again. How does that sound?

DirkEilander (approved):

PR #427, opened 2023-06-27
Fix yamlfmt

### LINKED ISSUE #426: Current yaml format pre-commit hook does not work on windows
### HydroMT version checks

### Reproducible Example

Add-Content .\data\catalog\deltares_data.yml `n
git add .
git commit -m "should fail on ymlfmt" 

### Current behaviour

Underwindows (PS specifically, but also unde CMD) the current yaml format hook does not work. Apparently it is a known issue that this hook does not work on windows. 

### Desired behaviour

We should either decide to just not have the hook or find one that does work on windows 

### Additional context

thanks for @Tjalling-dejong for flagging this

savente93:
## Issue addressed
Fixes #426 

## Explanation
The previous pre-commit hook for yaml formatting did not work on windows for whatever reason. So in order to remedy this I moved us over to the google pre-commit hook, which should be better maintained. If I recall correctly the google formatter has a bit fewer options for customization, but I tuned it to be as close to the format rules we already have. It made a few small changes, but overall should be fine, and now it works on windows as well, so all in all, I think this is fine. 

## Checklist

### DIFF
--- a/.github/ISSUE_TEMPLATE/documentation_improvement.yaml
+++ b/.github/ISSUE_TEMPLATE/documentation_improvement.yaml
@@ -9,8 +9,7 @@ body:
       label: HydroMT version checks
       options:
         - label: >
-            I have checked that the issue still exists on the latest versions of the docs
-            on `main` [here](https://github.com/Deltares/hydromt)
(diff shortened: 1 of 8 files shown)


Tjalling-dejong (approved):
Tested locally and works. LGTM 👍

PR #440, opened 2023-07-12
Osm tiling

### LINKED ISSUE #345: ENH: OSM tiling (memory usage)
### Feature Type

- [X] Adding new functionality to HydroMT

- [ ] Changing existing functionality in HydroMT

### Enhancement Description

Tile a gridded dataset according to the OpenStreetMap structure.
Functionality exists within HydroMT, but has been quoted out due to memory issues.

### Feature Description

Use the '.raster' accessor to tile a gridded dataset as per OSM's system.

### Additional Context

Some information can be found [here](https://wiki.openstreetmap.org/wiki/Zoom_levels)

B-Dalmijn:
## Issue addressed
Fixes #345

## Explanation
Created tiling function by starting from the bottom in terms of zoomlevel. This starting level is user defined or defaulted to 5.

## Checklist


hydromt/raster.py
```diff
@@ -2233,6 +2234,243 @@ def tile_window(shape, px):
         with open(join(root, f"{mName}.yml"), "w") as f:
             yaml.dump({mName: yml}, f, default_flow_style=False, sort_keys=False)
 
+    def to_osm_tiles(
+        self,
+        root: str,
+        zoom_levels: list,
+        bbox: tuple = (),
+        method: str = "nearest",
+        min_lvl: int = 5,
```
DirkEilander, line 2243: Would it be possible to calculate min_lvl from the original data, similar to to_xyz_tiles?

hydromt/raster.py
```diff
@@ -2233,6 +2234,243 @@ def tile_window(shape, px):
...
+                temp_transform = Affine(
+                    obj_res,
+                    0,
+                    obj_bounds[0] + l * obj_res,
+                    0,
+                    -obj_res,
+                    obj_bounds[3] - u * obj_res,
+                )
+
+                rasterio.warp.reproject(
```
DirkEilander, line 2408: Why do you use the rasterio reproject method instead of the raster.reproject method in the same script? That would avoid a lot of lines of code like creating a new tmp array below.

hydromt/raster.py
```diff
@@ -2233,6 +2234,243 @@ def tile_window(shape, px):
...
+            t_r = pyproj.Transformer.from_crs(
+                3857, obj.raster.crs.to_epsg(), always_xy=True
+            )
+
+            sd = join(root, f"{zl}")
+            create_folder(sd)
+            txt_path = join(sd, "filelist.txt")
+            file = open(txt_path, "w")
+
+            for transform, bounds, col, row in tile_window(zl, minx, miny, maxx, maxy):
```
DirkEilander, line 2370: Would it make sense to make a separate function for the part within this loop? In that case python takes care of your garbage collection much better and you don't have to work with del and gc.collect.

hydromt/raster.py
```diff
@@ -2233,6 +2234,243 @@ def tile_window(shape, px):
...
+        obj = obj.transpose(self.y_dim, self.x_dim)
+
+        mName = os.path.normpath(os.path.basename(root))
+
+        def create_folder(path):
+            if not os.path.exists(path):
+                os.makedirs(path)
+
+        def transform_res(dres, transformer):
+            return transformer.transform(0, dres)[0]
```
DirkEilander, line 2286: This function is made to transform coordinates, not resolution. For this (and also to calculate the initial zoom level, see other comment) we could make use of e.g. [rasterio.warp.calculate_default_transform](https://rasterio.readthedocs.io/en/latest/api/rasterio.warp.html#rasterio.warp.calculate_default_transform).

DirkEilander (commented):
Thanks for putting this together @B-Dalmijn! Some initial comments
- I've tried testing it with a local dataset in a project UTM CRS but than the dres after transforming (line 2305) becomes negative and the I get an error on the next line. 
- Could you add some more comments to explain what certain blocks of code are doing, For instance I don't understand what lines 2304-2306 are for.
- out of curiosity, would it be possible to reuse (part of) to_xyz_tiles after you have created the first zoom level (close to the native data resolution) for the subsequent zoom levels?

Let me know if something is unclear.

hydromt/raster.py
B-Dalmijn: Agreed, this was a quick but bad solution.

hydromt/raster.py
B-Dalmijn: Yes, that's what dres and nzl are supposed to do.  (line 2304-2306).
But at this moment in time, it's badly written. Apperently it also causes issues using more local references (which was to be expected)
I'll get busy on writing a proper implementation.

@DirkEilander, send me the dataset you used so I can test it.

hydromt/raster.py
B-Dalmijn: See issue #442

hydromt/raster.py
B-Dalmijn: I manually call the garbage collection between iterations to clear some memory that is not automatically cleared (in order to avoid creeping memory), a seperate function would not solve that.

hydromt/raster.py
B-Dalmijn: I'll will write this out in the issue when I reopen it. For now I think the implementation is fine as it is, but in the future it would indeed be nicer to use the raster.reproject method. However I'd like to solve issue #442 first instead more or less 'hacking' it.

hydromt/raster.py
B-Dalmijn: Fixed that.

hydromt/raster.py
B-Dalmijn: This is now hopefully fixed.

B-Dalmijn:
@DirkEilander Your last point (copying from 'to_xyz_tiles') is one I thought of for a bit. However I really need to have the locations of the tiles (windows basically) as they are fixed. That meant I could not translate the code from 'to_xyz_tiles' one to one, where this is not used whatsoever. Maybe talking in person I can explain more clearly.

hydromt/raster.py
B-Dalmijn: It's more like a check now for a warning now, but we could later on chose to not use the min_lvl kwargs anymore.

B-Dalmijn:
No clue why Sonarcloud failed where it did...

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+            GDAL driver (e.g., 'GTiff' for geotif files), or 'netcdf4' for netcdf files.
+        **kwargs
+            Key-word arguments to write raster files
+        """
+        resampling = getattr(Resampling, method, None)
+        if resampling is None:
+            raise ValueError(f"Resampling method unknown: {method}.")
+        px_size = 256  # Fixed resolution for osm
+
+        assert self._obj.ndim == 2, "Only 2d datasets are accepted..."
```
savente93, line 2289: Either this should be removed, or we should raise a proper `ValueException` I think

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+        obj = self._obj.copy()
+        obj = obj.transpose(self.y_dim, self.x_dim)
+
+        mName = os.path.normpath(os.path.basename(root))
```
savente93, line 2293: Everything else uses snake_case so I think this should too

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+            # Origin displacement
+            odx = np.floor(abs(-20037508.34 - minx) / dx)
+            ody = np.floor(abs(20037508.34 - maxy) / dx)
+
+            # Set the new origin
+            minx = -20037508.34 + odx * dx
+            maxy = 20037508.34 - ody * dx
+
+            # Create window generator
+            lu = product(
```
savente93, line 2332: same comment as with nzl

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+            "EPSG:3857",
+            *obj.shape,
+            *obj.raster.bounds,
+        )[0]
+        dres = _obj_tr[0]
+        nzl = int(np.ceil((np.log10((20037508.34 * 2) / (dres * 256)) / np.log10(2))))
+
+        def tile_window(zl, minx, miny, maxx, maxy):
+            # Origin displacement
+            odx = np.floor(abs(-20037508.34 - minx) / dx)
```
savente93, line 2324: I think dx should also be an argument here? or it should be calculated in this function.

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+        nodata = self.nodata
+        zls = {}
+        for zl in zoom_levels:
+            dx = (20037508.34 * 2) / (2**zl)
+            px = dx / px_size
+
+            if max(zoom_levels) > nzl:
+                logger.warning(
+                    f"Pixels at zoomlevel {zl} are smaller than the original data"
+                )
```
savente93, line 2365: when given this warning, would the user know what to do? or can we add some instructions here? (if the answer to the first question is yes, you can ignore this comment)

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+            if vrt_fn is not None:
+                obj = xr.open_dataarray(vrt_fn, engine="rasterio").squeeze(
+                    "band", drop=True
+                )
+
+            obj_bounds = obj.raster.bounds
+            obj_res = obj.raster.res[0]
+            obj_shape = obj.raster.shape
+
+            t_r = pyproj.Transformer.from_crs(
```
savente93, line 2377: same comment as `nzl`

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+                3857, obj.raster.crs.to_epsg(), always_xy=True
+            )
+
+            sd = join(root, f"{zl}")
```
savente93, line 2381: what does `sd` stand for?

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+            create_folder(sd)
+            txt_path = join(sd, "filelist.txt")
+            file = open(txt_path, "w")
```
savente93, line 2384: You're leaking memory here if an exception is raised during the inner loop because this will never get closed. Please use a context manager.

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+
+                u = math.floor(round(_window[3], 2))
+                l = math.floor(round(_window[0], 2))
+                h = obj_shape[0] - u - math.floor(round(_window[1], 2))
+                w = obj_shape[1] - l - math.floor(round(_window[2], 2))
+
+                temp = obj[u : u + h, l : l + w]
+
+                tile = np.empty((px_size, px_size), np.float64)
+                tile[:] = nodata
```
savente93, line 2413: i think this would be more efficient: 

```suggesion
tile = np.full((px_size,px_size), nodata,dtype=np.float64)
```

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+                        "x": np.arange(
+                            bounds[0] + 0.5 * transform[0],
+                            bounds[2] + 0.5 * transform[0],
+                            transform[0],
+                        )[:px_size],
+                    },
+                    data=tile,
+                )
+
+                del tile
```
savente93, line 2451: I'm not 100% sure, but I think that `data=tile`  creates a reference, so the tile might not actually get cleaned up when you call `del tile`. You might need a deep copy here.

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+                    path = join(ssd, f"{row}.nc")
+                    temp = temp.raster.gdal_compliant()
+                    temp.to_netcdf(path, engine="netcdf4", **kwargs)
+                elif driver in gis_utils.GDAL_EXT_CODE_MAP:
+                    ext = gis_utils.GDAL_EXT_CODE_MAP.get(driver)
+                    path = join(ssd, f"{row}.{ext}")
+                    temp.raster.to_raster(path, driver=driver, **kwargs)
+                else:
+                    raise ValueError(f"Unkown file driver {driver}")
+                file.write(f"{path}\n")
```
savente93, line 2467: If this is the only time you write to file, it might be more efficient to just open up the filew with a context manager, write to it, and close it again. especially if you're concerned about memory.

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+        **kwargs
+            Key-word arguments to write raster files
+        """
+        resampling = getattr(Resampling, method, None)
+        if resampling is None:
+            raise ValueError(f"Resampling method unknown: {method}.")
+        px_size = 256  # Fixed resolution for osm
+
+        assert self._obj.ndim == 2, "Only 2d datasets are accepted..."
+        obj = self._obj.copy()
```
savente93, line 2290: Not quite sure if this is true, but it doesn't seem like you actually have to mutate `obj` in any way in this function. If that is the case and you are having memory issues, you could see if you can avoid making a full copy, and just use a reference to the object directly

savente93 (commented):
I can't totally attest to the semantics of the code, but assuming that's okay I'd say this looks okay. I left a few comments regarding code quality and possibilities for optimizing memory if that is an issue. Should still wait for Dirks approval to make sure the calculations are actually correct.

hydromt/raster.py
B-Dalmijn: Fair enough, will change that

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+                os.makedirs(path)
+
+        if bbox:
+            minx, miny, maxx, maxy = bbox
+        else:
+            minx, miny, maxx, maxy = obj.raster.transform_bounds("EPSG:3857")
+
+        minx, miny = map(
+            max,
+            zip((minx, miny), [-20037508.34] * 2),
```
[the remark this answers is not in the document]
B-Dalmijn: That will be done at a later stage when the functionailty will be expanded upon.
It represents the upper left corner of the epsg:3857 domain.

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+        )
+
+        _obj_tr = rasterio.warp.calculate_default_transform(
+            obj.raster.crs,
+            "EPSG:3857",
+            *obj.shape,
+            *obj.raster.bounds,
+        )[0]
+        dres = _obj_tr[0]
+        nzl = int(np.ceil((np.log10((20037508.34 * 2) / (dres * 256)) / np.log10(2))))
```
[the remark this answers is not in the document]
B-Dalmijn: Number of zoom levels.

hydromt/raster.py
B-Dalmijn: No, it checks whether the resampling method is available in rasterio. A ValueError (as far as I know) would then be correct.

hydromt/raster.py
```diff
@@ -2247,6 +2248,245 @@ def tile_window(shape, px):
...
+            raise ValueError(f"Resampling method unknown: {method}.")
+        px_size = 256  # Fixed resolution for osm
+
+        assert self._obj.ndim == 2, "Only 2d datasets are accepted..."
+        obj = self._obj.copy()
+        obj = obj.transpose(self.y_dim, self.x_dim)
+
+        mName = os.path.normpath(os.path.basename(root))
+
+        def create_folder(path):
```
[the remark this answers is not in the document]
B-Dalmijn: Fair enough, but where would we define this method? Seems like a function that then should be in an util file...

hydromt/raster.py
savente93: sorry I think I messed up the selection there. I was referring to the `assert` statement

hydromt/raster.py
B-Dalmijn: If an exception is raised, the python interpreter stops and the garbage collection is called. Won't that just clean this stuff up? (genuinely asking).  c-like memory leaks are quite rare in python.

hydromt/raster.py
B-Dalmijn: Yup.

hydromt/raster.py
roeldegoede: General remark from the user point of view; the advantage of using the OSM-tiles (basically just a specific format of xyz-tiles) is that the tiles are fixed in position and size (256x256), and ofcourse coordinate reference system. I think that allows for having different formats, like the .png format, instead of the netcdf4 or geotiffs, that do NOT include any spatial information. These .pngs can then very easily be visualized in webviewers or any other mapbox/leaflet like applications.

I'm not sure if the method here, using the VRT file, still works for binary formats or the .png format that doesnt have any spatial information included? Webviewers but also QGIS don't need the VRT, so there seem to be different ways to easily read the correct tiles at least?

hydromt/raster.py
savente93: Depends on if someone in the stack catches your exception, which plenty of libs/applications do. Though I'm guessing it's not quite a memory leak, as much as a dangling file descriptor that's never closed? Unsure how that is handled. But regardless, I think it's best to be more careful, taking the first bit into consideration.

hydromt/raster.py
B-Dalmijn: Maybe, we want to assert that there are indeed two dimensions, 3d data is not accepted. Don't know if that then should be a ValueError...

hydromt/raster.py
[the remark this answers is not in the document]
B-Dalmijn: Done

hydromt/raster.py
B-Dalmijn: Done

hydromt/raster.py
B-Dalmijn: Haven't had the time to properly investigate this, did implement a deep copy.

hydromt/raster.py
B-Dalmijn: It's an inhouse function, which itself calls writing functions/ a context manager. Can't put one here,

hydromt/raster.py
B-Dalmijn: I agree it's probably better, but this will most likely not cause any harm. I already indented quite a lot due to the flow of control, so I'd rather not use the contextmanager from an aesthetic standpoint.

DirkEilander:
remark by @roeldegoede which was placed in an unrelated thread above. Moving it here so we won't forget.

> General remark from the user point of view; the advantage of using the OSM-tiles (basically just a specific format of xyz-tiles) is that the tiles are fixed in position and size (256x256), and ofcourse coordinate reference system. I think that allows for having different formats, like the .png format, instead of the netcdf4 or geotiffs, that do NOT include any spatial information. These .pngs can then very easily be visualized in webviewers or any other mapbox/leaflet like applications.
> 
> I'm not sure if the method here, using the VRT file, still works for binary formats or the .png format that doesnt have any spatial information included? Webviewers but also QGIS don't need the VRT, so there seem to be different ways to easily read the correct tiles at least?

hydromt/raster.py
DirkEilander: The copy method provides a new view (shallow copy) of the dataset is my understanding. So it shouldn't really matter in terms of memory.

tests/test_raster.py
```diff
@@ -527,5 +527,14 @@ def test_to_xyz_tiles(tmpdir, rioda_large):
         assert len(f.readlines()) == 1
 
 
-# def test_to_osm(tmpdir, dummy):
-#     dummy.raster.to_osm(
+def test_to_osm_tiles(tmpdir, rioda_large):
```
DirkEilander, line 530: It would be good to make this test a bit more precise, for instance check the bounding box of a specific output tile.

DirkEilander (commented):
Thanks for this work @B-Dalmijn. I've checked the tiles created for two different datasets in different projections and checked the tiles against [mercantile](p:\wflow_global\hydromt_staging\DTM0\osm1\13\5218\4513.tif). 

Functionally, we still need to discuss what is best from a user perspective:
- the zoom levels. Should these be calculated for you and/or user defined? Now a user needs to specify a list ( a min and max would do) and I can image the highest res level can be determined from the data and set as an optional argument.
- the output formats. For visualisation @roeldegoede suggests to write output to png. We could then use a tool like mercantile to be able to read the 4 higher res tiles that define the next tile without the need for a vrt.

Regarding the code. You left many useful comments by Sam regarding code readability unanswered. I'd be great if you could actually implement this.

I suggest to plan a meeting with @roeldegoede before continuing this PR to make sure the final touch is in line with user needs.

hydromt/raster.py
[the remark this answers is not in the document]
dalmijn: Agreed

hydromt/raster.py
dalmijn: Added comments all around

hydromt/raster.py
dalmijn: Can be a function of y_ext_pm, so yes and no. Ill give that as an argument

hydromt/raster.py
dalmijn: Done

hydromt/raster.py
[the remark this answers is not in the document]
dalmijn: Done

hydromt/raster.py
dalmijn: Ill elaborate the warning at bit, consider it done.

hydromt/raster.py
dalmijn: Sub-directory; again: added comments

tests/test_raster.py
dalmijn: Did that, maybe add a bit more later on.

dalmijn:
I processed the comments, reshuffled stuff a bit and made it more readable (I think). 
Added a new method to_webviewer_tiles.

savente93 (commented):
I don't feel I can say anything meaningful about the maths without taking a significant amount of time to understand it. The code looks good, with the comment that the `mercantile` package seems to not be actively maintained. if it is mostly because it's considered stable, (which I don't feel qualified to comment on) then it's okay, but otherwise it might be good if we can find a suitably maintained alternative. Other than that LGTM but I'll leave the final approval for Dirk

dalmijn:
It's designated as a stable release. I'd say it is indeed considered stable and just finished.

hydromt/raster.py
```diff
@@ -2340,28 +2353,61 @@ def to_slippy_tiles(
         }.get(driver.lower(), {})
         kwargs = {**kwargs0, **kwargs}
 
+        # Setting up information for zoomlevel calculation and
+        # determination of tile windows
+        # This section is purely for the resolution
```
DirkEilander, line 2358: Could this section be moved inside the if `max_lvl is None:` statement?

hydromt/raster.py
```diff
@@ -2340,28 +2353,61 @@ def to_slippy_tiles(
...
+            zip(bounds_4326_clip[:2], (-180, -y_ext)),
+        )
+        bounds_4326_clip[2:] = map(
+            min,
+            zip(bounds_4326_clip[2:], (180, y_ext)),
+        )
+        obj_clipped_to_pseudo = obj.raster.clip_bbox(
+            bounds_4326_clip,
+            crs="EPSG:4326",
+        )
```
DirkEilander, line 2371: Wouldn't this do the same? Or am I missing something? 
```suggestion
        obj_clipped_to_pseudo = obj.raster.clip_bbox(
            [-180, -y_ext, 180, y_ext],
            crs="EPSG:4326",
        )
```

DirkEilander (approved):
Thanks for the last updates @dalmijn! A few very small final changes left before merging. Can you apply these and merge?

hydromt/raster.py
dalmijn: No, I just wrote it way too explicitly..., kind of laughed when I saw this.

hydromt/raster.py
```diff
@@ -2292,6 +2292,13 @@ def to_slippy_tiles(
             If None, the zoomlevels will be determined based on the data resolution
         driver : str, optional
             file output driver, one of 'png', 'netcdf4' or 'GTiff'
+        cmap : str | object, optional
+            A colormap, either defined by a string and imported from matplotlib
+            via that string or as a ListedColormap object from matplotlib itself.
```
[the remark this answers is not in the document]
dalmijn: The object is called a ListedColormap, but I will change it.

dalmijn:
Done all the last stuff, so I'll merge it.

hydromt/raster.py
dalmijn: Done.

PR #441, opened 2023-07-12
from_gdf: gdf vs data index

### LINKED ISSUE #415: Geodataset from_gdf with different index len
### HydroMT version checks

### Reproducible Example

It seems behavior of vector.GeoDataArray.from_gdf() changed in hydromt 0.7.0 linked to this PR: https://github.com/Deltares/hydromt/pull/276/files#

Before if you had `da = hydromt.vector.GeoDataArray.from_gdf(gdf, da_ts, index_dim="index")` the len of index in gdf and da_ts did not have to be equal (eg some station in gdf may not have data in da_ts but for the rest of the index there is no issue).
Since this PR the index in gdf and da_ts should be exactly the same name/len.

### Current behaviour

Return error if index in gdf is different than index in da_ts:

image

### Desired behaviour

I think it would still be nice if we don't match exactly on len the index in gdf and da_ts. We could even return NaNs if needed. This used to work before the PR mentioned above.

### Additional context

_No response_

B-Dalmijn:
## Issue addressed
Fixes #415 

## Explanation
Added a new `index_merge` argument in `GeoDataset.from_gdf` and `GeoDataArray.from_gdf` to decide whether to merge based on the gdf index (default) or the intersection of both the gdf and data indices. Other merge options do not make sense as a GeoDataset/Array should contain a coordinate with geometries which cannot/should not have missing values. An error is raised in both cases if there is no overlap in indices between both datasets.

## Checklist

### DIFF
--- a/hydromt/vector.py
+++ b/hydromt/vector.py
@@ -740,6 +740,10 @@ class GeoDataArray(GeoBase):
         if index_dim is None:
             index_dim = gdf.index.name if gdf.index.name is not None else "index"
         geom_name = gdf.geometry.name
+        # Some index dimension checking
+        # TODO Check this!!! Might be unwanted
+        if index_dim in gdf.columns:
+            gdf.set_index(index_dim, inplace=True)
         # create DataArray from array_like data
         da = xr.DataArray(data=data, coords=coords, dims=dims, name=name)
         # check dims -> assume index dim is first dim if not provided
@@ -760,6 +764,7 @@ class GeoDataArray(GeoBase):
             hdrs = gdf.columns
         else:
             hdrs = [geom_name]
+        da = da.reindex({index_dim: gdf.index})
         da = da.assign_coords({hdr: (index_dim, gdf[hdr]) for hdr in hdrs})
         da = da.transpose(index_dim, ...).reindex({index_dim: gdf.index})
         # set geospatial attributes
--- a/hydromt/vector.py
+++ b/hydromt/vector.py
@@ -702,6 +702,7 @@ class GeoDataArray(GeoBase):
         name: str = None,
         index_dim: str = None,
         keep_cols: bool = True,
+        how: str = "gdf",
     ) -> xr.DataArray:
         """Parse GeoDataFrame object with point geometries to DataArray.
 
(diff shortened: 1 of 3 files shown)


hydromt/vector.py
```diff
@@ -740,6 +740,10 @@ def from_gdf(
         if index_dim is None:
             index_dim = gdf.index.name if gdf.index.name is not None else "index"
         geom_name = gdf.geometry.name
+        # Some index dimension checking
+        # TODO Check this!!! Might be unwanted
+        if index_dim in gdf.columns:
+            gdf.set_index(index_dim, inplace=True)
```
DirkEilander, line 746: I like to avoid changing the input data as the use may want to continue using gdf outside of this function. Therefore better to not use the inplace=True argument.

```suggestion
            gdf = gdf.set_index(index_dim)
```

hydromt/vector.py
```diff
@@ -760,6 +764,7 @@ def from_gdf(
             hdrs = gdf.columns
         else:
             hdrs = [geom_name]
+        da = da.reindex({index_dim: gdf.index})
```
DirkEilander, line 767: I think it would be nice to support different flavours of merging the data similar to the `how` argument in [pandas.merge](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.merge.html)

However I'm not sure it makes sense to create a GeoDataset with empty geometries in case the gdf index overlaps with only a fraction of the dataset index_dim. For now we could start supporting the `how='gdf_index'` (use  only keys from gdf_index - this is basically wat you implemented now) and `how='inner'` (use intersection of both indices). You can use the [pandas.index.intersection](https://pandas.pydata.org/docs/reference/api/pandas.Index.intersection.html?highlight=intersection) method.

DirkEilander (changes requested):
Thanks for addressing this issue @B-Dalmijn 
I have added two suggestions and it would also be good to create a small test to check this behavior. 

@hboisgon would be very happy if this is included in the release we are planning for Tuesday. Is it possible for you to finish it before?

hydromt/vector.py
B-Dalmijn:
```python
gdf.set_index(index_dim, inplace=True)
```
I was kinda wondering the same thing. We should, I think, discuss this, because there is a trade-off in terms of memory. Most of the time you wont get near memory issues, but it is something to keep in mind when copying the thing.
So I agree but we could discuss it.

hydromt/vector.py
B-Dalmijn: I thought of it as well, I came to the conclusion that it's created from the GeoDataFrame and the array of data is there as a supplement so to speak. So I focused on the fact that there should always be an index from the gdf and not necessarily from the array. 
However if we find this to be too simplistic and there are use cases where how = 'array' is needed, then we could indeed support that.
I will implement a how keyword (for 'gdf_index and inner'), that seems like a good idea. how = 'gdf_index' is indeed now the way is it.

hydromt/vector.py
B-Dalmijn: Changed it.

hydromt/vector.py
B-Dalmijn: Done, for now. Except for how = 'data'.

DirkEilander:
This PR was still missing tests and did not implement this functionality for GeoDatasets (only GeoDataArrays). As I believe @B-Dalmijn is with leave and we want this included in the next release I've updated the branch. @savente93 Could you review this PR to have a second pair of eyes on my changes? Thanks!

savente93:
Sure thing I'll have a look right now

savente93 (approved):
LGTM

DirkEilander (approved):

PR #443, opened 2023-07-13
Fix `make dev` for mamba 

### LINKED ISSUE #436: BUG: Makefile create environments does not work with conda/mamba
### HydroMT version checks

### Reproducible Example

```
export PY_ENV_MANAGER=conda
make dev
```
> conda create: error: one of the arguments -n/--name -p/--prefix is required

### Current behaviour

Creating environments does not work with conda or mamba (haven't tested micromamba). With conda/mamba instead of `mamba create -f full-environment.yml -y` the syntax is `mamba env create -f full-environment.yml` 

### Desired behaviour

- [ ] the make file should work on linux and window with mamba as default. Most of use use mambaforge as distribution.
- [ ] it should create an [editable install](https://pip.pypa.io/en/stable/topics/local-project-installs/) with `pip install -e .` (only for dev)
- [ ] the flavour should be reflected in the environment name, e.g., hydromt-full 

Note: these changes should also be reflected in the docs in the docs.

### Additional context

_No response_

savente93:
## Issue addressed
Fixes #436 

## Explanation
Simply adjusted the make commands to use the `mamba` api instead of `micromamba` as that should be the default. Tested the install on windows, and that seems to go fine as well, so should all be good 

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-PY_ENV_MANAGER		?= micromamba
+PY_ENV_MANAGER		?= mamba
 DOCKER_USER_NAME 	?= deltares
 OPT_DEPS			?= ""
 SPHINXBUILD   	 	 = sphinx-build
@@ -9,16 +9,9 @@ BUILDDIR      	 	 = docs/_build
 .PHONY: clean html
 
 dev: full-environment.yml
-	$(PY_ENV_MANAGER) create -f full-environment.yml -y
-	$(PY_ENV_MANAGER) -n hydromt run pip install .
-	$(PY_ENV_MANAGER) -n hydromt run pre-commit install
-
-env:
-	@# the subst is to make sure the is always exactly one "" around OPT_DEPS so people can
-	@# specify it both as OPT_DEPS=extra,io and OPT_DEPS="extra,io"
-	python3 make_env.py "$(subst ",,$(OPT_DEPS))"
-	$(PY_ENV_MANAGER) create -f environment.yml -y
-	$(PY_ENV_MANAGER) -n hydromt run pip install .
+	$(PY_ENV_MANAGER) env create -f full-environment.yml
+	$(PY_ENV_MANAGER) run -n hydromt-full pip install -e .
+	$(PY_ENV_MANAGER) run -n hydromt-full pre-commit install
 
 min-environment.yml:
 	pip install tomli
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 PY_ENV_MANAGER		?= mamba
 DOCKER_USER_NAME 	?= deltares
 OPT_DEPS			?= ""
+ENV_NAME			?= hydromt
 SPHINXBUILD   	 	 = sphinx-build
 SPHINXPROJ    	 	 = hydromt
 SOURCEDIR     	 	 = docs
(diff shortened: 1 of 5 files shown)


docs/dev/dev_install.rst
```diff
@@ -97,8 +97,8 @@ After the environment file has been created you can create an environment out of
 
 .. code-block:: console
 
-    $ micromamba env create -f environment.yml
-    $ micromamba activate hydromt
+    $ mamba env create -f environment.yml
+    $ mamba activate hydromt
```
DirkEilander, line 101: Should this be hydromt-full if we create it with `make dev`?
In the text above I think it's also better to focus on the flavours only. We can move the more detailed descriptions to a new section further below. 
```suggestion
    $ mamba activate hydromt-full
```

Makefile
```diff
@@ -9,16 +9,9 @@ BUILDDIR      	 	 = docs/_build
...
-
-env:
-	@# the subst is to make sure the is always exactly one "" around OPT_DEPS so people can
-	@# specify it both as OPT_DEPS=extra,io and OPT_DEPS="extra,io"
-	python3 make_env.py "$(subst ",,$(OPT_DEPS))"
-	$(PY_ENV_MANAGER) create -f environment.yml -y
-	$(PY_ENV_MANAGER) -n hydromt run pip install .
+	$(PY_ENV_MANAGER) env create -f full-environment.yml
+	$(PY_ENV_MANAGER) run -n hydromt-full pip install -e .
+	$(PY_ENV_MANAGER) run -n hydromt-full pre-commit install
```
DirkEilander, line 15: I assume this was deleted because of the changes i requested in make_env.py? Does it make sense to keep it but for single dependency groups or flavours only? If we delete it we should also remove a mention of it in the (dev/dev_install.rst) and the OPT_DEPS global argument above.

DirkEilander (changes requested):
Thanks for looking into this @savente93 ! 

- I tested it on windows with a miniconda installation and wsl with a mambaforge installation. The miniforge installation. I noticed that the miniconda installation on windows does not have a python3 symlink like. With my wsl mambafore installation python and python3 both point to the same python. Would it make sense to replace python3 by python or is this something I should deal with?
- I also noticed that "old" environment.yml files are not overwritten by `python make_env.py`. This might be unwanted if we want to ensure that you always install the latest deps.

Makefile
savente93: good point, I originally removed it just because it felt a little redundant with the flavours but on reading your comment I think it is still a good one to keep so I"ll add it back in. The one downside with it is that the environment will have a slightly weird name like `hydromt-extra,io` but I'm not sure that can be helped.

docs/dev/dev_install.rst
savente93: good point, I think I forgot to update that one

savente93:
> Would it make sense to replace python3 by python or is this something I should deal with?
sorry for this, the different distributions and what features/aliases/APIs they have is a bit overwhelming/confusing to me so I just missed this. I think it's good to just move it to the `python` alias, so I've updated it. I can't remember why I put that in there, but refering just to python and letting the user's system determine the correct one I think is the correct move. 

> I also noticed that "old" environment.yml files are not overwritten by python make_env.py. This might be unwanted if we want to ensure that you always install the latest deps.
I'm not sure what you mean here, because it does overwrite the files. you can use the `-o` option to specify the destination and it should overwrite whatever is already there. If you mean that we should clean up an `environment.yml` if someone runs `make_env.py` to make a different environment I disagree there, I don't want my tools to just randomly delete other files that it has deemed "no longer necessary" unless I tell it to. That can lead to big frustrations.

DirkEilander (approved):
Thanks for this PR @savente93. While testing/reviewing I've taken the liberty to go one step further with this PR
- I generalized the make_env.py script such that it can potentially be applied to other packages as well. Some info has been moved to the pyproject.toml file `[tool.pyproject2conda]` section. 
- I've also undone the changes which automatically generated the environment name as requested earlier by me as this yields issues in the github workflows and Dockerfile where environments named hydromt are expected. Instead the `make dev` method creates a hydromt-dev environment while an optional ENV_NAME can be provide to name the environment with `make env`. 
- I've restructured the dev_install.rst file a bit to make the different options make vs manual a bit more clear.

Let me know what you think. If you agree with my changes this is ready to be merged.

savente93:
having the pip deps in pyproject.toml is actually a really nice one, thanks for that, I like it a lot better! I'm not familiar with the tool itself, but if you know and trust it that's good enough for me. The CIs can be adjusted for the new names down the line if you want, but given that this should probably be in the release, and this works, I think this is fine for now. nice work!

Issue #152, opened 2022-07-29
Kwargs with None need to be passed using 'null' in yml files


JoostBuitink:
Currently, when passing `None` to a kwargs, the yml parser does not convert this to a Python `None` object. Instead, `null` should be specified in the yml file.

One possible solution is to accept for `None` in the yml file, and replace those with `null` before parsing to the yml reader. This would also mean that `null` values should be replaced with None before writing to a yml file.

hboisgon:
Not sure if we should change behavior for None or just write better doc about this ? Maybe good to change still as in our docstrings None is used everywhere

DirkEilander:
Proposed solution: document that None is not accepted and if it is a valid value for an argument it should be the default value. in yml files where you explicitly want to show that you use the default None that line could be commented in that case. In addition we can catch this error to throw a better error.

PR #444, opened 2023-07-14
Add support for configure using TOML files

### LINKED ISSUE #376: ENH: Support toml format in config.configread
### Feature Type

- [ ] Adding new functionality to HydroMT

- [X] Changing existing functionality in HydroMT

### Enhancement Description

Quite some models use TOML as format for their configuration file so would be good if we support it as well.
I think for now we only support yaml and ini.

We could still make toml library an optional library.

### Feature Description

For configread, add one more line to check toml extension:

https://github.com/Deltares/hydromt/blob/0def737fbdf4fa2fb7891c790ad0fbcaebb84733/hydromt/config.py#L51-L55

Same for configwrite.

Functions to read/write toml can be found in hydromt_wflow:
```
    def _configread(self, fn):
        with codecs.open(fn, "r", encoding="utf-8") as f:
            fdict = toml.load(f)
        return fdict

    def _configwrite(self, fn):
        with codecs.open(fn, "w", encoding="utf-8") as f:
            toml.dump(self.config, f)
```

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #376 

## Explanation
Very simple implementation, because almost everything is done using dictionaries all that was needed was just swap the yaml functionality for the toml. (doc updates to come) 

## Checklist


savente93:
Since Toml doesn't support any type of `None` value, I also took the liberty to Fix #152

DirkEilander:
Thanks for this this PR @savente93! I noticed that the PR does more than just fixing #376, which requests toml support for model configuration files in the Model.read_config / write_config methods. This PR also adds toml support for HydroMT data catalog and configuration files. We should have a discussion with the team/users if this is indeed wanted. 

My initial thoughts: 
In general I can imagine that different file formats will only confuse users. 
For the HydroMT configuration file I see a benefit though. If the model configuration is in toml (like for most PMT HYD models) it might actually be convenient for users if the HydroMT configuration can also be done in toml. 
For the HydroMT data catalog file I think we should stick to toml as it has many more "layers" for which I think toml is not very readable.  Let's discuss at the next stand-up how to proceed with this.

What do you think @hboisgon @alimeshgi @B-Dalmijn @Tjalling-dejong ?

alimeshgi:
I agree that using different file formats can potentially confuse users. Consistency in file formats often helps maintain simplicity and ease of use. Yes, let's discuss it in the next stand-up.

savente93:
Converted to draft until we can reach a consensus on https://github.com/Deltares/hydromt/discussions/452

DirkEilander (approved):
LGTM!

PR #456, opened 2023-07-25
return kwargs from adapter.to_file funtions

### LINKED ISSUE #182: DataAdapter.to_file methods should return driver "kwargs"
... for the DataCatalog.export_data method to be able to properly parse e.g. csv data after writing to file.

https://github.com/Deltares/hydromt/blob/main/hydromt/data_catalog.py#L472

savente93:
## Issue addressed
Fixes #182

## Explanation
The description wasn't super clear about how this was supposed to be used but it was super quick to do, so I just took the liberty. For now the returned source_kwargs in the export function is unused, but hopefully, those (i.e. @DirkEilander) can use it now that it is available. This PR doesn't actually add any functionality, therefore I don't think tests or doc updates are necessary. 

## Checklist


hydromt/data_adapter/dataframe.py
```diff
@@ -162,20 +164,21 @@ def to_file(
...
             obj.to_csv(fn_out, **kwargs)
+            kwargs["index_col"] = obj.index.name
         elif driver == "excel":
             fn_out = join(data_root, f"{data_name}.xlsx")
             obj.to_excel(fn_out, **kwargs)
         else:
             raise ValueError(f"DataFrame: Driver {driver} is unknown.")
 
-        return fn_out, driver
+        return fn_out, driver, kwargs
```
DirkEilander, line 181:
```suggestion
        return fn_out, driver, read_kwargs
```

hydromt/data_adapter/dataframe.py
```diff
@@ -162,20 +164,21 @@ def to_file(
...
             logger.warning(str(err))
-            return None, None
+            return None, None, None
 
         if driver is None or driver == "csv":
             # always write as CSV
             driver = "csv"
             fn_out = join(data_root, f"{data_name}.csv")
             obj.to_csv(fn_out, **kwargs)
+            kwargs["index_col"] = obj.index.name
```
DirkEilander, line 174: The kwargs in this method are used for writing a file (here csv) and not necessarily the same as for reading the same file. See different arguments in e.g. `pandas.DataFrame.to_csv` and `pandas.read_csv`. The kwargs that are returned here are later used to read back the same file and parse it into an identical python object (here pands.DataFrame). I would suggest to return a new dictionary with `read_kwargs`, which you can initialize at the start of the `to_file` method. Idem for the other DataAdapters. I agree with your approach to start with the read_kwargs that we know are required. We can easily add once we have this infrastructure.


```suggestion
       read_kwargs = dict()
        if driver is None or driver == "csv":
            # always write as CSV
            driver = "csv"
            fn_out = join(data_root, f"{data_name}.csv")
            obj.to_csv(fn_out, **kwargs)
            read_kwargs["index_col"] = obj.index.name
```

DirkEilander (changes requested):
I added a small comment, let me know if you have questions.

savente93:
@DirkEilander Perhaps a tangential question, but I noticed that the `export_data` in the `test_export_global_datasets` raises the following warning: `UserWarning: using the driver setting is deprecated. Please use vector_table instead.` I'm wondering if this is something we can solve here now? We would check if the export we're doing is  a `csv` based one but that feel really inelegant. What do you think?

savente93:
@DirkEilander implemented what we discussed. Should be good to go now.

DirkEilander (approved):
Made a small change to actually use the returned driver_kwargs.
Otherwise goot to be merged!

PR #457, opened 2023-07-25
Add version switch to docs

### LINKED ISSUE #188: Add a version switcher on the docs
I just checked: pydata-sphinx-theme has version switcher support now 🎉 

https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/version-dropdown.html

This required some custom javascript / templates before, but it seems quite convenient now!

savente93:
## Issue addressed
Fixes #188

## Explanation
simplay followed the link and followed the docs. Works just fine. The only downside is that the older versions of the docs don't have this added so once you switch, the menu for it disapears. Can go back and retroactively add this back to the older versions if necessary, but this could give issues with reproducability (e.g. not sure if this plugin is compatable with older doc versions.) 

## Checklist

Currently the url of the version switcher points to a file that will be deleted when the branch is deleted. Once this is approved, I'll switch it over to where the file will live after merging. This will temporary break the docs between update and merging as the file needs to be found for the docs to be generated, but should this should be fixed as soon as the merge is complete, so it should be fine.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -234,6 +234,8 @@ autoclass_content = "both"
 # Add any paths that contain custom static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
+bare_version = hydromt.__version__
+doc_version = bare_version[: bare_version.find("dev") - 1]
 html_static_path = ["_static"]
 html_css_files = ["theme-deltares.css"]
 html_theme_options = {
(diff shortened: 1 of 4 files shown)


docs/conf.py
```diff
@@ -257,14 +259,18 @@ def clean_str(s):
     "logo": {
         "text": "HydroMT Core",
     },
-    "navbar_end": ["navbar-icon-links"],  # remove dark mode switch
+    "navbar_end": ["navbar-icon-links", "version-switcher"],  # remove dark mode switch
+    "switcher": {
+        "json_url": "https://raw.githubusercontent.com/Deltares/hydromt/doc-version-switcher/docs/_static/switcher.json",
```
hboisgon, line 264: I assume this is the link you will update? Then all looks good to me :)

hboisgon (approved):
Nice change! Thanks @savente93

DirkEilander (approved):
Looks good to me! Just one small addition, could you add a step to the release guide docs to make sure the switcher.json file gets updated on every release?

DirkEilander:
>  The only downside is that the older versions of the docs don't have this added so once you switch, the menu for it disapears. 

Can't we add this file to the root of the gh-pages branch directly? The url would than be "https://github.com/Deltares/hydromt/blob/gh-pages/switcher.json" and that should work for all versions right?

savente93:
> > The only downside is that the older versions of the docs don't have this added so once you switch, the menu for it disapears.
> 
> Can't we add this file to the root of the gh-pages branch directly? The url would than be "https://github.com/Deltares/hydromt/blob/gh-pages/switcher.json" and that should work for all versions right?

That's actually a really good idea. Although it won't make it appear on the older versions, that is a config setting, so we'd have to go back to all the previous versions of the docs and add it to the config to solve this issue

docs/conf.py
savente93: indeed!

PR #460, opened 2023-07-27
Read of model when update in same root

### LINKED ISSUE #256: Behavior of model read when update: redefine as prone to errors?
When updating a model, we only do a full read if the model output folder is different than the model to update.
This means that when updating a model component (eg maps), the existing ones are not "officially" read before being updated and could therefore be completely overwritten.

Current flow:

- update in model api: read not started without a model_out option
- setup_maps_from_*: read new data and process
- set_maps(): if self._maps empty (which is the case without read) then assign data from the setup function directly
- write_maps(): uses xr.to_netcdf() with default mode is 'w' for write and not append which results in existing maps being overwritten

I tried to check why with wflow and staticmaps we never got that with setup_lulcmaps.
The reading of staticmaps is done unexplicitely in two points: 

- in update: when calling self.region, this starts a read of config + staticmaps
- in setup_lulcmaps: for reproject_like we use as argument ds_like = self.staticmaps which will force the read.

I think we should explicitly take care of the read within the model_api.
A way to do that is to maybe when calling set_* methods like set_maps start with a read if self._assert_read_mode
This still allows to avoid a costly full model read if unnecessary but makes sure model data will not be overwritten when updating.

Using this will however potentially change the behavior for updating forcing (for eg wflow) but I think this would be okay (either all components precip. temp, pet would be replaced or you would with this allow to replace precip only for exemple).

@DirkEilander what are your thoughts?

hboisgon:
## Issue addressed
Fixes #256

## Explanation
See #256 for discussions

## Checklist


savente93:
@savente93 pick this up

hboisgon:
@savente93 I put some placeholders in the code where I think you can try to add filehandling. Thanks again in advance for the help!

hboisgon:
Hi @savente93 , thanks for the help! I tested and seems to work like a charm for our test bench :)
The only thing I am not too sure about is what the `forceful_cleanup` option does? It seems it's set to False anyway in the tests and works. 
If it's needed, please leave and then I added a CLI --fo option (similar to build) so that CLI user can also set it.

After that, could you do the final review? Unless @DirkEilander you prefer to also have a look?

savente93:
To give a bit more context for the `forced_cleanup` flag: what the code does is just try to overwrite the destination file with the temporary one until either the maximum number of tries is reached or until it succeeds which is meant to solve timing issues (e.g. the file was closed but we still got a permission error because of data races). The forced cleanup also attempts to close the original file handle, which may or may not be necessary, but could cause trouble if the user will try to read from the same file handle after this function is called. it just seemed like a good idea to give the user a bit more flexibility, but if this is not necessary we could just remove it. Happy to defer to your judgement

hboisgon:
Thanks! Then it's clear to me and I think a good idea indeed so let's keep it :)
It's ready for review then

hydromt/cli/main.py
```diff
@@ -255,6 +256,7 @@ def update(
     opt,
     data,
     dd,
+    fo,
```
savente93, line 259: why `fo` and not `fc` though I wonder?

hydromt/models/model_grid.py
```diff
@@ -374,8 +375,12 @@ def read_grid(self, fn: str = "grid/grid.nc", **kwargs) -> None:
             Additional keyword arguments to be passed to the `_read_nc` method.
         """
         self._assert_read_mode
-        for ds in self._read_nc(fn, **kwargs).values():
-            self.set_grid(ds)
+        # Load grid data in r+ mode to allow overwritting netcdf files
+        if self._read and self._write:
+            kwargs["load"] = True
+        for ds in self._read_nc(fn, single_var_as_array=False, **kwargs).values():
```
savente93, line 381: I'm not 100% sure how python handle's function calls like this, I think it might be more efficient to do something like this: 
```suggesion
loaded_nc_files = self._read_nc(fn, single_var_as_array=False, **kwargs)
for ds in loaded_nc_filies.values():
```

savente93 (approved):
few small comments, but other than that, I think she's ready to go chief!

hydromt/models/model_api.py
```diff
@@ -115,12 +125,41 @@ def __init__(
...
         self._read = True
         self._write = False
+        self._defered_file_closes = []
 
         # model paths
         self._config_fn = self._CONF if config_fn is None else config_fn
         self.set_root(root, mode)  # also creates hydromt.log file
         self.logger.info(f"Initializing {self._NAME} model from {dist} (v{version}).")
 
+    def _cleanup(self, forceful_cleanup=False, max_close_attempts=2) -> List[str]:
```
DirkEilander, line 135: Could you add a docsting briefly explaining what this method does?
I think I understand, but that's always helpful for future reference

hydromt/models/model_api.py
```diff
@@ -1336,7 +1385,7 @@ def write_states(self, fn="states/{name}.nc", **kwargs) -> None:
             filename relative to model root and should contain a {name} placeholder,
             by default 'states/{name}.nc'
         **kwargs:
-            Additional keyword arguments that are passed to the `RasterDatasetAdapter`
+            Additional keyword arguments that are passed to the `_write_nc`
```
DirkEilander, line 1388: Should we refer to internal function `_write_nc` in the docstring, which itself does not have a docstring?
This method could be added to the external api (remove underscore) with docstrings. It could potentially even be moved to the io.py script. Alternately it we mention only the underlying xarray.to_netcdf method here. That's were most kwargs go to.

hydromt/models/model_api.py
```diff
@@ -28,6 +30,14 @@
 
 logger = logging.getLogger(__name__)
 
+DeferedFileClose = TypedDict(
+    "DeferedFileClose",
+    {"ds": xr.Dataset, "org_fn": str, "tmp_fn": str, "close_attempts": int},
+)
+
+
+TMP_DATA_DIR = TemporaryDirectory()
```
DirkEilander, line 39: This creates a tempdir every time we import hydromt right? Is that what we want?
Could we also create a temp dir only when necessary in the _write_nc method. Possibly even located inside the model root?

DirkEilander (commented):
Nice work @hboisgon and @savente93 !
I've added a few questions / comments, but nothing major.

hydromt/models/model_api.py
savente93: Added a docstring, essentially containing my comment above.

hydromt/models/model_api.py
savente93: Maybe it's a reflex, but if possible I'd prefer keeping things private as much as we can, so I would prefer refering to `xarray.to_netcdf` instead. @hboisgon what do you think?

hydromt/cli/main.py
hboisgon: It means force overwrite and we define that flag for the build method (in that case you also overwrite files but in build mode you did not try to open them first), so I thought I could re-use the flag. If it bothers you you could also rename forceful_cleanup to force_overwrite as well for consistency?

hydromt/models/model_grid.py
hboisgon: I dont know for sure but if you think it's better, we can change.

hydromt/models/model_api.py
hboisgon: I saw that we referenced it in a lot of places already but without an API docstring. The thing is that for the equivalent _read_nc method, the new 'load' option I implemented is actually then not passed to the open_dataset **kwargs so I have to document this better anyhow. I think these functions could easily be moved to io.py if you also agree?

hydromt/cli/main.py
savente93: ahhh that makes sense. Yeah I think it's better to stay consistent, but I don't mind doing the renaming

hydromt/models/model_api.py
savente93: I don't have strong feelings on moving it to `io.py` but if there is a way you think that the documentation could be improved, then I'm always happy with that. I'll leave it up to you though, I trust your judgement.

hydromt/cli/main.py
savente93: Done :)

hydromt/models/model_api.py
hboisgon: I did not realize but as @savente93 implemented the nice temporary file writting and saving to defered file list, it became quite tough to move this function to ``io.py``. So in the end I decide to leave both ``read_nc`` and ``write_nc`` in model_api.py, removed the underscore and added docstrings. Should all be consistent now!

PR #473, opened 2023-08-08
Avoid RecursionError when using Model.set methods inside Model.read methods

### LINKED ISSUE #472: The recent update in #460 breaks HydroMT-SFINCS
### HydroMT version checks

### Reproducible Example

```
from hydromt import Model, raster

class MyModel(Model):
    def __init__(self, root=None, mode="r"):
        super().__init__(root, mode=mode)

    def read_maps(
        self,
    ) -> None:
        da = raster.full_from_transform(
            transform=[0.5, 0.0, 3.0, 0.0, -0.5, -9.0],
            shape=(4, 6),
        )
        # this will cause a recursion error in read-mode
        self.set_maps(da, name="test")
        # and is preferred over the current implementation
        # which does not make use of the checks in set_maps method
        self._maps["test"] = da

mod = MyModel(mode="r+")
mod.read_maps() # RecursionError 
```

### Current behaviour

You get a RecursionError when using set methods inside read methods for any model component since #460 
This is used a lot in plugins and the new implementation causes errors downstream, see e.g. https://github.com/Deltares/hydromt_sfincs/actions/runs/5797018073/job/15711743003

### Desired behaviour

I prefer to use the set_ methods inside the read methods to avoid duplicating code that does checks on the data. For developers implementing hydromt for their model I've always explicitly advised to avoid setting directly using `self._maps[name] = data` for the same reason.

### Additional context

_No response_

DirkEilander:
## Issue addressed
Fixes #472 

## Explanation
I moved the initialization of the model component to the property definition. 
When called the first time, we initialize the component and try to read it if in read-mode.
From the second time forward a call of the property returns the initialized component without trying to read.
Now we can use the `set_` methods again inside the `read_` methods like we used to do.

## Checklist

This could also break code downstream, if the internal properties are called (e.g. `self._maps` instead of `self.maps`). In the new implementation calling the internal static properties should be restricted to the `set_` methods and the dynamic property methods only which happen in the core. Hence, an easy fix should be to replace all calls to the internal static property by the dynamic property.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -21,7 +21,8 @@ Added
 Changed
 -------
 - possibility to ``load`` the data in the model read_ functions for netcdf files (default for read_grid in r+ mode). (PR #460)
--
+- Internal model components (e.g. `Models._maps`, `GridModel._grid``) are now initialized with None and should not be accessed directly,
+  call the corresponding model property  (e.g. `Model.maps`, `GridModel.grid`) instead. (PR #473)
 
 Fixed
 -----
--- a/hydromt/models/model_api.py
+++ b/hydromt/models/model_api.py
@@ -107,17 +107,17 @@ class Model(object, metaclass=ABCMeta):
         # placeholders
         # metadata maps that can be at different resolutions
         # TODO do we want read/write maps?
-        self._config = dict()  # nested dictionary
-        self._maps = dict()  # dictionary of xr.DataArray and/or xr.Dataset
+        self._config = None  # nested dictionary
+        self._maps = None  # dictionary of xr.DataArray and/or xr.Dataset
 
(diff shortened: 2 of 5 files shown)


savente93 (approved):
Good to see that you added a test for this. I see nothing out of the ordinary, so if this test reproduces the problem and passes with this patch then it should be ready to merge!

hboisgon (commented):
Nice and more elegant way to fix model reading methods :) with this we might be changing behavior on what to expect with an empty object (eg you had to modify self.res and self.dims etc) so these changes may affect the plugins. So I think good to document the changes in the changelog.
I also discussed with @DirkEilander that the changes for Mesh will be applied in PR #412 after this has been merged.

So good to merge once the changelog is updated :)

PR #479, opened 2023-08-10
Region from datacatalog

DirkEilander:
## Issue addressed
This PR uses the DataCatalog to read region arguments for geom and grid. This enables to set the region in a data catalog file (useful in case of HydroMT-SFINCS where we sometimes reuse the region geometry in other setup methods) and also adds the file to the used sources to improve reproducibility.

## Checklist


hydromt/workflows/basin_mask.py
```diff
@@ -15,18 +15,18 @@
...
-from ..io import open_raster
 from ..models import MODELS
 
 logger = logging.getLogger(__name__)
 
 __all__ = ["get_basin_geometry", "parse_region"]
 
 
-def parse_region(region, logger=logger):
+def parse_region(region, logger=logger, data_catalog=DataCatalog()):
```
savente93, line 29: putting a new datacatalog as a default argument is going to create a new one every time this is parsed, but not every time the funciton gets called, which makes the control flow very confusing. If you want to keep a default argument like this, you should set it to `None` and make a new object in the case this is None. Personally I'd be in favor of making the data_catalog a required argument.

hydromt/workflows/basin_mask.py
```diff
@@ -158,7 +155,7 @@ def parse_region(region, logger=logger):
     return kind, kwargs
 
 
-def _parse_region_value(value):
+def _parse_region_value(value, data_catalog=DataCatalog()):
```
savente93, line 158: same comment as `parse_region`

hydromt/workflows/basin_mask.py
```diff
@@ -173,12 +170,13 @@ def _parse_region_value(value):
...
-        kwarg = dict(geom=gpd.read_file(value))
-    elif isinstance(value, gpd.GeoDataFrame):  # geometry
-        kwarg = dict(geom=value)
     elif isinstance(value, (str, Path)) and isdir(value):
         kwarg = dict(root=value)
+    elif isinstance(value, (str, Path)):
+        geom = data_catalog.get_geodataframe(value)
+        kwarg = dict(geom=geom)
+    elif isinstance(value, gpd.GeoDataFrame):  # geometry
+        kwarg = dict(geom=value)
```
savente93, line 179: please add a new line, as the next `if` is presumably unrelated to this chain. I think it would also be good to add an `else` branch which raises a `ValueError` if someone passes in an unknown type

savente93 (changes requested):
Couple small comments but otherwise looks good!

savente93 (approved):
Looking all good now!

Issue #372, opened 2023-06-01
ENH: improve adapter for GeoDataset


hboisgon:
### Feature Type

- [X] Adding new functionality to HydroMT

- [ ] Changing existing functionality in HydroMT

- [ ] Removing existing functionality in HydroMT

### Enhancement Description

Currently the GeoDataset adapter is quite limited if the files are not in direct netcdf format.
It would be nice to support a little more options:

- fn_data can be netcdf file instead of csv only (we say in the dosctrings it can but this is not the case, see link below)
- fn_data can be a data_catalog entry: this would be handy especially to allow reading options to properly parse csv file (now format as to be exactly what the reader requires)
- fn_data can be multiple (csv) files: for now geodataset require one file per variable with all IDs but for some data this could be organized as one file per ID (with all variables)

https://github.com/Deltares/hydromt/blob/0def737fbdf4fa2fb7891c790ad0fbcaebb84733/hydromt/io.py#L309-L313

### Feature Description

Parts can be implemented in io.open_geodataset and other parts in geodataset.get_data.

Tjalling-dejong:
@DirkEilander I am bit lost on how option 2 was a solution again? This option does not allow for using a datacatalog entry for instance.

Tjalling-dejong:
I moved the third point on multiple fn_data files to a separate issue: #453

DirkEilander:
> @DirkEilander I am bit lost on how option 2 was a solution again? This option does not allow for using a datacatalog entry for instance.

Correct, it is not a solution to the issues, but a way to keep the io.open_geodataset method as a driver next to option 1. However, we need to discuss if we want to keep the open_geodataset method. We also discussed that adding additional drivers to read nc  inside this method is not ideal as it may become very unclear if a user wants to pass arguments to the different drivers. We also agreed passing DataAdapters to an io function is not a good idea. IO methods (a.k.a. drivers) are called from the Adapters and not the other way around.

DirkEilander:
- We keep GeoDatasetAdapter, but only for single files (combined locs and data). This means deprecating the vector driver and the `io.open_geodataset` method. 
- We create a new DatasetAdapter for non-spatial N-Dim data because pandas only works 'nicely' with 2D data. To start with this would support the classic netcdf/zarr and the `io.open_timeseries_from_table` method. The `DataCatalog.get_dataset` methods should have an index argument to filter on.
- Add `{index}`  to resolve path method options, to read files based on index
[...]

PR #486, opened 2023-08-23
Combining multiple CSVs into one Dataset

### LINKED ISSUE #476: New method/driver to read and combine multiple csv files into a single xr.Dataset
### Kind of request

Adding new functionality

### Enhancement Description

This issue is a sub-issue of the discussions in #372.

Need a new method in io.py
This could look like the below.

````
def open_mfcsv(
   fns: List[Path] files to open (not sure yet if we keep the wildcard or use the list out of resolve_path method,
   path_format: str path argument of the hydromt data catalog with the placeholders not parsed yet ({variable}, {year}, {index})
   driver_kwargs: dict argument to pass to pandas.read_csv to parse the files
   concat_dim: List or str ?
): ->xr.Dataset
````

Not fully sure yet on how to combine the files.
Ideas:

- have a formatted path with placeholders or wildcard to help identify - or use assume one file and single concat dim use a default index based on range of number of files
- assume that time dates are parsed (time dimension should be called time) and variables should also be okay to be automatically merged - issue is more the additional index dim (eg station ID, soil layer)

### Use case

Hydrometeo stations with one csv file per location (example of GB_CAMELS dataset).

### Additional Context

_No response_

### PR DESCRIPTION
**Pull request by savente93**
## Issue addressed
Fixes #476 

## Explanation

## Checklist

### DIFF
--- a/hydromt/io.py
+++ b/hydromt/io.py
@@ -4,7 +4,7 @@ import io
 import logging
 from os.path import abspath, basename, dirname, isfile, join, splitext
 from pathlib import Path
-from typing import Any, Dict, Literal
+from typing import Any, Dict, Literal, Union
 
 import dask
 import geopandas as gpd
@@ -229,7 +229,7 @@ def open_mfraster(
 
 
 def open_mfcsv(
-    fns: Dict[str | int, str | Path],
+    fns: Dict[Union[str, int], Union[str, Path]],
     concat_dim: str,
     driver_kwargs: Dict[str, Any],
     variable_axis: Literal[0, 1] = 1,
--- a/hydromt/io.py
+++ b/hydromt/io.py
@@ -4,7 +4,7 @@ import io
 import logging
 from os.path import abspath, basename, dirname, isfile, join, splitext
 from pathlib import Path
-from typing import Any, Dict, Literal, Union
+from typing import Any, Dict, Literal, Optional, Union
 
 import dask
 import geopandas as gpd
@@ -231,7 +231,7 @@ def open_mfraster(
 def open_mfcsv(
     fns: Dict[Union[str, int], Union[str, Path]],
     concat_dim: str,
-    driver_kwargs: Dict[str, Any],
+    driver_kwargs: Optional[Dict[str, Any]] = None,
     variable_axis: Literal[0, 1] = 1,
 ) -> xr.Dataset:
     """Open multiple csv files as single Dataset.
@@ -262,7 +262,9 @@ def open_mfcsv(
     # we're gonna use the structure of the first file found to check
     # all others against
     csv_kwargs = {"index_col": 0}
-    csv_kwargs.update(**driver_kwargs)
+    if driver_kwargs is not None:
+        csv_kwargs.update(**driver_kwargs)
+
     first_id, first_fn = next(iter(fns.items()))
     first_df = pd.read_csv(first_fn, **csv_kwargs)
     if variable_axis == 0:
@@ -284,10 +286,10 @@ def open_mfcsv(
 
         df[concat_dim] = id
 
-        if not df.index.equals(first_index):
+        if not df.index.dtype() == first_index.dtype():
             raise RuntimeError(
-                f"file {fn} has inconsistent index: {df.index}"
-                f"Expected {first_index}"
+                f"file {fn} has inconsistent index type: {df.index.dtype()}"
+                f"Expected {first_index.dtype()}"
             )
 
         dfs.append(df)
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: hydromt/io.py [hunk 1]
```diff
@@ -226,6 +228,79 @@ def open_mfraster(
     return ds
 
 
+def open_mfcsv(
+    fns: Dict[Union[str, int], Union[str, Path]],
+    concat_dim: str,
+    driver_kwargs: Dict[str, Any],
```
hboisgon, line 234: Maybe here default with empty dict instead of mandatory argument?

tests/test_io.py
```diff
@@ -189,6 +189,70 @@ def test_raster_io(tmpdir, rioda):
...
+@pytest.mark.skip(reason="Not yet supported")
+def test_open_mfcsv_by_var(tmpdir, dfs_segmented_by_vars):
+    os.mkdir(tmpdir.join("data"))
+    fns = {}
+    for var, df in dfs_segmented_by_vars.items():
+        fn = tmpdir.join("data", f"{var}.csv")
+        df.to_csv(fn)
+        fns[var] = fn
+
+    ds = hydromt.io.open_mfcsv(fns, {}, "id", "time", 0)
```
hboisgon, line 244: Update this test? (surprised that this works with the arguments in this order)

hboisgon (commented):
First small comment on the PR

hboisgon:
> So I think this solution works quite well for now. The only thing that is remaining to figure out, is how to handle the paths/wildcards and extract what we need from that. currently the `fns` needs to be a dictionary that provides a mapping of `id -> fn`. I think this should be fairly easily to do when we're expanding the wildcards etc. but I didn't implement it yet, because I'm not very sure how wildcards etc. are handled else where in the code base. So we just need to agree on how we want to handle that. @hboisgon Do you have thoughts or opinions on this?

@savente93 Thanks for the updates, I think more clear already!
For parsing id/concat_dim values based on fn, I thought maybe we could use a new wildcard in the data source path (like {id} or {index} like we now have {variable}, {year}) but I am not sure if this is easy to do. So using the data catalog resolve_path method you get the list of fns, and passing back the data catalog path (eg discharge_{id}_*.csv), you could get the id value out of the filename).
Again not sure if this is a good solution so we could also choose to delay when we implement the rest of the steps of issue #372

tests/test_io.py
savente93: I doesn't actually work. this test is skipped because I didn't yet implement this, but I didn't want to have to discard the test case. I'll update it though

savente93:
Thanks for you comments Hélène I'll see how much work they are and let you know.

savente93:
I think it will actually be easier to solve once we have a data catalog entry to test against. So I'm just going to open this  up for review as is, I can workshop the arguements etc, a bit once work statrs on the other issues. I'm happy to pick those up as well, but I would also appreciate some help in craftning the data catalog entry for the data set you'd like to include. (was it CAMEL-GB?) that way I know I'm building the correct mechanisms.

hydromt/io.py
```diff
@@ -226,6 +228,81 @@ def open_mfraster(
     return ds
 
 
+def open_mfcsv(
+    fns: Dict[Union[str, int], Union[str, Path]],
+    concat_dim: str,
+    driver_kwargs: Optional[Dict[str, Any]] = None,
```
hboisgon, line 234: You can also initialise with an empty dict rather than None

hydromt/io.py
```diff
@@ -226,6 +228,81 @@ def open_mfraster(
...
+    dfs = []
+    for id, fn in fns.items():
+        df = pd.read_csv(fn, **csv_kwargs)
+        if variable_axis == 0:
+            df = df.T
+
+        df[concat_dim] = id
+
+        if not df.index.dtype == first_index.dtype:
+            raise RuntimeError(
```
hboisgon, line 290: Maybe more a ValueError or TypeError?

hboisgon (approved):
Thanks @savente93 ! Looks good now! Let's do the id -> fn mapping when we work on the data catalog part (I can help you with that too of course!). I have two small comments but you can decide to ignore them and merge.

hydromt/io.py
savente93: yeah, I agree value error is a better one.

hydromt/io.py
savente93: it is considered really bad practices to initialise default arguments with things like lists or dictionaries. (e.g. https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) because it has really counterintuitive results. So unless there is a good reason for, I think it's better to keep the default as None.

PR #487, opened 2023-08-23
Fix api inherritence of models using __mro__

### LINKED ISSUE #253: Bug: definition of api attribute in model_api.py
### HydroMT version checks

### Reproducible Example
If a Model instance is a child of a hydromt Model subclass (example class DFlowFMModel(MeshModel)), then the api property is not fully parsed and only returns the api components of MeshModel and DFlowFMModel but not of Model.

Load a plugin that is a child of a submodel class (eg SfincsModel(GridModel) or HabitatModel(MeshModel) or HabitatGridModel(GridModel))

```
mod = HabitatGridModel(root="test", mode="w")
mod.api
```

### Current behaviour

Only returns the objects of HabitatGridModel and GridModel but not of Model:

```
{'grid': <class 'xarray.core.dataset.Dataset'>}
```

### Desired behaviour

Returns all attributes also of Model parent class:

```
{'grid': <class 'xarray.core.dataset.Dataset'>, 'crs': <class 'pyproj.crs.crs.CRS'>, 'config': typing.Dict[str, typing.Any], 'geoms': typing.Dict[str, geopandas.geodataframe.GeoDataFrame], 'maps': typing.Dict[str, typing.Union[xarray.core.dataarray.DataArray, xarray.core.dataset.Dataset]], 'forcing': typing.Dict[str, typing.Union[xarray.core.dataarray.DataArray, xarray.core.dataset.Dataset]], 'region': <class 'geopandas.geodataframe.GeoDataFrame'>, 'results': typing.Dict[str, typing.Union[xarray.core.dataarray.DataArray, xarray.core.dataset.Dataset]], 'states': typing.Dict[str, typing.Union[xarray.core.dataarray.DataArray, xarray.core.dataset.Dataset]]}
```

### Additional context
Problem seems to be in def api property in model_api.py

When running the following line, only MeshModel is found: https://github.com/Deltares/hydromt/blob/c5f7ac322e44a5381bede16664e21fe8d5074d82/hydromt/models/model_api.py#L119 

Repercussions are for the test for  _test_model_api and _test_equal that are used during testing of hydromt and plugins. Then only a small fraction is actually tested compared to the full models.
Apparently this is not used in the dashboards or docs, only for testing.

savente93:
## Issue addressed
Fixes #253 

## Explanation
I'll be honest, this is not the most elegant solution I've ever found but it does do the job. I'm not a fan of this design (the dictionary) but this solution does work, and refactoring the dictionary away could have unintended consequences that I don't have insight into so for now this is okay. However, if this continues to have problems, then we should seriously consider a redesign of this part of the system. 

The problem was that the dictionary was only being updated using the direct parent class and not any of the ancestors. This was fixed by using the `__mro__` (Method Resolution Order) attribute, which gives access to basically all ancestor classes as opposed to only the direct parent. 

## Checklist

### DIFF
--- a/hydromt/models/model_api.py
+++ b/hydromt/models/model_api.py
@@ -41,7 +41,6 @@ class Model(object, metaclass=ABCMeta):
 
     """General and basic API for models in HydroMT."""
 
-    # FIXME
     _DATADIR = ""  # path to the model data folder
     _NAME = "modelname"
     _CONF = "model.ini"
--- a/hydromt/models/model_api.py
+++ b/hydromt/models/model_api.py
@@ -25,6 +25,7 @@ from shapely.geometry import box
 from .. import config, log, workflows
 from ..data_catalog import DataCatalog
 from ..raster import GEO_MAP_COORD
+from ..utils import _classproperty
 
 __all__ = ["Model"]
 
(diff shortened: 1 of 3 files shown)


savente93:
oh also, since this is an internal thing as far as I can see, I didn't think it needed a changelog entry, but if the reviewer disagrees, let me know and I'll add it

hboisgon (approved):
LGTM

PR #490, opened 2023-08-28
add user warning when data crs conflicts with provided crs

### LINKED ISSUE #462: Add a debug message for ignoring user defined crs when reading raster
### Kind of request

### Enhancement Description

The feature improves the clarity of `RasterDatasetAdapter.get_data` method, especially for cases where local data are used. 

Current behavior:
In data_catalog.yml, when a raster file has crs in it, the user specified crs will not be used, and the user does not get a notification about this behavior. 

Expected behavior:
A debug message or a warning would be helpful to make it clear that crs specified by the user in data_catalog.yml is ignored. 

### Use case

When a user create a data_catalog.yml entry for a local data, they might input argument crs accompany the dataset. If there are mismatch between the dataset crs and argument crs, a user might have incorrectly expected the raster file would get the crs specified in the data catalogue, or they might have assigned the user raster crs by mistake. 

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #462 

## Explanation
Does what it says on the tin. Wasn't able to add a test for the rasterdataset adapter that the warning is actually emitted because the kwargs don't get passed to the underlying funtions always, but I figured it was good to keep the warning there anyway just to be sure. The logic is the same as with the other adapters so it should be fine. 

## Checklist

### DIFF
--- a/hydromt/data_adapter/geodataset.py
+++ b/hydromt/data_adapter/geodataset.py
@@ -323,8 +323,20 @@ class GeoDatasetAdapter(DataAdapter):
             ds_out = ds_out[variables]
 
         # set crs
-        if ds_out.vector.crs is None and self.crs is not None:
-            ds_out.vector.set_crs(self.crs)
+        if self.crs is not None:
+            if ds_out.vector.crs is None:
+                ds_out.vector.set_crs(self.crs)
+            elif ds_out.vector.crs != self.crs:
+                raise UserWarning(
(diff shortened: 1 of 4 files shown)


savente93:
Once again the step seems to fail due to some segfault I can't pin down but the tests do pass, so justas with theprevious, I'm just opening this one up for review anyway.

savente93:
Fixed in #481 instead

PR #495, opened 2023-08-30
add model version attr to Model base object

### LINKED ISSUE #397: Make HydroMT aware of model version
### Kind of request

Adding new functionality

### Enhancement Description

t would be nice to make HydroMT aware of which model/software version it is working with. 

### Use case

For example, depending on software version, some additional input data might be needed and vice versa, or format of the readers/writers may be different. Tracking versions could also enable easily to update one model from a software version to another.

### Additional Context

Needs refinement but I guess we could add a model property for version and maybe work on dictionaries for which setup functions to use (and output to produce) and which readers/writers. Could also work within the functions with if/else statements but that would not make update of a model from one version to another easy.

savente93:
## Issue addressed
Fixes #397 

## Explanation
From our discussion today during the sprint planning I understand that this should be enough to close the issue. If anything else also needs to be done on this please let me know.

## Checklist

### DIFF
--- a/hydromt/models/model_api.py
+++ b/hydromt/models/model_api.py
@@ -52,6 +52,8 @@ class Model(object, metaclass=ABCMeta):
     # TODO: change it back to setup_region and no res --> deprecation
     _CLI_ARGS = {"region": "setup_basemaps", "res": "setup_basemaps"}
     _TMP_DATA_DIR = None
+    # model version should be filled by the plugins
+    _MODEL_VERSION = None
 
     _API = {
         "crs": CRS,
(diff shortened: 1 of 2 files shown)


hydromt/models/model_api.py
```diff
@@ -52,6 +52,8 @@ class Model(object, metaclass=ABCMeta):
     # TODO: change it back to setup_region and no res --> deprecation
     _CLI_ARGS = {"region": "setup_basemaps", "res": "setup_basemaps"}
     _TMP_DATA_DIR = None
+    # model version should be filled by the plugins
+    _MODEL_VERSION = None
```
DirkEilander, line 56:
```suggestion
    # supported model version should be filled by the plugins
    # e.g. _MODEL_VERSION = ">=1.0, <1.1"
    _MODEL_VERSION = None
```

DirkEilander (changes requested):
Just added a bit more context in the code in one suggestion.
And can you also add this to the changelog?

savente93:
Good points thank you! Should be all good now :)

DirkEilander (approved):
Thanks!

PR #502, opened 2023-09-06
add tables property to base model class

### LINKED ISSUE #377: ENH: Add a new ``tables`` object to Model API
### Feature Type

Adding new functionality to HydroMT

### Enhancement Description

In HydroMT, we do not have an object for model tabular data which do not have a GIS dimension.
I think it would be nice to accommodate under a simple self.tables which would be a dict of pandas.DataFrame.

Both hydromt_wflow and hydromt_fiat are starting to need this.

### Feature Description

Would need to add _tables, tables, read/write/set_tables methods.

Examples that are quite generic are available in [hydromt_wflow](https://github.com/Deltares/hydromt_wflow/blob/rating_curve/hydromt_wflow/wflow.py)

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #377 

## Explanation
Copied and extended the code from Wflow as referenced in the issue. It seemed like a good idea to support more file types than just csv, so I implemented so common ones, but others can easily be added if necessary. 

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -128,3 +128,4 @@ dask-worker-space/
 #ruff linting
 .ruff_cache
 .envrc
+pyrightconfig.json
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,7 @@ Added
 - new ``force-overwrite`` option in ``hydromt update`` CLI to force overwritting updated netcdf files. (PR #460)
 - add ``open_mfcsv`` function in ``io`` module for combining multiple CSV files into one dataset. (PR #486)
 - Adapters can now clip data that is passed through a python object the same way as through the data catalog. (PR #481)
+- Model class now has methods for getting, setting, reading and writing arbitrary tabular data. (PR #502)
 
 Changed
 -------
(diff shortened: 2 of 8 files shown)


DirkEilander (changes requested):
Thanks for this PR @savente93! A couple of comments to discuss

- You deviated from the approach of other model properties by not implementing a _dynamic_ (not sure this is the right terminology) public `tables` property and initiating the internal `_tables` property with None. Using a _dynamic_ `tables` property allows to trigger reading upon first call (i.e. when the _tables property is still None; this should also be changed).  See e.g. `Model.maps` property.
- You also implemented a different get methods and an iter methods which we don't have for other properties. I think we can keep those but these should point to the `tables` instead of `_tables` property to trigger the reading.
- I'm also not sure if we need to support different drivers in the read and write methods here. Most models will probably use a custom format and hence overwrite the read and write methods. Thus writing to different table formats like excel and/or parquet will likely not be used. I suggest to limit it to a single format similar to the other properties to keep thinks simple on our side.

DirkEilander (approved):
I've made some final changes to
1. make the method between properties more consistent, i.e., order of arguments in set_tables; usage of single `set_tables` method rather than singular and plural; and fn argument in read and write. 
2. add tables to default read and write components all (sub)model classes.
3. added tables to append test

savente93:
@DirkEilander Thanks for fixing the tests :)

PR #503, opened 2023-09-06
Detect spatial and temporal extent of datasets

### LINKED ISSUE #496: Functionality to document and detect what slices in time and space a dataset covers
### Kind of request

Adding new functionality

### Enhancement Description

This is essentially a copy of the necessary feature from the STAC spec. There an extent is needed to show what areas of spaced and time a dataset covers. In addition to being necessary for STAC compatibility, it is also just a very nice feature to have. It should be possible to document these in the data catalog entry itself, but there should also be functionality for the data adapters to deduce this from the data itself and report it. If at all possible the STAC spec should be followed for easier compatibility when we start work on STAC support. 

### Use case

The best way to do this will have to be determined but a possible solution for the data catalog could be: 

```yaml 
entry_name:
  crs: 4326
  data_type: RasterDataset
  driver: netcdf
  extent:
    spatial: 
       bbox: [-10,-10,10,10] 
     temporal:
       range: ["2023-08-30","2023-09-10"]
  kwargs:
    chunks:
      latitude: 250
      longitude: 240
      time: 30
    combine: by_coords
    decode_times: true
    parallel: true
  path: path/to/file.nc
``` 
A few consdierations for the format above are: 
- the units of course must be in the correct crs 
- You could argue this belongs to the meta sub dict, but since it's required for STAC I'm not sure I want to make it optional here. Therefore I think it might have to be outside the `meta` key but this is debatable

```python
spatial_extent = data_catalog.get_geodataset("data_set_name").get_spatial_extent()
temporal_extent = data_catalog.get_geodataset("data_set_name").get_temporal_extent()
```

what the `spatial_extent` etc. objects look like can still to be determined and there is something to be said for supporting multiple formats as it could also be useful for other adapters, but we will have to see how much work this would be to implement.

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #496 

## Explanation
The spatial and temporal extent are simply the total bounding box in both space and time. The tests for making sure these are parsed correctly from the data catalog are present but commented out, because we decided to don't want to add it to the catalogs just yet, so this can't be tested for now. However, it's fairly straight forward so any problems that arrive from it should be easily solved in the relevant PR.
 
Now quite sure now much documentation on this needed to be done, if more is necessary please let me know. 

## Checklist

### DIFF
--- a/docs/user_guide/data_prepare_cat.rst
+++ b/docs/user_guide/data_prepare_cat.rst
@@ -36,6 +36,9 @@ The ``rename``, ``nodata``, ``unit_add`` and ``unit_mult`` options are set per v
       crs: EPSG/WKT
       data_type: RasterDataset/GeoDataset/GeoDataFrame/DataFrame
       driver: raster/raster_tindex/netcdf/zarr/vector/vector_table/csv/xlsx/xls
+      extent:
+        spatial: [xmin, ymin, xmax, ymax]
+        temporal: [start_time, end_time]
       driver_kwargs:
         key: value
       filesystem: local/gcs/s3
@@ -110,6 +113,8 @@ A full list of **optional data source arguments** is given below
 - **unit_mult**: multiply the input data by a value for unit conversion (e.g. 1000 for conversion from m to mm of precipitation).
 - **attrs** (optional): This argument allows for setting attributes like the unit or long name to variables.
   *NOTE*: New in HydroMT v0.7.2
+- **extent** (optional): The spatial and temporal extent the data covers. The spatial extent should always be in WGS84.
+  *NOTE*: New in HydroMT v0.8.1
 - **placeholder** (optional): this argument can be used to generate multiple sources with a single entry in the data catalog file. If different files follow a logical
   nomenclature, multiple data sources can be defined by iterating through all possible combinations of the placeholders. The placeholder names should be given in the
   source name and the path and its values listed under the placeholder argument.
--- a/docs/user_guide/data_prepare_cat.rst
+++ b/docs/user_guide/data_prepare_cat.rst
@@ -36,9 +36,6 @@ The ``rename``, ``nodata``, ``unit_add`` and ``unit_mult`` options are set per v
       crs: EPSG/WKT
       data_type: RasterDataset/GeoDataset/GeoDataFrame/DataFrame
       driver: raster/raster_tindex/netcdf/zarr/vector/vector_table/csv/xlsx/xls
-      extent:
-        spatial: [xmin, ymin, xmax, ymax]
-        temporal: [start_time, end_time]
       driver_kwargs:
         key: value
       filesystem: local/gcs/s3
@@ -113,8 +110,6 @@ A full list of **optional data source arguments** is given below
 - **unit_mult**: multiply the input data by a value for unit conversion (e.g. 1000 for conversion from m to mm of precipitation).
 - **attrs** (optional): This argument allows for setting attributes like the unit or long name to variables.
   *NOTE*: New in HydroMT v0.7.2
-- **extent** (optional): The spatial and temporal extent the data covers. The spatial extent should always be in WGS84.
-  *NOTE*: New in HydroMT v0.8.1
 - **placeholder** (optional): this argument can be used to generate multiple sources with a single entry in the data catalog file. If different files follow a logical
   nomenclature, multiple data sources can be defined by iterating through all possible combinations of the placeholders. The placeholder names should be given in the
   source name and the path and its values listed under the placeholder argument.
--- a/hydromt/data_adapter/dataframe.py
+++ b/hydromt/data_adapter/dataframe.py
@@ -83,6 +83,8 @@ class DataFrameAdapter(DataAdapter):
         attrs: dict, optional
             Additional attributes relating to data variables. For instance unit
             or long name of the variable.
+        extent: None
+            Not used in this adapter. Only here for compatability with other adapters.
         driver_kwargs, dict, optional
             Additional key-word arguments passed to the driver.
         name, catalog_name: str, optional
--- a/hydromt/data_adapter/dataframe.py
+++ b/hydromt/data_adapter/dataframe.py
@@ -139,9 +139,6 @@ class DataFrameAdapter(DataAdapter):
         time_tuple : tuple of str or datetime, optional
             Start and end date of the period of interest. By default, the entire time
             period of the DataFrame is included.
-        logger : Logger, optional
-            Logger object to log warnings or messages. By default, the module
-            logger is used.
         **kwargs : dict
             Additional keyword arguments to be passed to the file writing method.
 
@@ -158,13 +155,7 @@ class DataFrameAdapter(DataAdapter):
 
         """
         kwargs.pop("bbox", None)
-        try:
-            obj = self.get_data(
-                time_tuple=time_tuple, variables=variables, logger=logger
-            )
-        except IndexError as err:  # out of bounds for time
-            logger.warning(str(err))
-            return None, None, None
+        obj = self.get_data(time_tuple=time_tuple, variables=variables, logger=logger)
 
         read_kwargs = dict()
         if driver is None or driver == "csv":
@@ -196,6 +187,20 @@ class DataFrameAdapter(DataAdapter):
         based on the properties of this DataFrameAdapter. For a detailed
         description see: :py:func:`~hydromt.data_catalog.DataCatalog.get_dataframe`
         """
+        # load data
+        fns = self._resolve_paths(variables)
+        df = self._read_data(fns, logger=logger)
+        # rename variables and parse nodata
+        df = self._rename_vars(df)
+        df = self._set_nodata(df)
+        # slice data
+        df = DataFrameAdapter._slice_data(df, variables, time_tuple, logger=logger)
+        # uniformize data
+        df = self._apply_unit_conversion(df, logger=logger)
+        df = self._set_metadata(df)
+        return df
+
+    def _resolve_paths(self, variables=None):
         # Extract storage_options from kwargs to instantiate fsspec object correctly
         so_kwargs = {}
         if "storage_options" in self.driver_kwargs:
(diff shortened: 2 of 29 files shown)


hydromt/data_adapter/rasterdataset.py
```diff
@@ -622,3 +625,29 @@ def _parse_zoom_level(
         zl = zls[-1] if all(smaller) else zls[max(smaller.index(False) - 1, 0)]
         logger.info(f"Getting data for zoom_level {zl} based on res {zoom_level}")
         return zl
+
+    def detect_spatial_range(self, ds=None) -> geometry:
+        """Detect spatial range."""
+        if ds is None:
+            ds = self.get_data()
+        return box(*ds.raster.bounds)
```
savente93, line 633: note to self: always convert to WGS84 (EPSG 4326) (follow STAC) just return numbers

savente93: Update on note: STAC makes no reference to CRS, so presumably they assume the provided extents are always in the same CRS as the data, however, I think we should still have everything in WGS84 to make searching easier. We can always do a transform when converting to STAC

savente93: @DirkEilander Thoughts?

docs/user_guide/data_prepare_cat.rst
```diff
@@ -36,6 +36,9 @@ The ``rename``, ``nodata``, ``unit_add`` and ``unit_mult`` options are set per v
       crs: EPSG/WKT
       data_type: RasterDataset/GeoDataset/GeoDataFrame/DataFrame
       driver: raster/raster_tindex/netcdf/zarr/vector/vector_table/csv/xlsx/xls
+      extent:
+        spatial: [xmin, ymin, xmax, ymax]
+        temporal: [start_time, end_time]
```
DirkEilander, line 41: Two quick comments for discussion: 
- It would be good to discuss the naming of these properties and also whether these should be nested under extent. In other parts of HydorMT we use e.g. `bbox` for spatial extent. 
- Also many datasets wil have a single datetime instead of start and end. Do we enforce that the temporal extent has two values?

hydromt/data_adapter/rasterdataset.py
Jaapel: Most seem to be WGS84 indeed. Pretty surprising indeed. If this extension is supported then of course getting the CRS can be done by looking at collection / item metadata.

Jaapel: Example to load into xarray: https://github.com/opendatacube/odc-stac

docs/user_guide/data_prepare_cat.rst
savente93: I'm okay with using bbox instead if that's more consistent. 
The temporal ranges are inclusive, so as written the detection code will return a tuple with the same value, (or at least it should). I could go either way on this, enforcing is much nicer for the code since it eliminates a bunch of error prone null checks, while I can imagine allowing just one value is nicer for user.

savente93: Actually I decided to remove the stuff from the documentation since we're not adding it to the catalogs yes, however the discussion points are still good to resolve

hydromt/data_adapter/rasterdataset.py
savente93: What do we do/return if the CRS is not set?

Jaapel: You can:
- try to read some description and parse for CRS description
- try WGS84

Any other options come to mind?
Perhaps some of the files already contain the CRS information (headers in zarr / COG)

DirkEilander: CRS is required for any dataset in HydroMT. If it is not set in the catalog it should be set in the data source itself.

savente93: Then we should be okay. I implemented the separate functions Dirk suggested, so I think we should be okay now

tests/test_data_adapter.py
```diff
@@ -418,3 +417,24 @@ def test_cache_vrt(tmpdir, rioda_large):
     cat = DataCatalog(join(root, f"{name}.yml"), cache=True)
     cat.get_rasterdataset(name)
     assert len(glob.glob(join(cat._cache_dir, name, name, "*", "*", "*.tif"))) == 16
+
+
+def test_detect_extent(geodf, geoda, rioda, ts):
```
DirkEilander, line 422: The coverage of the the new methods is not great, see e.g. https://app.codecov.io/gh/Deltares/hydromt/blob/detect-extent/hydromt%2Fdata_adapter%2Fgeodataset.py

Perhaps by creating a small util functions to detect the temporal extend and reuse these between adapters this can be improved.

DirkEilander (changes requested):
Nice work @savente93 ! 
Main point to discuss is how we save/name these new properties internally in the adapters.
Also some more testing is required. Let me know if you have questions on my comments.

hydromt/data_adapter/geodataframe.py
```diff
@@ -389,3 +400,45 @@ def _set_metadata(self, gdf):
...
+    ) -> Tuple[Tuple[float, float, float, float], int]:
+        """Detect spatial range."""
+        if ds is None:
+            ds = self.get_data()
+
+        crs = pyproj.CRS.from_user_input(ds.geometry.crs).to_epsg()
+        bounds = ds.geometry.total_bounds
+        return bounds, crs
+
+    def detect_bbox_geographic(
```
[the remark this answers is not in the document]
savente93: Good point, hadn't thought of that, I'll give it a try

hydromt/data_adapter/dataframe.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         unit_add: dict = {},
         meta: dict = {},
         attrs: dict = {},
+        extent=None,
```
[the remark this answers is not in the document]
savente93: We do actually, because in catalog we call `adapter(...)` so if this keyword is not known it will either throw an error, or pass it down as a kwarg. So we need it here, but I agree that it should be made more clear that nothing is done with it here

hydromt/data_adapter/geodataframe.py
```diff
@@ -122,6 +132,7 @@ def __init__(
             version=version,
         )
         self.crs = crs
+        self.extent = extent
```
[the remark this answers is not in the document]
savente93: To be discussed at next stand up on Thursday

tests/test_data_adapter.py
savente93: I think coverage is much better now. Can you check that you're happy with it?

hydromt/data_adapter/dataframe.py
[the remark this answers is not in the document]
DirkEilander: As long as DataFrame sources do not define the extent in their catalog it will be passed to the DataFrameAdapter right? If think it's better to throw a meaningful error when parsing the data catalogs rather than masking it.

hydromt/data_adapter/geodataframe.py
```diff
@@ -389,3 +399,64 @@ def _set_metadata(self, gdf):
...
+        return bbox, crs
+
+    def detect_bbox(
+        self,
+        gdf=None,
+    ) -> Tuple[Tuple[float, float, float, float], int]:
+        """Detect the bounding box and crs of the dataset.
+
+        If no dataset is provided, it will be fetched accodring to the settings in the
+        addapter. also see :py:meth:`hydromt.GeoDataframeAdapter.get_data`. the
```
DirkEilander, line 437:
```suggestion
        If no dataset is provided, it will be fetched acodring to the settings in the
        adapter. also see :py:meth:`hydromt.GeoDataframeAdapter.get_data`. the
```

hydromt/data_adapter/geodataset.py
```diff
@@ -471,3 +481,116 @@ def _apply_unit_conversion(self, ds, logger=logger):
...
+        return temporal_extent
+
+    def detect_bbox(
+        self,
+        ds=None,
+    ) -> Tuple[Tuple[float, float, float, float], int]:
+        """Detect the bounding box and crs of the dataset.
+
+        If no dataset is provided, it will be fetched accodring to the settings in the
+        addapter. also see :py:meth:`hydromt.GeoDatasetAdapter.get_data`. the
```
DirkEilander, line 545:
```suggestion
        If no dataset is provided, it will be fetched according to the settings in the
        adapter. also see :py:meth:`hydromt.GeoDatasetAdapter.get_data`. the
```

hydromt/data_adapter/geodataset.py
```diff
@@ -471,3 +481,116 @@ def _apply_unit_conversion(self, ds, logger=logger):
...
+
+        crs = ds.vector.crs.to_epsg()
+        bounds = ds.vector.bounds
+        return bounds, crs
+
+    def detect_time_range(self, ds=None) -> Tuple[datetime, datetime]:
+        """Detect the temporal range of the dataset.
+
+        If no dataset is provided, it will be fetched accodring to the settings in the
+        addapter. also see :py:meth:`hydromt.GeoDatasetAdapter.get_data`.
```
DirkEilander, line 576:
```suggestion
        If no dataset is provided, it will be fetched according to the settings in the
        adapter. also see :py:meth:`hydromt.GeoDatasetAdapter.get_data`.
```

hydromt/data_adapter/rasterdataset.py
```diff
@@ -622,3 +633,116 @@ def _parse_zoom_level(
...
+        return time_range
+
+    def detect_bbox(
+        self,
+        ds=None,
+    ) -> Tuple[Tuple[float, float, float, float], int]:
+        """Detect the bounding box and crs of the dataset.
+
+        If no dataset is provided, it will be fetched accodring to the settings in the
+        addapter. also see :py:meth:`hydromt.RasterdatasetAdapter.get_data`. the
```
DirkEilander, line 697:
```suggestion
        If no dataset is provided, it will be fetched according to the settings in the
        adapter. also see :py:meth:`hydromt.RasterdatasetAdapter.get_data`. the
```

DirkEilander (changes requested):
Thanks for the update! It looks good, just a couple of small comments as final touch so that it's ready to merge after the discussion Thu morning.

hydromt/data_adapter/geodataset.py
```diff
@@ -471,3 +481,116 @@ def _apply_unit_conversion(self, ds, logger=logger):
...
+            whether to detect the time range if it is not set. If False, and it's not
+            set None will be returned.
+
+        Returns
+        -------
+        range: Tuple[np.datetime64, np.datetime64]
+            A tuple containing the start and end of the time dimension. Range is
+            inclusive on both sides.
+        """
+        temporal_extent = self.extent.get("time_tuple", None)
```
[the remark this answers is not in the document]
savente93: You're right, forgot to update that, my bad. good catch!

hydromt/data_adapter/dataframe.py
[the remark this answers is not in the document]
savente93: I did not consider that, you are right! Removed the keyword from Dataframe :)

hydromt/data_adapter/geodataframe.py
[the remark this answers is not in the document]
savente93: the format extent = {"bbox":[...],"time_range":[...]} was agreed on

DirkEilander (approved):
Thanks @savente93! It looks good to me

PR #505, opened 2023-09-06
fixes extreme value analysis for application on a grid

### LINKED ISSUE #469: EVA method not implemented to work with 2D dask array
### HydroMT version checks

### Reproducible Example

I tried using eva method as shown in the examples/doing_extreme_value_analysis.ipynb with a larger datasets and found memory leaks.

The file I test with is 83MB and I can provide it if needed.

### Current behaviour

There are several leaks so far:
 
**get_peaks**

in L268, `first` method does not seem to work with dask array. As we want the time label rather than the discharge values, I suggest to use another method that is dask compatible (mean works) or use maybe a pandas method on the time values rather than on the full dataarray.

https://github.com/Deltares/hydromt/blob/50e50b5c67c09f644e2a1a3ee9521030068675fe/hydromt/stats/extremes.py#L264-L270

Error:

> NotImplementedError: 'first' with skipna=True is not yet implemented on dask arrays

**fit_extremes**

L409 calls a da_params.values which triggers a load and in my case MemoryError.

https://github.com/Deltares/hydromt/blob/50e50b5c67c09f644e2a1a3ee9521030068675fe/hydromt/stats/extremes.py#L408-L419

Error:

> numpy.core._exceptions._ArrayMemoryError: Unable to allocate 27.5 MiB for an array with shape (4016, 31, 58) and data type float32

I see this is because if several distributions are used, it is not known which one will fit best. In case there is only one distribution, the following avoids the memory error:

```
    # split output
    if len(distributions) > 1:
        idist = da_params.isel(dparams=-1).values.astype(int)
        distributions = np.atleast_1d(np.array(distributions)[idist])
    else:
        # avoid memory error if only one distribution was used
        distributions = np.atleast_1d(np.full(da_params.isel(dparams=-1).shape, distributions[0]))
```

**get_lmom**

Last error I had (this one I did not solve) was on line 908 that `negative values are not allowed`. Not sure if it's because I use a 2D grid or something else. After that I switched to xclim.indices.stats.frequency_analysis instead of eva

https://github.com/Deltares/hydromt/blob/7a7883770099ce9b61595dd22e82d39308246eb5/hydromt/stats/extremes.py#L906-L909

### Desired behaviour

Avoiding memory issue as much as possible. Some suggestions are mentioned above.

### Additional context

_No response_

DirkEilander:
## Issue addressed
Fixes #469 

## Explanation

## Checklist


hydromt/stats/extremes.py
```diff
@@ -254,44 +255,45 @@ def get_peaks(
...
+    lmax = duck.apply_along_axis(func, da.get_axis_num(time_dim), da)
     # apply POT threshold
     peaks = da.where(lmax)
     if ev_type.upper() == "POT":
-        peaks = da.where(peaks > da.quantile(qthresh, dim="time"))
-    # TODO get extreme rate - this rate is peak rate per period!! not correct
-    # da_rate = np.isfinite(peaks).sum("time") / nperiods
-    da_rate = np.isfinite(peaks).sum("time") / duration
+        peaks = da.where(peaks > da.quantile(qthresh, dim=time_dim))
+    # TODO get extreme rate per year
```
savente93, line 287: seems like this TODO is fixed now?

savente93 (approved):
One small nitpick, but otherwise the code looks good to me! :)

hboisgon:
Hi @DirkEilander, nice changes thanks! Seems the code should be much more robust! I tried using the new eva functions with discharge outputs from wflow and I get an error. I am guessing maybe because there are some inactive cells in the model for which you have NaN for every timestep. Seems you do something about it in _return_values_1d but not in _fitopt_1d:

```
File ~\D\Repos\hydromt\hydromt\hydromt\stats\extremes.py:415, in fit_extremes.<locals>._fitopt_1d(x, distributions, criterium)
    414 def _fitopt_1d(x, distributions=distributions, criterium=criterium):
--> 415     params, d = lmoment_fitopt(x, distributions=distributions, criterium=criterium)
    416     if len(params) == 2:
    417         params = np.concatenate([[0], params])

File ~\D\Repos\hydromt\hydromt\hydromt\stats\extremes.py:768, in lmoment_fitopt(x, distributions, criterium)
    765 x = x[~np.isnan(x)]
    767 # derive first four L-moments from data
--> 768 lmom = get_lmom(x, 4)
    769 # derive parameters of distribution function
    770 params = {}

File ~\D\Repos\hydromt\hydromt\hydromt\stats\extremes.py:935, in get_lmom(x, nmom)
    933 for r in range(1, nmom):
    934     Num1 = np.kron(np.ones((r, 1)), np.arange(r + 1, n + 1))
--> 935     Num2 = np.kron(np.ones((n - r, 1)), np.arange(1, r + 1)).T
    936     Num = np.prod(Num1 - Num2, axis=0)
    938     Den = np.prod(np.kron(np.ones((1, r)), n) - np.arange(1, r + 1))

File ~\Miniconda3\envs\hydromt\lib\site-packages\numpy\core\numeric.py:204, in ones(shape, dtype, order, like)
    201 if like is not None:
    202     return _ones_with_like(shape, dtype=dtype, order=order, like=like)
--> 204 a = empty(shape, dtype, order)
    205 multiarray.copyto(a, 1, casting='unsafe')
    206 return a

ValueError: negative dimensions are not allowed
```

Is it something you could maybe still fix in that PR? I can share the wflow output I used to test with you.
Thanks!

DirkEilander:
@hboisgon thanks for your suggestion. I've tried to include it. Can you check if it works with your data. For grids with many NaNs would however suggest to first remove these before applying this method, that will be much faster I think. You can do so by first stacking the x and y dimension, then dropping NaNs along the new stacked dimension, then apply the statistics and finally (if needed) unstack back to the grid filling it with NaNs.

hboisgon:
Thanks @DirkEilander looks good now! Maybe your suggestion could go to an extra part in the eva notebook? :) else good to go!

hboisgon (approved):

hydromt/stats/extremes.py
DirkEilander: fixed. I updated the comment

DirkEilander:
> Thanks @DirkEilander looks good now! Maybe your suggestion could go to an extra part in the eva notebook? :) else good to go!

I think that's more on the level of a stack overflow question. I'm going to merge and close this PR, we can always add it later.

PR #506, opened 2023-09-11
Add data catalog version compatibility checking

### LINKED ISSUE #461: ENH: Solve which data catalog version to use based on the user hydromt version
### Kind of request

### Enhancement Description

If we would now update a predefined data catalog file based on recent developments in #438 and #334, users with a hydromt version previous to those developments would no longer be able to used that data catalog unless specifying the version with 
`DataCatalog(data_libs=[{lib}={version}])`

There are two possible solutions in my opinion:
- Selecting the last compatible data catalog version . This could happen in the `DataCatalog.from_predefined_catalogs` method. This could be done by adding the minimal hydromt version to the [predefined_catalogs.yml](https://github.com/Deltares/hydromt/blob/main/data/predefined_catalogs.yml) or loop over the data catalog files until we find a match.
- Include the data catalog files and predifined_catalog file in the release itself rather than downloading it from github. We would than need to make sure that the catalog files work with each release. However, this would limit our flexibility on how fast we can update the data catalog (add new sources etc.)

TODO decide on strategy of predefined catalogs (in release of seperate versioning on github). @hboisgon @savente93 what are your thoughts?

### Use case

see above

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #461 
## Explanation
As discussed in the linked issue, we added the `hydromt_version` key to the meta section on the catalogs. This should be a version specification set according to [PEP 440](https://peps.python.org/pep-0440/#version-specifiers). Due to the amount of different versions we also now have floating around in the code base, I've decided to rename all of them to avoid coalistion/confusion as happened to me. If I incorrectly renamed one please point it out. Otherwise I think this should all be good to go. It'd be really nice to have this before the next release!

## Checklist


docs/changelog.rst
```diff
@@ -21,6 +21,7 @@ Added
 - Model objects now have a _MODEL_VERSION attribute that plugins can use for compatibility purposes (PR # 495)
 - Model class now has methods for getting, setting, reading and writing arbitrary tabular data. (PR #502)
 - Relevant data adapters now have functionality for reporting and detecting the spatial and temporal extent they cover (# 503)
+- Data catalogs have a ``version`` and ``hydromt_version`` meta key that is used to determine compatibility between the catalog and the installed hydromt version. (PR #506)
```
DirkEilander, line 24: Only the hydromt_version is used to determine compatibility right?

```suggestion
- Data catalogs have a ``hydromt_version`` meta key that is used to determine compatibility between the catalog and the installed hydromt version. (PR #506)
```

DirkEilander (approved):
nice work @savente93! I've tested it an all works as expected. 

Just made one small change to only check compatibility if hydromt_version is set in the catalog. If you disagree we can revert this of course.

PR #512, opened 2023-09-15
Cache data catalog archives when an archive is set in the data catalog root.

### LINKED ISSUE #511: Allow the root of a data catalog to point to an archive
### Kind of request

Adding new functionality

### Enhancement Description

The yml file of the artifact data catalog is now part of the archive. Hence, the archive needs to be downloaded first before we can read its meta data including the required hydromt version. To find a compatible version this could require downloading several archives. Therefore it would be better to store the yml files of data catalog archives outside the archive with its root property pointing to the archive.

### Use case

The artifact_data catalog.

### Additional Context

_No response_

DirkEilander:
## Issue addressed
Fixes #511 

## Explanation
Cache data catalog archives when an archive is set in the data catalog root.
This should replace the current artifact_data entries in the predefined data catalog once released.

## Checklist


hydromt/data_catalog.py
```diff
@@ -699,8 +721,17 @@ def from_yml(
...
+        version = meta.get("version", None)
         if root is None:
             root = meta.get("root", os.path.dirname(urlpath))
+        if root.split(".")[-1] in ["gz", "zip"]:
+            # if root is an archive, unpack it at the cache dir
+            root = self._cache_archive(root, name=catalog_name, version=version)
+            # save catalog to cache
+            with open(join(root, f"{catalog_name}.yml"), "w") as f:
+                data_dict = {"meta": {k: v for k, v in meta.items() if k != "root"}}
+                data_dict.update(copy.deepcopy(yml))
```
savente93, line 733: why a deepcopy here? doesn't seem necessary to me

tests/test_data_catalog.py
```diff
@@ -294,6 +293,16 @@ def test_from_archive(tmpdir):
...
 
 
+def test_from_yml_with_archive(tmpdir):
+    yml_fn = join(CATALOGDIR, "artifact_data.yml")
+    data_catalog = DataCatalog(yml_fn)
+    sources = list(data_catalog.sources.keys())
+    assert len(sources) > 0
+    root = dirname(data_catalog[sources[0]].path)
+    yml_dst_fn = join(root, "artifact_data.yml")
+    assert isfile(yml_dst_fn)
```
savente93, line 303: it's not super clear what this test actually asserts to me, shouldn't we also assert that the catalog is actually cached to disk?

savente93 (changes requested):
Very nice Dirk, I was thinking about this but it seems you beat me to it. Two small comments to resolve, but after that should be ready for merging!

hydromt/data_catalog.py
DirkEilander: I agree

savente93 (approved):
Good to go now!

PR #513, opened 2023-09-15
Support for checking if DataCatalog "contains" a source 

### LINKED ISSUE #501: BUG? Check if a data source is in DataCatalog
### HydroMT version checks

### Reproducible Example

>>> import hydromt
>>> dc = hydromt.DataCatalog("artifact_data")
>>> "merit_hydro" in dc
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'DataCatalog' is not iterable

### Current behaviour

This returns an Error that DataCatalog is not iterable while this use to work (return True or False) before and is used at least in hydromt_wflow plugin a couple of times...

Anyway maybe it was not the best way to check if a data source is in data catalog and the option *'merit_hydro' in dc.sources* is better and works but as it is used, I thought good to investigate why it does not work anymore and inform plugins accordingly before releasing.

### Desired behaviour

The Error could stay but the change in behavior should be reported.

### Additional context

_No response_

DirkEilander:
## Issue addressed
Fixes #501 

## Explanation
Implementing a `__contains__` method allows for checking if data catalog contains a source with `name in catalog`

Also fixed the used_only argument of the to_yml method. This did no work properly as it only saved the source name and not the catalog or version.

## Checklist

### DIFF
--- a/hydromt/data_adapter/data_adapter.py
+++ b/hydromt/data_adapter/data_adapter.py
@@ -201,11 +201,17 @@ class DataAdapter(object, metaclass=ABCMeta):
         self.meta = {k: v for k, v in meta.items() if v is not None}
         # variable attributes
         self.attrs = {k: v for k, v in attrs.items() if v is not None}
+        # keep track of wether the data is used
+        self._used = False
 
     @property
     def data_type(self):
         """Return the datatype of the addapter."""
         return type(self).__name__.replace("Adapter", "")
+
+    def mark_as_used(self):
+        """Mark the data adapter as used."""
+        self._used = True
 
     def summary(self):
         """Return a dictionary summary of the data adapter."""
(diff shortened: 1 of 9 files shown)


savente93:
do note that if we use the `__contains__` method then this cannot be used to e.g. check that a specific version is available, only entry with that name. if this is important that I'd suggest implementing a function like `contains_source` with the same implementation but with more arguments

DirkEilander:
> do note that if we use the `__contains__` method then this cannot be used to e.g. check that a specific version is available, only entry with that name. if this is important that I'd suggest implementing a function like `contains_source` with the same implementation but with more arguments

Could suggestion @savente93! Would you have time to implement this?

savente93:
Yeah I can take a look I think, do you want me to take over this PR from you then?

savente93 (approved):
Good work @DirkEilander. It all looks good from my side. If you can take a look at my changes and check that you're happy with it then she's ready to merge!

DirkEilander:
> Good work @DirkEilander. It all looks good from my side. If you can take a look at my changes and check that you're happy with it then she's ready to merge!

Thanks Sam! Nice addition. I think this is ready to be merged

PR #514, opened 2023-09-18
Support for writing COG and reading zoom levels from internal COG overviews

### LINKED ISSUE #433: ENH: Enable accessing overviews of COG data with zoom_level argument in DataCatalog.get_rasterdataset
### Kind of request

Adding new functionality

### Enhancement Description

Currently zoomlevels can only be used when the data is saved in different files. However, Cloud Optimized GeoTiffs (COG) have internal so-called overviews.  It would be very useful to be able to access a specific overview using the `zoom_level` argument in `DataCatalog.get_rasterdataset` method. 

### Use case

This would reduce the amount of data required when e.g. building a 100m resolution model based on 1m data. 

### Additional Context

COG overviews can be read with rioxarray using the `overview_level` argument (see https://corteva.github.io/rioxarray/html/examples/COG.html):

```python
rioxarray.open_rasterio(cog_url, overview_level=4)
```

COGs with overviews are easy to create using GDAL tools or e.g. [rio-cogeo](https://cogeotiff.github.io/rio-cogeo/CLI/) and have multiple advantages such a fast visualisation in GIS.

DirkEilander:
## Issue addressed
Fixes #433 

## Explanation

## Checklist

Also added support for writing COG with overviews in `RasterDataset.to_raster`

### DIFF
--- a/hydromt/data_adapter/data_adapter.py
+++ b/hydromt/data_adapter/data_adapter.py
@@ -318,7 +318,7 @@ class DataAdapter(object, metaclass=ABCMeta):
         fmt = {}
         # update based on zoomlevel (size = 1)
         if "zoom_level" in keys:
-            zoom_level = 0 if zoom_level is None else zoom_level
+            zoom_level = zoom_level or 0
             fmt.update(zoom_level=zoom_level)
         # update based on dates and variables  (size >= 1)
         for date, var in product(dates, vrs):
--- a/hydromt/data_adapter/data_adapter.py
+++ b/hydromt/data_adapter/data_adapter.py
@@ -318,6 +318,7 @@ class DataAdapter(object, metaclass=ABCMeta):
         fmt = {}
         # update based on zoomlevel (size = 1)
         if "zoom_level" in keys:
+            zoom_level = 0 if zoom_level is None else zoom_level
             fmt.update(zoom_level=zoom_level)
         # update based on dates and variables  (size >= 1)
         for date, var in product(dates, vrs):
--- a/hydromt/data_adapter/rasterdataset.py
+++ b/hydromt/data_adapter/rasterdataset.py
@@ -7,7 +7,7 @@ import warnings
 from datetime import datetime
 from os import PathLike
 from os.path import join
-from typing import NewType, Optional, Tuple, Union, cast
+from typing import Dict, NewType, Optional, Tuple, Union, cast
 
 import geopandas as gpd
 import numpy as np
@@ -15,6 +15,7 @@ import pandas as pd
 import pyproj
 import rasterio
 import xarray as xr
+from rasterio.errors import RasterioIOError
 
 from .. import gis_utils, io
 from ..raster import GEO_MAP_COORD
@@ -270,7 +271,7 @@ class RasterDatasetAdapter(DataAdapter):
         """
         # load data
         fns = self._resolve_paths(time_tuple, variables, zoom_level, geom, bbox, logger)
-        ds = self._read_data(fns, geom, bbox, cache_root, logger)
+        ds = self._read_data(fns, geom, bbox, cache_root, zoom_level, logger)
         # rename variables and parse data and attrs
         ds = self._rename_vars(ds)
         ds = self._validate_spatial_dims(ds)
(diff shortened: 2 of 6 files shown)


hydromt/data_adapter/data_adapter.py
```diff
@@ -318,6 +318,7 @@ def _resolve_paths(
         fmt = {}
         # update based on zoomlevel (size = 1)
         if "zoom_level" in keys:
+            zoom_level = 0 if zoom_level is None else zoom_level
```
savente93, line 321: This can also be expressed as `zoom_level = zoom_level or 0` which I personally find a bit more readable, but up to you which you like better, I don't have very strong feelings about it.

hydromt/data_adapter/rasterdataset.py
```diff
@@ -307,8 +308,7 @@ def _resolve_paths(
             else:
                 os.environ["AWS_NO_SIGN_REQUEST"] = "NO"
 
-        # parse zoom level (raster only)
-        if len(self.zoom_levels) > 0:
+        if zoom_level is not None and "{zoom_level}" in self.path:
```
savente93, line 311: isn't this going to trigger if the path contains basically just the digit of the zoom_level anywhere in the path? Not quite sure if there is a way around it right now, but something to think about at least.

hydromt/data_adapter/rasterdataset.py
```diff
@@ -377,6 +377,13 @@ def _read_data(self, fns, geom, bbox, cache_root, logger=logger):
                 fns = fns_cached
             if np.issubdtype(type(self.nodata), np.number):
                 kwargs.update(nodata=self.nodata)
+            if zoom_level is not None and "{zoom_level}" not in self.path:
```
savente93, line 380: same comment as above

hydromt/data_adapter/rasterdataset.py
savente93: Oh also, I think you meant
```suggestion
        if zoom_level is not None and f"{zoom_level}" in self.path:
```
This checks for the literal string "{zoom_level}" and not the value

hydromt/data_adapter/rasterdataset.py
```diff
@@ -578,70 +585,114 @@ def _set_metadata(self, ds):
...
-                lat = 0
-                if bbox is not None:
-                    lat = (bbox[1] + bbox[3]) / 2
-                elif geom is not None:
-                    lat = geom.to_crs(4326).centroid.y.item()
+            dst_crs = pyproj.CRS.from_user_input(dst_crs)
+            dst_crs_unit = dst_crs.axis_info[0].unit_name
+            dst_res = src_res
+            if dst_crs_unit != src_res_unit:
+                known_units = ["degree", "metre", "US survey foot", "meter", "foot"]
```
savente93, line 661: would km also be a common one? or is that one not used often?

savente93 (approved):
Code looks good! few small comments, but those you can either take or leave depending on how you feel about them, feel free to merge whenever.

hydromt/data_adapter/rasterdataset.py
DirkEilander: It should actually check if a `"{zoom_level}"` is in the path, e.g., True in case of `"path/to/my/files/{zoom_level}/data.tif"` 
The value is then computed by `_parse_zoom_level` when a zoom_level is passed to `get_data`

hydromt/data_adapter/data_adapter.py
DirkEilander: if find my notation easier to read because `None` is not explicitly mentioned in your notation. But both are fine.

hydromt/data_adapter/rasterdataset.py
DirkEilander: Not in pyproj crs to my knowledge

hydromt/data_adapter/rasterdataset.py
savente93: aaaah right gotcha, my bad

hydromt/data_adapter/data_adapter.py
savente93: True, though I personally find the order of inline if statements in Python somewhat confusing. Again, personal preference, feel free to keep it the way you had it if you want :)

PR #517, opened 2023-09-20
Add ruff Bug-bear linting

savente93:
## Explanation
Flake8 bug bear is a linter that checks for various common bad practices in python that are common footguns. This PR adds it as a lint and fixes all it's problems for the current code base, fixing a couple of bugs along the way. I also adjusted the docs workflow a little but since that workflow spends a lot of it's time writing the results to disk, which is wasted effort if we're not going to publish them. The resulting behavior of both the workflow, and the python code after this small refactor is identical or better.

## Checklist

### DIFF
--- a/hydromt/data_adapter/dataframe.py
+++ b/hydromt/data_adapter/dataframe.py
@@ -30,12 +30,12 @@ class DataFrameAdapter(DataAdapter):
         driver: Optional[str] = None,
         filesystem: str = "local",
         nodata: Optional[Union[dict, float, int]] = None,
-        rename: dict = {},
-        unit_mult: dict = {},
-        unit_add: dict = {},
-        meta: dict = {},
-        attrs: dict = {},
-        driver_kwargs: dict = {},
+        rename: dict = None,
+        unit_mult: dict = None,
+        unit_add: dict = None,
+        meta: dict = None,
+        attrs: dict = None,
+        driver_kwargs: dict = None,
         name: str = "",  # optional for now
         catalog_name: str = "",  # optional for now
         provider: Optional[str] = None,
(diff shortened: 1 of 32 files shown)


hydromt/data_adapter/dataframe.py
```diff
@@ -89,12 +89,25 @@ def __init__(
...
+        if attrs is None:
+            attrs = {}
+        if meta is None:
+            meta = {}
+        if unit_add is None:
+            unit_add = {}
+        if unit_mult is None:
+            unit_mult = {}
+        if rename is None:
+            rename = {}
```
DirkEilander, line 103: Not sure this is "good practice", but can we not shorten this by setting eg`driver_kwargs = driver_kwargs or {}` when we pass the argument to the method (or in other places set a class attribute). Declaring all these variables first makes the script quite long. This happens at many places.

DirkEilander (approved):
I'm generally fine with this extra check. However, the code could be shortened a lot in many places where we used to set lists or dicts as default parameters. I've done a suggestion in a separate comment to shorten this. 

Two additional questions/suggestions:
- Have you also tested this works on windows?
- Perhaps also good to add the dev docs that it's good practice to use None instead of empty dict or lists as default parameters.

hydromt/data_adapter/dataframe.py
savente93: correct. The two are equivalent, so the shorter version is equaly good practicise. I thought I'd use the short version everywhere, but apparently a few places slipped through the cracks. I'll update it!

savente93: I've used the shorter version wherever I could find it. if you find other places where it could be applied, feel free to let me know

savente93:
You make good points, I'll update the dev docs. Though I'm not sure what you mean I need to have tested on windows? If you mean the linter, that is already included in ruff which we already use so, and the tests pass which should be cross platform, so I'll be extremely surprised if this breaks anyting, but I'll give it a try just to be sure

PR #523, opened 2023-09-21
Readded contains

### LINKED ISSUE #522: Reïncorperate __contains__
### HydroMT version checks

### Reproducible Example/ Information

When testing the Wflow plugin, it failed, while not making any changes since the previous time I ran it (it passed).
@DirkEilander noted that this functionality:
```Python
def __contains__(self, key: str) -> bool:
    """Check if source is in catalog."""
    return key in self._sources
```
was removed but should still be there as the plugins depend on it.

### Current behaviour

Not incorperated

### Desired behaviour

Incorperated 

### Additional context

_No response_

dalmijn:
## Issue addressed
Fixes #522 

## Explanation
It broke the Wflow plugin and probably others. Just readded some code

## Checklist

### DIFF
--- a/hydromt/data_catalog.py
+++ b/hydromt/data_catalog.py
@@ -417,6 +417,10 @@ class DataCatalog(object):
         """Iterate over sources."""
         return iter(self.iter_sources())
 
+    def __contains__(self, key: str) -> bool:
+        """Check if source is in catalog."""
+        return key in self._sources
+
     def contains_source(
         self,
         source: str,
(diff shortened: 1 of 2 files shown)


DirkEilander (approved):
Thanks @dalmijn, looks good to me!

@savente93 I had missed earlier that the `__contains__` was removed in favor of `contains_source`. In my opinion it would be good to have both. At least for now to not break to many things in plugins downstream. We can discuss how the v1 implementation should look like and add a DeprecationWarning in a next version if we decide to not use this implementation.

PR #531, opened 2023-09-27
Vector model updates

### LINKED ISSUE #474: Rename LumpedModel.response_units to VectorModel.vector
### Kind of request

### Enhancement Description

Rename LumpedModel.response_units to VectorModel.vector
This is more consistent with GridModel and Meshmodel.

### Use case

_No response_

### Additional Context

_No response_

### LINKED ISSUE #497: Create VectorModel basic functions
### Kind of request

None

### Enhancement Description

We want to update the VectorModel class with a `vector` attribute (instead of response_unit)
The set_* read_* and write_* for the `VectorModel.vector` property need to be updated to work with the latest GeoDataset methods.

The `vector` method should have a optional geom argument (required at initialization) and optional data (required when setting/updating a variable) argument.

By default we write to nc file (without the geometry) and a geojson file with the geometry. 

Links to #474 

### Use case

and future semi-distributed model plugin

### Additional Context

For the UWBM this is required to be implemented in the next release.

hboisgon:
## Issue addressed
Fixes #474 
Fixes #497 

## Explanation
See related issues

## Checklist

Because we are going to maybe write some generic methods, I changed that VectorModel geometry should be of type Polygon and not Point or Line. If this is not okay we can change but this will make the development of generic methods more complex.

### DIFF
--- a/hydromt/models/model_vector.py
+++ b/hydromt/models/model_vector.py
@@ -92,7 +92,7 @@ class VectorMixin:
         # Add to vector
         # 1. self.vector does not have a geometry yet or overwrite_geom
         # then use data directly
-        if self.geometry is None or overwrite_geom:
+        if self.vector_geometry is None or overwrite_geom:
             if data.vector.geometry is None:
                 raise ValueError("Cannot instantiate vector without geometry in data")
             else:
@@ -104,7 +104,9 @@ class VectorMixin:
             # data has a geometry - check if it is the same as self.vector
             if data.vector.geometry is not None:
                 if not np.all(
-                    data.vector.geometry.geom_almost_equals(self.geometry, decimal=4)
+                    data.vector.geometry.geom_almost_equals(
+                        self.vector_geometry, decimal=4
+                    )
                 ):
                     raise ValueError("Geometry of data and vector do not match")
             # add data (with check on index)
@@ -288,7 +290,7 @@ class VectorMixin:
 
     # Other vector properties
     @property
-    def geometry(self) -> gpd.GeoSeries:
+    def vector_geometry(self) -> gpd.GeoSeries:
         """Returns the geometry of the model vector as gpd.GeoSeries."""
         # check if vector is empty
         if len(self.vector.sizes) == 0:
@@ -381,7 +383,7 @@ class VectorModel(VectorMixin, Model):
         if "region" in self.geoms:
             region = self.geoms["region"]
         elif len(self.vector) > 0:
-            gdf = self.geometry.to_frame("geometry")
+            gdf = self.vector_geometry.to_frame("geometry")
             region = gpd.GeoDataFrame(geometry=[box(*gdf.total_bounds)], crs=gdf.crs)
         return region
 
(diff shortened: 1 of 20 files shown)


hboisgon:
Still need to improve the tests to test better the different options of updated set/read/write methods

hboisgon:
After discussion with @DirkEilander we decided to still support all kinds of geometries in VectorModel.vector for now. When developing generic methods we will see if good documentation and good checks on geometry types at the beginning of the setup methods will work out.

hydromt/vector.py
```diff
@@ -575,15 +581,17 @@ def to_gdf(self, reducer=None):
...
             method by which multidimensional data is reduced to 1 dimensional
-            e.g. numpy.mean
+            e.g. numpy.mean. If str (e.g. mean), will call the numpy method is found.
 
         Returns
         -------
         gdf: geopandas.GeoDataFrame
             GeoDataFrame
         """
+        if reducer is not None and isinstance(reducer, str):
```
savente93, line 593: `reducer is not None` is not needed here as `None` is not an instance of `str`, but it's a small nitpick. Up do you if you want to change it.

savente93 (approved):
One very tiny nitpick, but other than that all looks good. I like the tests, they seem to cover the relevant bits, so if they pass then I trust everything is working as intended :)

savente93:
Fixes #536

hydromt/models/model_vector.py
```diff
@@ -0,0 +1,433 @@
...
+            # write geometry
+            gdf = ds.vector.geometry.to_frame("geometry")
+            gdf.to_file(join(self.root, fn_geom))
+            # write_nc requires dict - use dummy key
+            nc_dict = {"vector": ds.drop_vars("geometry")}
+            self.write_nc(nc_dict, fn, **kwargs)
+
+    # Other vector properties
+    @property
+    def geometry(self) -> gpd.GeoSeries:
```
DirkEilander, line 291: Should we rename this to `vector_geometry` ?

hydromt/models/model_vector.py
```diff
@@ -0,0 +1,433 @@
...
+        ds = self.vector
+
+        # If fn is None check that vector contains only 1D data
+        if fn is None:
+            # If the user did specify a reducer, data can be more than 1D data
+            if "reducer" in kwargs and kwargs["reducer"] is not None:
+                self.logger.warning(
+                    "If 2D/3D data found,"
+                    f"they will be reduced to 1D using {kwargs['reducer']}"
+                )
```
DirkEilander, line 243: Why is this functionality "hidden" in the kwargs? I would suggest to either make the reducer explicit or (preferred) remove it as I don't think you commonly want to reduce model input data at this stage.

DirkEilander (commented):
Nice work @hboisgon!

In general this is good to go. I left two small comments to discuss about a the `geometry` property name and the reducer functionality in `write_vector`. Hope you find some time in between your courses this week to reply so we can wrap this up.

PR #534, opened 2023-09-29
Small Model updates

### LINKED ISSUE #475: Rename the generic setup method to distinguish better between the setup_schematisation and add data methods
### Kind of request

### Enhancement Description

So far it's not too clear if the generic setup methods create a grid/mesh schematisation or add data to it.
We kind of have a convention than method name starts with setup_ so renaming would be for example setup_grid_data_from_geodataframe instead of setup_grid_from_geodataframe.

### Use case

_No response_

### Additional Context

_No response_

### LINKED ISSUE #498: Relax forcing to support DataFrame objects
### Kind of request

Adding new functionality

### Enhancement Description

For lumped models (single geo unit) pandas DataFrames can desrcibe the forcing and xarray DataArray/Dataset objects make this too complex. Therefore we would like to relax forcing to support pandas DataFrame objects

This requires updating the `set_forcing` to allow pandas DataFrames with variables in the columns (explain in docstring). This can be done with a type check at the start of the method and converting it to a dataset using the [to_xarray method](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_xarray.html). 

### Use case

urban water balance model

### Additional Context

For discussion

### LINKED ISSUE #524: Overwritting existing models
### HydroMT version checks

### Reproducible Example

hydromt update wflow wflow_piave -o wflow_piave_update -i update.ini -d artifact_data -vv

hydromt update wflow wflow_piave -o wflow_piave_update -i update.ini -d artifact_data --fo -vv

> ERROR: Model dir already exists and cannot be overwritten

### Current behaviour

When updating a model in an already existing folder, the models cannot be overwritten even with the --force-overwrite option on.

### Desired behaviour

Like we implemented for build, I think if the user specifies --fo / --force-overwrite in update, hydromt should be able to overwrite the existing model.

### Additional context
Doing a check below if force_overwrite then set root with mode="w+" and else keep mode="w"

https://github.com/Deltares/hydromt/blob/7438afdf5b5712878833ffb40263ab73d47e51a3/hydromt/models/model_api.py#L292-L299

hboisgon:
## Issue addressed
Fixes #524 
Fixes #498 

## Explanation
See related issues

## Checklist

## Additional notes
For #475 I only renamed the setup_ methods and not the workflow methods. Not sure if I should rename them as well?


DirkEilander (changes requested):
Most changes look good. However for the renaming we need to have a bit more input from users.  What are clear naming conventions and do we apply it to all components incl geoms and maps that don't have an initial setup method? I suggest to revert the renaming and keep that for a separate PR after organizing a small user panel feedback session.

DirkEilander (approved):
This PR is ready to merge! The renaming has been taken out and parked for discussion in #475

PR #538, opened 2023-10-02
Make better use of caching in CI

savente93:
## Issue addressed
NA

## Explanation
 As with all purely CI PRs there sadly arent' any tests that I can show to demonstrate it works, but I've worked on an dtested it in a seperate repository that the reviewer can look at should they want: https://github.com/savente93/ci-playground. Because the dependency caches tend to be quite large, there's not enough space in the cache allowance to let each PR have it's own cache so I went with the following setup: 
1. Every week on sunday all caches get cleared (to make sure they don't drift). There is also a manual dispatch for this one
2. then caches for each of the python version tests suites as well as the docs generation is generated on main. this is a read only cache that doesn't get updated until it's regenerated
3. If the resulting environment.yml is unchanged the tests can run without doing any updating
4. if it is changed, the main cache is still loaded and and update is preformed to get the new dependencies. This sadly has to happen every time the CI is run, but still improves on the current situation.
5. Since I've managed to cache mamba itself too, the rest of the flow can't continue without a cache hit, so I've added a step to just exit the flow if no cache is restred with a bit of a nicer error message.
6. The byte codes generated are also included in the cache, this might help with the issues experienced in #535 

## Checklist


.github/workflows/docs.yml
```diff
@@ -22,69 +21,68 @@ on:
...
+          # the below two settings mean we'll alway srestore the cache
+          # but the cache hit output can tell us if we have to update afterwards
+          key: docs-${{ hashFiles('environment.yml')  }}
+          restore-keys: |
+            docs
 
-      - name: Update environment & write to cache
-        run: mamba env update -n hydromt -f environment.yml
-        if: steps.cache.outputs.cache-hit != 'true'
+      - name: Fail on no cache restore
```
Jaapel, line 53: Does this mean the workflow fails if there is no cache available? What would be the expected cases for when no caching is available?

.github/workflows/refresh-cache.yml
```diff
@@ -0,0 +1,95 @@
+
```
Jaapel, line 1: Why do we refresh our cache? To obtain new versions of packages? Because there are seperate tools for that one I think

.github/workflows/docs.yml
savente93: it does, this should actually never happen, but because mamba itself is in the cache the flow would fail anyway, this is just to provide a bit of a nicer error message. In the case the cache is unavailable, it should be regenerated (which happens on a schedule)

.github/workflows/refresh-cache.yml
savente93: Various reasons. Obtained new versions of packages, integrate new packages into the cache, make sure all the tests are running from the same cache. Because mamba solving is not stable, it's important to keep all the tests in sync by periodically refreshing the cache that they all use. Besides, the cache limit is at 10GB which we bump up against about weekly, instead of trying to do very clever cache invalidation I thought it was much more maintainable to just clean out and regenerate the cache every week, that way we make sure we keep within our limits and that everything uses the up to date versions of everything.

Jaapel: It seems that dependabot does not do mamba / conda, so I think this is way better than nothing :)

Jaapel (approved):

PR #569, opened 2023-10-10
Raster cache props

DirkEilander:
## Issue addressed
I profiled the `to_slippy_tiles` methods to better understand its poor performance. This was mostly the result of many different calls to the `pyproject.CRS.from_user_input method` when `self.crs` was called in addition to other raster properties. And to some extend because of the gdal based reproject instead of simple numpy based averaging for coarsening of the raster.

## solutions implemented
- Improves performance of raster accessor by caching raster properties `crs`, `transform`, `res`, `origin`, and `rotation`
- in `to_slippy_tiles`  support different dtypes and use numpy based average upscaling
- Clean up of some remainder PCR Raster methods

## Checklist
- [x] Updated changelog.rst if needed (assumed not needed)

### DIFF
--- a/hydromt/gis_utils.py
+++ b/hydromt/gis_utils.py
@@ -54,7 +54,6 @@ YATTRS = {
         "units": "m",
     },
 }
-PCR_VS_MAP = {"ldd": "ldd"}
 GDAL_DRIVER_CODE_MAP = {
     "asc": "AAIGrid",
     "blx": "BLX",
--- a/hydromt/gis_utils.py
+++ b/hydromt/gis_utils.py
@@ -365,7 +365,7 @@ def affine_to_coords(transform, shape, x_dim="x", y_dim="y"):
     if not isinstance(transform, Affine):
         transform = Affine(*transform)
     height, width = shape
-    if transform.b == 0:
+    if transform.b == transform.d == 0:
         x_coords, _ = transform * (np.arange(width) + 0.5, np.zeros(width) + 0.5)
         _, y_coords = transform * (np.zeros(height) + 0.5, np.arange(height) + 0.5)
         coords = {
--- a/hydromt/gis_utils.py
+++ b/hydromt/gis_utils.py
@@ -364,7 +364,7 @@ def affine_to_coords(transform, shape, x_dim="x", y_dim="y"):
     if not isinstance(transform, Affine):
         transform = Affine(*transform)
     height, width = shape
-    if transform.b == transform.d == 0:
+    if np.isclose(transform.b, 0) and np.isclose(transform.d, 0):
         x_coords, _ = transform * (np.arange(width) + 0.5, np.zeros(width) + 0.5)
         _, y_coords = transform * (np.zeros(height) + 0.5, np.arange(height) + 0.5)
         coords = {
--- a/hydromt/raster.py
+++ b/hydromt/raster.py
@@ -623,17 +623,17 @@ class XRasterBase(XGeoBase):
     @property
     def origin(self) -> tuple[float, float]:
         """Return origin of grid (x0, y0) tuple."""
-        if self._origin is not None:
+        if self._origin is not None:  # cached origin
             return self._origin
-        elif self._transform is not None:
+        elif self._transform is not None:  # cached transform
             x0, y0 = self._transform * (0, 0)
-        else:
+        else:  # from coordinates
             x0, y0 = 0, 0
             xs, ys = self.xcoords.data, self.ycoords.data
             dx, dy = self.res
-            if xs.ndim == 1:
+            if xs.ndim == 1:  # regular non-rotated
                 x0, y0 = xs[0] - dx / 2, ys[0] - dy / 2
-            elif xs.ndim == 2:
+            elif xs.ndim == 2:  # rotated
                 alpha = math.radians(self.rotation)
                 beta = math.atan(dx / dy)
                 c = math.hypot(dx, dy) / 2.0
--- a/hydromt/raster.py
+++ b/hydromt/raster.py
@@ -1,8 +1,9 @@
 #!/usr/bin/env python
 # -*- coding: utf-8 -*-
-# Credits: This script is based on the rioxarray package (Apache License, Version 2.0)
-# source file: https://github.com/corteva/rioxarray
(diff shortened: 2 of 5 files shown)


hydromt/raster.py
```diff
@@ -262,53 +267,66 @@ def crs(self) -> CRS:
...
+                            crs = self._obj[var].attrs.pop(name)
+                            break
+            if crs is not None:
+                # avoid Warning 1: +init=epsg:XXXX syntax is deprecated
+                if isinstance(crs, str):
+                    crs = crs.removeprefix("+init=")
+                try:
+                    input_crs = pyproj.CRS.from_user_input(crs)
+                except RuntimeError:
+                    pass
```
savente93, line 320: Maybe we should also raise an error here? otherwise might be good to add a comment that this is intentional

hydromt/raster.py
```diff
@@ -520,21 +545,34 @@ def res(self) -> tuple[float, float]:
...
+                dy = ys[1] - ys[0]
+            elif xs.ndim == 2:
+                ddx0 = xs[1, 0] - xs[0, 0]
+                ddy0 = ys[1, 0] - ys[0, 0]
+                ddx1 = xs[0, 1] - xs[0, 0]
+                ddy1 = ys[0, 1] - ys[0, 0]
+                dx = math.hypot(ddx1, ddy1)  # always positive!
+                dy = math.hypot(ddx0, ddy0)
+        rot = self.rotation
+        if rot != 0:
```
savente93, line 573: if `rot` is a float and not an integer this is very error prone. In that case please use `np.isclose`

hydromt/gis_utils.py
```diff
@@ -365,7 +364,7 @@ def affine_to_coords(transform, shape, x_dim="x", y_dim="y"):
     if not isinstance(transform, Affine):
         transform = Affine(*transform)
     height, width = shape
-    if transform.b == 0:
+    if transform.b == transform.d == 0:
```
savente93, line 367: if `transform.b` or `transform.d` are floats and not an integers this is very error prone. In that case please use `np.isclose`

hydromt/raster.py
```diff
@@ -553,37 +592,56 @@ def rotation(self) -> float:
...
-            if ddx1 < 0:
-                rot = 180 + rot
-            elif ddy1 < 0:
-                rot = 360 + rot
+        if self._rotation is not None:  # cached rotation
+            return self._rotation
+        rot = None
+        if self._transform is not None:  # cached transform
+            # NOTE: it is not always possible to get the rotation from the transform
+            if self._transform.b == self._transform.d == 0:
```
savente93, line 600: if transform.b or transform.d are floats and not an integers this is very error prone. In that case please use np.isclose

hydromt/raster.py
```diff
@@ -2962,33 +3050,22 @@ def to_mapstack(
...
-                    tags=tags,
-                    windowed=windowed,
-                    mask=mask,
-                    logger=logger,
-                    **profile_kwargs,
-                )
+        for var in self.vars:
+            if "/" in var:
+                # variables with in subfolders
+                folders = "/".join(var.split("/")[:-1])
```
savente93, line 3056: If you're doing path operations here this is probably going to break on other oses that have different path seperators. Please use the `os.path` functions for these operations

hydromt/raster.py
```diff
@@ -2962,33 +3050,22 @@ def to_mapstack(
...
+                os.makedirs(join(root, folders), exist_ok=True)
+                var0 = var.split("/")[-1]
```
savente93, line 3058: If you're doing path operations here this is probably going to break on other oses that have different path seperators. Please use the os.path functions for these operations

savente93 (changes requested):
Looks good Dirk, performance improvements are always very nice. The code looks good except for a few minor changes that I think are good to adress, after that it should be good to merge. Could you also give a short description of the benchmark setups you used? a description in this PR of the data used and the actions taken should be fine, it will be very useful for me down the line when I want to setup more automatic benchmarking..

hydromt/raster.py
DirkEilander: this applies to the variable name rather then the path.
I will update line 3056 though.

savente93 (approved):
Thanks for the updates, all good now!

dalmijn (approved):

hydromt/raster.py
```diff
@@ -262,53 +267,66 @@ def crs(self) -> CRS:
...
+                # check default > GEO_MAP_COORDS attrs, then global attrs
+                if name in self.attrs:
+                    crs = self.attrs.get(name)
+                    break
+                if name in self._obj.attrs:
+                    crs = self._obj.attrs.pop(name)
+                    break
+            if crs is None:  # check data var and coords attrs
+                for var in names:
+                    for name in crs_names:
```
dalmijn, line 309: You could use `product` from itertools here..
That's just my taste. This is also fine

dalmijn, line 309: You could use `product` from itertools here..
That's just my taste. This is also fine

dalmijn:
The rest of it seems good to me, @DirkEilander. Thank you for these improvements. The only thing I am left wondering about is maybe add a general update method for updating crs and transform add the end of certain methods. (If maybe the future more stuff like that gets added). This can, in my opinion, increase the ease of maintainability. Was also kind of wondering about a decorater for that case, but that could be to much effort for now.

hydromt/raster.py
DirkEilander: done. thanks for the suggestion

DirkEilander:
> The rest of it seems good to me, @DirkEilander. Thank you for these improvements. The only thing I am left wondering about is maybe add a general update method for updating crs and transform add the end of certain methods. (If maybe the future more stuff like that gets added). This can, in my opinion, increase the ease of maintainability. Was also kind of wondering about a decorater for that case, but that could be to much effort for now.

Thanks for the review Brendan. This could indeed be improved further. However, it is not trivial how to update the cached properties for all methods as method may modify the properties in different ways. Set methods would be a nice addition indeed. However, this need a bit more thinking. Should set_transform also modify the coordinates, check with the coordinates, or only set the cached transform property potentially introducing a mis match?

PR #576, opened 2023-10-11
Update example installation instructions in docs

### LINKED ISSUE #465: Conflict HydroMT version vs examples version
### HydroMT version checks

### Kind of issue

Docs are wrong

https://deltares.github.io/hydromt/latest/getting_started/examples_index.html

### Documentation problem

In the docs, we asked the user to usually install hydromt with conda-forge.
But then to run the examples, we advise to clone the current dev version of the repository in order to download the jupyter notebooks.

Depending on changes (especially before releasing), the content in the examples might not work with the latest conda-release so examples would fail.
This has been reported for hydromt-wflow: https://jmp75.github.io/work-blog/

At least in the docs mention that users should download the zip folder from the hydromt release version they are using or warn that if they use main/clone they may run into problems.

Not sure if we could publish our examples more smartly, with maybe a function that would check the current installed hydromt version and downloads the right examples from there?

### LINKED ISSUE #546: Installation guide for examples is not up-to-date
### HydroMT version checks

### Kind of issue

Docs are wrong

https://deltares.github.io/hydromt/latest/getting_started/examples_index.html

### Documentation problem

With make.py, the environment.yml file does not exist anymore so the recommended installation does not work anymore.

Related issue:
https://github.com/Deltares/hydromt_wflow/issues/206

Either update by explaining how to install and use make and add the tomli dependency.

Or do create maybe in the examples folder the correct environment.yml file (maybe can be automatically created in one of our github workflow?).
I still think that for most hydromt users who works with command line and are less handy with python, we should not complicate things too much, especially when it comes to installing and trying out the tool for the first time. So I would be in favor of the second option

savente93:
## Issue addressed
Fixes #546 
Fixes #465 

## Explanation
I've decided to solve both of the issues above by simply explaining in the examples that either you need a dev installation for the latest examples, or you should check out the examples of the version you have installed. I have done this for a couple of reasons:
1. if you want to run the examples you already need the repository, at which point, making a dev install is a small step beyond that
2. given that you already have the repo, we can assume the user has at least some very rudimentary git knowledge, so checking out the tag of the version installed which has the versions of the examples they want is a simple operations
3. We also have binder for running the examples. 
4. it is by far the simplest option on our side. 

## Checklist

One other thing we could do is make the documentation link we list point to the latest stable release instead of the latest/dev version. That way people will find the examples that should be compatible with the versions they install from conda/pip. This would require a slight adjustment in how the docs are generated so I'll leave that for another issue/PR to figure out if that is something we want, but I did want to bring up the idea at least.

### DIFF
--- a/docs/getting_started/installation.rst
+++ b/docs/getting_started/installation.rst
@@ -77,6 +77,11 @@ Install HydroMT in a new environment
 
     This is our recommended way of installing HydroMT!
 
+.. Note::
+
+    If you intend to run the most up to date examples, you should follow the instalation guide at `the dev guide <https://deltares.github.io/hydromt/latest/dev/dev_install.html>`_ instead, as the dev version of hydromt is necessary to run the current examples. WIth the version installed through the steps
+    in this page, you will be able to run the examples at the time of the last release.
+
 To install HydroMT in a new environment called `hydromt` from the conda-forge channel do:
 
 .. code-block:: console
--- a/examples/README.rst
+++ b/examples/README.rst
@@ -12,11 +12,12 @@ Local installation
 
 To run these examples on your local machine you need a copy of the repository and
 an installation of HydroMT including some additional packages. The most reliable
-way to do this installation is by creating a new environment as described below.
+way to do this installation is by using the binder imlementation provided above, however
+if you wish to exectute them locally you should create a new environment as described below.
 
 1 - Install python and conda/mamba
 **********************************
-If not already done, you'll need Python 3.8 or greater and a package manager such as conda or mamba. These package managers help you to install (Python) packages and
+If not already done, you'll need Python 3.9 or greater and a package manager such as conda or mamba. These package managers help you to install (Python) packages and
 `manage environments <https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html>`_ such that different installations do not conflict.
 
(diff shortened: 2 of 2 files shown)


DirkEilander:
> One other thing we could do is make the documentation link we list point to the latest stable release instead of the latest/dev version. That way people will find the examples that should be compatible with the versions they install from conda/pip. This would require a slight adjustment in how the docs are generated so I'll leave that for another issue/PR to figure out if that is something we want, but I did want to bring up the idea at least

This is not bad idea! I think it can be done by changing the link in [index.html](https://github.com/Deltares/hydromt/blob/gh-pages/index.html) in the `gh-pages` branch to point to "stable" rather than "latest".

DirkEilander:
I'm happy for these changes to be merged. Whether it solves #465 satisfactory I'll leave up to @hboisgon We can just see what feedback we get after these changes (and changing the docs landing page) before we put more work into this following the suggestion in that issue.

DirkEilander (approved):
I'm happy for these changes to be merged. Whether it solves https://github.com/Deltares/hydromt/issues/465 satisfactory I'll leave up to @hboisgon We can just see what feedback we get after these changes (and changing the docs landing page) before we put more work into this following the suggestion in that issue.

hboisgon:
I still think the new explanations are not fully clear. Most of our users are command line users I think so as examples is one of the first thing they look at, everything should be clear step by step copy pasting and not going from one page to the other. There was also some spelling mistakes. I'll open a new branch to correct it.

PR #582, opened 2023-10-12
remove explicit gdal dep

DirkEilander:
## Issue addressed
remove gdal dependency 
links to https://github.com/Deltares/hydromt_delft3dfm/issues/13

## Explanation
gdal is not directly imported, so not a direct dependency of hydromt and could potentially make hydromt pip installable.

## Checklist

### DIFF
--- a/.binder/Dockerfile
+++ b/.binder/Dockerfile
@@ -1,7 +1,7 @@
 # currently we haven't published an official hydromt image yet, so for now
 # I import from an export I've put on my own account. this should be moved
 # to the official image as soon as it's published.
-FROM savente/hydromt:slim as binder
+FROM deltares/hydromt:slim as binder
 # Binder hard requires all of these steps when they build the imae
 # therefore these steps aren't taken sooner
 
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -34,7 +34,7 @@ jobs:
       - name: generate envs
         run: |
           pip install tomli
-          make full-environment.yml
+          python make_env.py -p 3.11 -o full-environment.yml full
 
       - name: Build and test
         uses: docker/build-push-action@v4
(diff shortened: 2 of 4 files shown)


DirkEilander:
@savente93 is this tested properly in the CI because of the cached environments

savente93:
I'm not sure, mamba env update may not delete already installed issues, so it's quite possible that the tests have access to a gdal even though they're not supposed to. There's not really a way to do this with the current cache setup, you'll have to make a new environment without the gdal to properly test this if you want to be sure

DirkEilander:
> I'm not sure, mamba env update may not delete already installed issues, so it's quite possible that the tests have access to a gdal even though they're not supposed to. There's not really a way to do this with the current cache setup, you'll have to make a new environment without the gdal to properly test this if you want to be sure

That's no problem, just something we should be aware of. I've just rebuild the environment and tested locally and that seems to work too.

DirkEilander:
Apart from the failing docker workflow this seems to work. @savente93 Could you check what is going on there?

dalmijn:
I don't know if it matters, but we do use gdal directly in gis_utils. Here we call it via a subprocess to create vrt's `create_vrt`.

DirkEilander:
> I don't know if it matters, but we do use gdal directly in gis_utils. Here we call it via a subprocess to create vrt's `create_vrt`.

Thanks for the comment. GDAL is still available because rasterio and shapely use it. But instead of getting making it an explicit dependency we rely on the version that is shipped those packages. It means we cannot do `from osgeo import gdal`, but we can call gdal tooling via rasterio/shapely and with subprocess. Removing gdal seems to resolve issues we've been having with pip install hydromt. 
Also the passing tests actually call create_vrt so that should still work!

DirkEilander:
Thanks @savente93 for the quick fix!

savente93 (approved):
LGTM

PR #585, opened 2023-10-16
Update installation of examples

### LINKED ISSUE #465: Conflict HydroMT version vs examples version

hboisgon:
## Issue addressed
Fixes #465

## Explanation
I think we should encourage users to install hydromt via conda-forge as much as possible and installing examples should also be made easy for users that are not too familiar with python/git.

In this PR, I then generate and commit on the fly an environment.yml for the examples if pyproject.toml is modified so that users can install hydromt from conda-forge as well as the optional dependencies.
In addition, we have a new command line for users to download the examples folder based on their installed hydromt version.

## Checklist

### DIFF
--- a/.github/workflows/update-examples-env.yml
+++ b/.github/workflows/update-examples-env.yml
@@ -1,0 +1,29 @@
+---
+name: Update examples environment
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - pyproject.toml
+  pull_request:
+    branches:
+      - main
+    paths:
+      - pyproject.toml
+
+jobs:
+  generate_examples_environment:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Generate environment file
+        run: pip install tomli && python make_env.py slim -n hydromt -r -o ./examples/environment.yml
+
+      - name: Commit environment file
+        uses: actions/update-file@v4
+        with:
+          path: ./examples/environment.yml
+          content: |
+            {{ include('examples/environment.yml') }}
--- a/.gitignore
+++ b/.gitignore
@@ -102,6 +102,7 @@ ENV/
 env.bak/
 venv.bak/
 *environment.yml
+!/examples/environment.yml
 
 # Spyder project settings
 .spyderproject
--- a/examples/environment.yml
+++ b/examples/environment.yml
@@ -1,0 +1,20 @@
+name: hydromt
+
+channels:
+- conda-forge
+
+dependencies:
+- cartopy
+- gcsfs
+- hydromt
+- jupyterlab
+- matplotlib-base
+- notebook
+- openpyxl
+- pillow
+- pip
+- requests
+- s3fs
+- xugrid>=0.6.5
+- pip:
+  - pyet
--- a/make_env.py
+++ b/make_env.py
@@ -48,6 +48,7 @@ parser.add_argument("--output", "-o", default="environment.yml")
 parser.add_argument("--channels", "-c", default=None)
 parser.add_argument("--name", "-n", default=None)
 parser.add_argument("--py-version", "-p", default=None)
+parser.add_argument("-r", "--release", action="store_true")
 args = parser.parse_args()
 
 #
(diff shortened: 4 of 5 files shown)


DirkEilander:
> I think we should encourage users to install hydromt via conda-forge as much as possible and installing examples should also be made easy for users that are not too familiar with python/git.

FYI: based on preliminary tests it seems that since PR #582 pip install is also possible without gdal errors. Still installing in a clean environment with conda is probably less prone to errors.

hboisgon:
> > I think we should encourage users to install hydromt via conda-forge as much as possible and installing examples should also be made easy for users that are not too familiar with python/git.
> 
> FYI: based on preliminary tests it seems that since PR #582 pip install is also possible without gdal errors. Still installing in a clean environment with conda is probably less prone to errors.

Nice! Will still use conda as for examples we need to install quite a lot of extra deps

savente93:
I'm a little confused, how can the user execute this command if they don't already have a hydromt installation?

savente93:
I hate to say it but I'm really not in favor of adding this. I think it adds a whole new surface area that will need maintenance, possibly forever, that, in my opinion, has a very small benefit. Perhaps this is tangential, but if HydroMT is going to go stable next year I think we have to start being much more careful in what kind of functionalities we add, and how long they have to stick around for. 

If users want to run the examples, there are executed versions of the examples in the documentation, and if users want to experiment without going through the installation process, there is binder as well as the docker images (though binder is probably more relevant to this discussion). 

As mentioned in my previous comment I also don't understand how users can execute this code without already having a HydroMT installation for which they need git. If they really want old versions of the examples, I'm happy to expand the documentation with some commands they can copy paste to get a hold of the right version. 

I think we'd be much better served by making sure our documentation and other processes are made as robust and smooth as possible, including for people who struggle with python (though HydroMT is a python tool, so I'd argue we can expect at least some basic python knowledge, but that is a different discussion), rather than adding less well maintained elephant paths for them. 

At the very least, if people are very intent on adding this, I'd like to see the functionality tested. 

I really hope this comes across as constructive because that's how I meant it, but I am skeptical if this is the right way to go. Hopefully I've managed to explain why in a good manner.

hboisgon:
After your remarks @savente93 and discussions with @DirkEilander then we decide to do two things for now:

- after PR #592 install hydromt via pip using the flavor slim
- for now explain better to users how to download the right examples based on their installed hydromt version (two ways with git clone and checkout for git users or manual download and unzipping to be explained)

This is still a temporary solution, we need to think of a better way to deal with examples in the future. A suggestion is to maybe create a separate hydromt-tutorials repository (what about examples from plugins? do we do the same?) or start publishing the examples with the releases or something else. We will decide what to do in the next release (will open a new issue for it).

For now in this PR, I will add the two changes mentioned above and remove the environment file and the download examples command.

Some extra remarks for you @savente93 :

- most hydromt users know very basic python and most of them know nothing about git
- for our users, I think they should install hydromt from conda-forge or pip release rather than dev installation
- binder or printed examples are nice but users cannot easily re-use them locally (eg on binder local data would not be available or the p drive data is not). Also nice examples on how to read data, build a model are nice for users to re-use for their own cases / data so this is why we would like to encourage users to re-use our examples notebooks outside of the docs/binder environment.

DirkEilander (approved):
Thanks @hboisgon! LGTM

Issue #31, opened 2021-05-12
Implement a generic clip method in the model API


hboisgon:
Add a generic clip method + CLI in the model API for:

- staticmaps
- staticgeoms
- forcing
- states

DirkEilander:
for an new functionality: https://github.com/Deltares/hydromt_wflow/issues/112

DirkEilander:
TODO:
- [ ] create `clip_forcing` and `clip_states` methods in `Model`
- [ ] create `clip_mesh` in `MeshMixin`
- [ ] create `clip_vector` in `VectorMixin`
- [ ] add these methods in the CLI clip method after checking if the mehod is available in the model

alimeshgi:
Closing  this issue. the referenced architecture (staticmaps, staticgeoms, MeshMixin, VectorMixin) was replaced by the component-based model API in v1.0, rendering this request obsolete. The clip CLI command was also removed during that refactoring. Please reopen with a fresh specification targeting the current component architecture (GridComponent, MeshComponent, VectorComponent) if model-level spatial clipping is still desired.

PR #612, opened 2023-10-20
Hot fix: update cli method

savente93:
## Explanation
this method still needed to be updated. This method is not present here, but defined and used in the plug ins. This should probagly be documented or abstracted a bit better at some point. Later tho.

## Checklist


dalmijn (approved):

DirkEilander:
Thanks for this quick update. The CLI clip method with `clip_grid` wil only work for `GridModel` in it's current implementation. It would be good to think about how to do this more robust and make it also work for other model types. At least a check if we are dealing with a `GridModel` before applying `clip_grid` would be good to do. To extend the functionality to other model types new `MeshModel.clip_mesh`, `VectorModel.clip_vector` methods would be required. See also #31

PR #615, opened 2023-10-25
Move formatting from black to ruff

savente93:
## Explanatin
Ruff just implemented a formater that is compatible with black (they claim 99.9% the same behavior) but is an order of magnitude faster. Since we also already install ruff for the pre-commit hooks and CIs it also simplifies that process. Therefore it would make sense to switch to that. 

## Checklist


Tjalling-dejong (approved):
LGTM 👍

PR #620, opened 2023-10-26
Improve adapter geodataset

### LINKED ISSUE #372: ENH: improve adapter for GeoDataset

Tjalling-dejong:
## Issue addressed
Fixes #372 

## Checklist

### DIFF
--- a/hydromt/data_adapter/dataset.py
+++ b/hydromt/data_adapter/dataset.py
@@ -1,0 +1,81 @@
+"""Implementation for the dataset DataAdapter."""
+import logging
+from datetime import datetime
+from pathlib import Path
+from typing import List, NewType, Optional, Tuple
+
+import xarray as xr
+
+from .data_adapter import DataAdapter
+
+logger = logging.getLogger(__name__)
+
+DatasetSource = NewType("DatasetSource", str | Path)
+
+
+class DatasetAdapter(DataAdapter):
+    """DatasetAdapter for non-spatial n-dimensional data."""
+
+    _DEFAULT_DRIVER = ""
+    _DRIVERS = {
+        "nc": "netcdf",
+    }
+
+    def __init__(
+        self,
+        path: str | Path,
+        driver: Optional[str] = None,
+        filesystem: Optional[str] = None,
+        nodata: Optional[dict | float | int] = None,
+        rename: Optional[dict] = None,
+        unit_mult: Optional[dict] = None,
+        unit_add: Optional[dict] = None,
+        meta: Optional[dict] = None,
+        attrs: Optional[dict] = None,
+        driver_kwargs: Optional[dict] = None,
+        storage_options: Optional[dict] = None,
+        name: Optional[str] = "",
+        catalog_name: Optional[str] = "",
+        provider: Optional[str] = None,
+        version: Optional[str] = None,
+    ):
+        """Docstring."""
+        super().__init__(
+            path,
+            driver,
+            filesystem,
+            nodata,
+            rename,
+            unit_mult,
+            unit_add,
+            meta,
+            attrs,
+            driver_kwargs,
+            storage_options,
+            name,
+            catalog_name,
+            provider,
+            version,
+        )
+
+    def to_file(
+        self,
+        data_root: str | Path,
+        data_name: str,
+        time_tuple: Optional[Tuple[str | datetime]] = None,
+        variables: Optional[List[str]] = None,
+        driver: Optional[str] = None,
+        **kwargs,
+    ) -> Tuple[str, str]:
+        """Docstring."""
+        pass
+
+    def get_data(
+        self,
+        variables: Optional[List[str]] = None,
+        time_tuple: Optional[Tuple[str | datetime]] = None,
+        single_var_as_array: Optional[bool] = True,
+        logger: Optional[logging.Logger] = logger,
+    ) -> xr.Dataset:
+        """Docstring."""
+        pass
--- a/hydromt/data_adapter/dataset.py
+++ b/hydromt/data_adapter/dataset.py
@@ -1,11 +1,16 @@
 """Implementation for the dataset DataAdapter."""
 import logging
+import os
 from datetime import datetime
+from os.path import join
 from pathlib import Path
 from typing import List, NewType, Optional, Tuple
 
+import numpy as np
+import pandas as pd
 import xarray as xr
 
+from .. import io
 from .data_adapter import DataAdapter
 
 logger = logging.getLogger(__name__)
@@ -41,21 +46,21 @@ class DatasetAdapter(DataAdapter):
     ):
         """Docstring."""
         super().__init__(
-            path,
-            driver,
-            filesystem,
-            nodata,
-            rename,
-            unit_mult,
-            unit_add,
-            meta,
-            attrs,
-            driver_kwargs,
-            storage_options,
-            name,
-            catalog_name,
-            provider,
-            version,
+            path=path,
+            driver=driver,
+            filesystem=filesystem,
+            nodata=nodata,
+            rename=rename,
+            unit_mult=unit_mult,
+            unit_add=unit_add,
+            meta=meta,
+            attrs=attrs,
+            driver_kwargs=driver_kwargs,
+            storage_options=storage_options,
+            name=name,
+            catalog_name=catalog_name,
+            provider=provider,
+            version=version,
         )
 
     def to_file(
@@ -68,7 +73,36 @@ class DatasetAdapter(DataAdapter):
         **kwargs,
     ) -> Tuple[str, str]:
         """Docstring."""
-        pass
+        obj = self.get_data(
+            time_tuple=time_tuple,
+            variables=variables,
+            logger=logger,
+            single_var_as_array=variables is None,
+        )
+        if driver is None or driver == "netcdf":
+            # always write netcdf
+            driver = "netcdf"
+            dvars = [obj.name] if isinstance(obj, xr.DataArray) else obj.vector.vars
+            if variables is None:
+                encoding = {k: {"zlib": True} for k in dvars}
+                fn_out = join(data_root, f"{data_name}.nc")
+                obj.to_netcdf(fn_out, encoding=encoding)
+            else:  # save per variable
+                if not os.path.isdir(join(data_root, data_name)):
+                    os.makedirs(join(data_root, data_name))
+                for var in dvars:
+                    fn_out = join(data_root, data_name, f"{var}.nc")
+                    obj[var].to_netcdf(fn_out, encoding={var: {"zlib": True}})
+                fn_out = join(data_root, data_name, "{variable}.nc")
+        elif driver == "zarr":
+            fn_out = join(data_root, f"{data_name}.zarr")
+            if isinstance(obj, xr.DataArray):
+                obj = obj.to_dataset()
+            obj.to_zarr(fn_out, **kwargs)
+        else:
+            raise ValueError(f"Dataset: Driver {driver} unknown.")
+
+        return fn_out, driver
 
     def get_data(
         self,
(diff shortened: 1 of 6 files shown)


hydromt/data_catalog.py
```diff
@@ -1594,6 +1595,85 @@ def get_geodataset(
         )
         return obj
 
+    def get_dataset(
```
savente93, line 1598: Not sure if this PR is the right place for it, or if this is a thing for v1.0, but I'm really not a fan of this name, because it implies (to me atleast) that the other dataset kinds like Geodataset etc. inheret from this one. It also does not imply to me anything that the dataset might actually contain, but I do realise it is mostly in line with the other names, so I would be okay with keeping it like this for now, and including this in the larger renaming discussion for 1.0

hydromt/data_catalog.py
```diff
@@ -1594,6 +1595,85 @@ def get_geodataset(
...
+        **kwargs,
+    ) -> xr.Dataset:
+        """Return a clipped, sliced and unified Dataset.
+
+        To slice the data to the time period of interest, provide the
+        `time_tuple` argument. To return only the dataset variables
+        of interest provide the `variables` argument.
+
+        NOTE: Unless `single_var_as_array` is set to False a single-variable data source
+        will be returned as xarray.DataArray rather than Dataset.
```
savente93, line 1615: Especially now we need to start being a little more careful with making clear whether we mean a HydroMT dataset or an xarray dataset.

hydromt/data_catalog.py
```diff
@@ -1594,6 +1595,85 @@ def get_geodataset(
...
+
+        Arguments
+        ---------
+        data_like: str, Path, xr.Dataset, xr.DataArray
```
savente93, line 1619: missing `SourceSpecDict`

hydromt/data_catalog.py
```diff
@@ -1594,6 +1595,85 @@ def get_geodataset(
...
+                data_like, provider, version
+            )
+        if isinstance(data_like, (str, Path)):
+            if isinstance(data_like, str) and data_like in self.sources:
+                name = data_like
+                source = self.get_source(name, provider=provider, version=version)
+            elif os.path.exists(abspath(data_like)):
+                path = str(abspath(data_like))
+                if "provider" not in kwargs:
+                    kwargs.update({"provider": "local"})
```
savente93, line 1651: not necessarily true, if I only have one provider but I load it from an `s3` path, it's not a local provider I'd say. If it's indeed a local path this is correct

hydromt/data_catalog.py
```diff
@@ -1594,6 +1595,85 @@ def get_geodataset(
...
+        """
+        if isinstance(data_like, dict):
+            data_like, provider, version = _parse_data_like_dict(
+                data_like, provider, version
+            )
+        if isinstance(data_like, (str, Path)):
+            if isinstance(data_like, str) and data_like in self.sources:
+                name = data_like
+                source = self.get_source(name, provider=provider, version=version)
+            elif os.path.exists(abspath(data_like)):
```
savente93, line 1648: does abspath work on something like an `s3` path?

hydromt/data_catalog.py
```diff
@@ -1594,6 +1595,85 @@ def get_geodataset(
...
+                self.add_source(name, source)
+            else:
+                raise FileNotFoundError(f"No such file or catalog source: {data_like}")
+        elif isinstance(data_like, (xr.DataArray, xr.Dataset)):
+            data_like = DatasetAdapter._slice_data(
+                data_like,
+                variables,
+                time_tuple,
+                logger=self.logger,
+            )
```
savente93, line 1663: Not sure if this was merged yet but this might be a good place to apply the new error handeling pattern we're trying to introduce. If it's not already in the underlying data adapters I"m okay with merging like this aand making a ticket to implement this further

hydromt/data_adapter/utils.py
```diff
@@ -0,0 +1,105 @@
...
+
+    Parameters
+    ----------
+    obj : xr.Dataset | xr.DataArray
+        Dataset.
+    data_root : str | Path
+        root to write the data to.
+    data_name : str
+        filename to write to.
+    variables : Optional[List[str]], optional
```
savente93, line 29: second optinal is redundant

hydromt/data_adapter/utils.py
```diff
@@ -0,0 +1,105 @@
...
+        list of dataset variables to write, by default None
+
+    Returns
+    -------
+    fn_out: str
+        Absolute path to output file
+    """
+    dvars = [obj.name] if isinstance(obj, xr.DataArray) else obj.data_vars
+    if variables is None:
+        encoding = {k: {"zlib": True} for k in dvars}
```
savente93, line 39: perhaps this needs to be an argument as well? or must it always be `zlib` for netcdf?

hydromt/data_adapter/utils.py
```diff
@@ -0,0 +1,105 @@
...
+    return fn_out
+
+
+def shift_dataset_time(dt: int, ds: xr.Dataset, logger: logging.Logger) -> xr.Dataset:
+    """Shifts time of a xarray dataset.
+
+    Parameters
+    ----------
+    dt : int
+        time delta to shift the time of the dataset
```
savente93, line 84: what's the unit to shift by? seconds? days? fortnights?

hydromt/data_adapter/utils.py
```diff
@@ -0,0 +1,105 @@
...
+        time shifted dataset
+    """
+    if (
+        dt != 0
+        and "time" in ds.dims
+        and ds["time"].size > 1
+        and np.issubdtype(ds["time"].dtype, np.datetime64)
+    ):
+        logger.debug(f"Shifting time labels with {dt} sec.")
+        ds["time"] = ds["time"] + pd.to_timedelta(dt, unit="s")
```
savente93, line 102: might be nice to make the time unit an optional argument you take here with `s` as a default, but is not necessary

tests/test_data_adapter.py
```diff
@@ -301,6 +302,132 @@ def test_geodataset_set_nodata(artifact_data: DataCatalog):
...
+    dataset_adapter = DatasetAdapter(path=zarr_path, driver="fake-driver")
+    with pytest.raises(ValueError, match=r"Dataset: Driver fake-driver unknown"):
+        dataset_adapter.get_data(variables=["col1", "col2"])
+
+
+def test_dataset_set_nodata(tmpdir, timeseries_ds):
+    path = str(tmpdir.join("test.nc"))
+    timeseries_ds.to_netcdf(path)
+
+    nodata = -999
```
savente93, line 363: out of curiosity, does it also accept float nodata?

tests/test_data_adapter.py
```diff
@@ -301,6 +302,132 @@ def test_geodataset_set_nodata(artifact_data: DataCatalog):
...
+
+def test_dataset_to_stac_catalog(tmpdir, timeseries_ds):
+    path = str(tmpdir.join("test.nc"))
+    timeseries_ds.to_netcdf(path)
+    dataset_adapter = DatasetAdapter(path=path, name="timeseries_dataset")
+
+    stac_catalog = dataset_adapter.to_stac_catalog()
+    assert isinstance(stac_catalog, StacCatalog)
+    stac_item = next(stac_catalog.get_items("timeseries_dataset"), None)
+    assert list(stac_item.assets.keys())[0] == "test.nc"
```
savente93, line 428: If possible I'd also like to see a `from_stac` implementation but there is also a ticket to improve this for other datasets (#626) if it's a lot of work to include that here we can leave it for that one.

hydromt/data_adapter/dataset.py
```diff
@@ -0,0 +1,431 @@
...
+        driver_kwargs: Optional[dict] = None,
+        storage_options: Optional[dict] = None,
+        name: Optional[str] = "",
+        catalog_name: Optional[str] = "",
+        provider: Optional[str] = None,
+        version: Optional[str] = None,
+    ):
+        """Initiate data adapter for n-dimensional timeseries data.
+
+        This object contains all properties required to read supported files into
```
savente93, line 54: do we know which file types are supported? can we link to it somewhere? if not I think we should make a ticket for this. It would be good to add to the documentation.

hydromt/data_adapter/dataset.py
```diff
@@ -0,0 +1,431 @@
...
+            Additional keyword arguments that are passed to the `to_zarr`
+            function.
+
+        Returns
+        -------
+        fn_out: str
+            Absolute path to output file
+        driver: str
+            Name of driver to read data with, see
+            :py:func:`~hydromt.data_catalog.DataCatalog.get_dataset`
```
savente93, line 161: I actually like this, but I'm not sure if the other adapters do this? If not we can make a ticket to add that later, but it's a nice pattern for sure.

hydromt/data_adapter/dataset.py
```diff
@@ -0,0 +1,431 @@
...
+        data_name : str
+            Name of output file without extension.
+        variables : list of str, optional
+            Names of Dataset variables to return. By default all dataset variables
+            are returned.
+        time_tuple : tuple of str, datetime, optional
+            Start and end date of period of interest. By default the entire time period
+            of the dataset is returned.
+        driver : str, optional
+            Driver to write file, e.g.: 'netcdf', 'zarr', by default None
```
savente93, line 150: good to mention that `netcdf` is used in the `None` option

hydromt/data_adapter/dataset.py
```diff
@@ -0,0 +1,431 @@
...
+            )
+        else:
+            raise ValueError(f"Dataset: Driver {driver} unknown.")
+
+        return fn_out, driver
+
+    def get_data(
+        self,
+        variables: Optional[List[str]] = None,
+        time_tuple: Optional[Tuple[str | datetime]] = None,
```
savente93, line 185: same as other time_tuple comments

hydromt/data_adapter/dataset.py
```diff
@@ -0,0 +1,431 @@
...
+        ds = self._set_metadata(ds)
+        # return array if single var and single_var_as_array
+        return self._single_var_as_array(ds, single_var_as_array, variables)
+
+    def _read_data(self, fns, logger=logger):
+        kwargs = self.driver_kwargs.copy()
+        if len(fns) > 1 and self.driver in ["zarr"]:
+            raise ValueError(
+                f"Dataset: Reading multiple {self.driver} files is not supported."
+            )
```
savente93, line 215: we should probably document, or implement this if it is actually possible for the ther formats. Could you make tickets as necessary?

hydromt/data_adapter/dataset.py
```diff
@@ -0,0 +1,431 @@
...
+            ds = ds.to_dataset()
+        elif variables is not None:
+            variables = np.atleast_1d(variables).tolist()
+            if len(variables) > 1 or len(ds.data_vars) > 1:
+                mvars = [var not in ds.data_vars for var in variables]
+                if any(mvars):
+                    raise ValueError(f"Dataset: variables not found {mvars}")
+                ds = ds[variables]
+        if time_tuple is not None:
+            ds = DatasetAdapter._slice_temporal_dimension(ds, time_tuple, logger=logger)
```
savente93, line 320: This will fail I think if the slice would return an empty dataset because of the underlying code. This would be a good place to apply the new error pattern

savente93 (changes requested):
Nice work Tjalling. I've got a few things I'd like to see adressed, but the bulk looks good, and the tests seem well setup so already well done!

hydromt/data_catalog.py
Tjalling-dejong: yes, I agree. The name can be a bit confusing. With regards to xarray dataset, I think it is fine the way it is. The get_dataset method will always return a xr.Dataset anyway. 
That being said it is a good idea to take this up in the discussion about names.

hydromt/data_catalog.py
Tjalling-dejong: this is still code copied from the get_geodataset method. I see that in the mean time this has changed.

hydromt/data_catalog.py
Tjalling-dejong: I dont know

hydromt/data_adapter/utils.py
Tjalling-dejong: it was already hard-coded in the geodatasetadapter as zlib. I now made it a default for this function.

tests/test_data_adapter.py
Tjalling-dejong: yes, you can set any no data value

hydromt/data_adapter/dataset.py
Tjalling-dejong: for now netcdf and zarr are supported. We still have to work on supporting multiple csv files.

hydromt/data_adapter/dataset.py
Tjalling-dejong: See #651

hydromt/data_catalog.py
Tjalling-dejong: I changed it so that is the same as get_geodataset

hydromt/data_catalog.py
Tjalling-dejong: maybe best to do this in a separate issue

DirkEilander:
One quick remark. Please also update the API reference in the docs (docs/api.rst)

savente93 (changes requested):
The changes look good but you need to be up to date with main to make sure it still works on python 3.9. I'll be happy to approve once it's updated and the tests pass

savente93 (approved):
Nice, LGTM

PR #621, opened 2023-10-27
Handle nodata data adapter

### LINKED ISSUE #584: `GeoDataFrameAdapter` slices data even if the source is empty for the geometry
### HydroMT version checks

### Reproducible Example

```python
import geopandas as gpd
from hydromt.data_adapter import GeoDataFrameAdapter
from shapely import box

fn = "s3://hydromt-data/hydrography/reservoirs/reservoir-db.fgb"
geom = (9.654999999999973, 0.3475000000002808, 9.861666666666451, 0.4866666666668209)
if __name__ == "__main__":
    da = GeoDataFrameAdapter(path=fn, driver="vector")
    gdf = da.get_data(bbox=None, geom=gpd.GeoSeries(box(*geom)).set_crs(epsg=4326))
    gdf
```

### Current behaviour

Raises Exception because of no data in the current DataFrame

### Desired behaviour

Should not raise exception

### Additional context

_No response_

Jaapel:
## Issue addressed
Fixes #584 

## Explanation
`DataAdapter._slice_data` and `DataCatalog.get_<data type>` now have a `handle_nodata` argument.

## Checklist

`DataAdapter.get_data` not yet implemented with this feature, but defaults still maintain the same behaviour. Only the error type returned is different. I still propose to refactor the DataAdapter vs. DataCatalog relation, so that we can change this behaviour in one place only.


Jaapel:
@savente93 I get a SonarCube exception for code I did not commit, is this desired behaviour?

savente93:
@Jaapel Sonar cube has a very high false positive rate, especially when it comes to duplication. I looked around but cannot find any place where I can tune or disable it. I think it's fine to ignore for now.

hydromt/nodata.py
```diff
@@ -0,0 +1,22 @@
...
+from logging import Logger
+
+from .exceptions import NoDataException
+
+
+class NoDataStrategy(Enum):
+    """Strategy to handle nodata values."""
+
+    RAISE = "raise"
+    IGNORE = "ignore"
```
savente93, line 12: Perhaps a "coerce" option might be useful in some places (like STAC), but that is harder to handle generically so I'll leave it up to you whether you want to add it here.

hydromt/nodata.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def _exec_strat(msg: str, strategy: NoDataStrategy, logger: Logger) -> None:
```
savente93, line 15: Could you rename it to `_exec_nodata_strat`? out in the wild it's not that clear what it's actually referring too.

savente93 (approved):
Very nice, two small comments, but otherwise LGTM

hydromt/nodata.py
Jaapel: We can add it once we implement it :)

PR #632, opened 2023-11-03
Add validation methods for data catalog

### LINKED ISSUE #236: Create validate methods for data catalog and configuration file
Right now, if something is wrong in a data catalog yml file or configuration ini file, the error messages can be a bit cryptic (example typo diver instead of driver in yml...).
Would be nice to get some validate methods either in the CLI or before started build/update so that users know something is wrong in the yml or ini file and not within hydromt.

savente93:
## Issue addressed
Fixes #236 

## Explanation
Use Pydantic models for validating rules about the data catalog. This PR has some limitations which I'll list below but also lays a lot of groundwork. I'm actually pretty happy with the results. When this gets merged I'll make tickets to address the limitations. In the spirit of agile I think it's better to just merge this now, and address the further issues down the line so we can keep iterating it as we add features to it. 

### Limitations (tick boxes are for tickets created):
- [ ] No Data is validated, only the config files
- [ ] The model config can't currently actually validate anything (though there are tests for it) because I couldn't figure out how the dynamic function loading worked, so I'll need some help with this. 
- [ ] I made a start on being able to validate region arguments, and got a fair way but it is incomplete. I didn't quite understand for example how sub basin ids worked. I think the code that is there could also do with some refactoring when I get around to it. 
- [ ] Because model config and region checking aren't functional yet, they are not accessible from the CLI, though the objects to do the validation can be imported.
- [ ] The validation is only used in the new `check` subcommand of the cli and isn't actually used throughout the code base yet. I think it's better to test the feature first for a bit and see how it does before we spread it throughout the code base. 

## Checklist

### DIFF
--- a/hydromt/validators/data_catalog.py
+++ b/hydromt/validators/data_catalog.py
@@ -1,0 +1,132 @@
+from pathlib import Path
+from typing import Any, Dict, Optional, Union
+
+from pydantic import AnyUrl, BaseModel, ConfigDict
+from pydantic_core import Url
+
+from hydromt.typing import Bbox, Number, TimeRange
+
+
+class SourceSpecDict(BaseModel):
+    source: str
+    provider: Optional[str] = None
+    version: Optional[Union[str, int]] = None
+
+    @staticmethod
+    def from_dict(input_dict):
+        return SourceSpecDict(**input_dict)
+
+
+class Extent(BaseModel):
+    time_range: TimeRange
+    bbox: Bbox
+
+
+class DataCatalogMetaData(BaseModel):
+    root: Optional[Path] = None
+    version: Optional[Union[str, int]] = None
+    name: Optional[str] = None
+    model_config = ConfigDict(extra="allow")
+
+    @staticmethod
+    def from_dict(input_dict):
+        if input_dict is None:
+            return DataCatalogMetaData()
+        else:
+            return DataCatalogMetaData(**input_dict)
+
+
+class DataCatalogItemMetadata(BaseModel):
+    category: Optional[str] = None
+    paper_doi: Optional[str] = None
+    paper_ref: Optional[str] = None
+    source_license: Optional[str] = None
+    source_url: Optional[AnyUrl] = None
+    source_version: Optional[str] = None
+    notes: Optional[str] = None
+
+    @staticmethod
+    def from_dict(input_dict):
+        if input_dict is None:
+            return DataCatalogItemMetadata()
+        else:
+            item_source_url = input_dict.pop("source_url", None)
+            if item_source_url:
+                source_url = Url(item_source_url)
+            else:
+                source_url = None
+            return DataCatalogItemMetadata(**input_dict, source_url=item_source_url)
+
+
+class DataCatalogItem(BaseModel):
+    name: str
+    data_type: str
+    driver: str
+    path: Path
+    crs: Optional[int] = None
+    filesystem: Optional[str] = None
+    kwargs: Dict[str, Any] = {}
+    storage_options: Dict[str, Any] = {}
+    rename: Dict[str, str] = {}
+    nodata: Optional[Number] = None
+    meta: Optional[DataCatalogItemMetadata] = None
+    unit_add: Optional[Dict[str, Number]] = None
+    unit_mult: Optional[Dict[str, Number]] = None
+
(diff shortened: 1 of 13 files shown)


savente93:
The things that sonar cloud is complaining about are very silly like the fact that one of the source_urls is `http` and not `https` so I think it's okay to ignore on this occasion

hydromt/validators/data_catalog.py
```diff
@@ -0,0 +1,163 @@
...
+
+    name: str
+    data_type: str
+    driver: str
+    path: Path
+    crs: Optional[int] = None
+    filesystem: Optional[str] = None
+    kwargs: Dict[str, Any] = {}
+    storage_options: Dict[str, Any] = {}
+    rename: Dict[str, str] = {}
```
Jaapel, line 90: It is better to use non-mutable defaults. `Field(default_factory=dict)`

hydromt/validators/data_catalog.py
```diff
@@ -0,0 +1,163 @@
...
+        item_metadata = DataCatalogItemMetadata.from_dict(input_dict.pop("meta", {}))
+        item_kwargs = input_dict.pop("kwargs", {})
+        item_storage_options = input_dict.pop("storage_options", {})
+        return DataCatalogItem(
+            **input_dict,
+            name=entry_name,
+            kwargs=item_kwargs,
+            storage_options=item_storage_options,
+            meta=item_metadata,
+        )
```
Jaapel, line 118: This is an alternative for the built-in `DataCatalogItem.model_validate` static method (parses python dict and returns an instance of the model). Do you want to overwrite this method or is this an alternative?

hydromt/validators/data_catalog.py
```diff
@@ -0,0 +1,163 @@
...
+
+
+class DataCatalogValidator(BaseModel):
+    """A validated complete data catalog."""
+
+    meta: Optional[DataCatalogMetaData] = None
+    sources: Dict[str, DataCatalogItem] = {}
+    aliases: Dict[str, str] = {}
```
Jaapel, line 126: Same as above, `Field(default_factory=dict)`

hydromt/validators/data_catalog.py
```diff
@@ -0,0 +1,163 @@
...
+
+    model_config: ConfigDict = ConfigDict(
+        str_strip_whitespace=True,
+        extra="forbid",
+    )
+
+    @staticmethod
+    def from_dict(input_dict):
```
Jaapel, line 134: Same here, overwrite `BaseModel.model_validate` or new method?

hydromt/validators/model_config.py
```diff
@@ -0,0 +1,22 @@
...
+from typing import Any, Dict
+
+from pydantic import BaseModel
+
+
+class HydromtStep(BaseModel):
+    """A Pydantic model for the validation of model config files."""
+
+    # fn: Callable
+    # args: Dict[str, Any]
```
Jaapel, line 11: Some commented code here

hydromt/validators/model_config.py
```diff
@@ -0,0 +1,22 @@
...
+    fn: str
+    args: Dict[str, Any]
+
+    @staticmethod
+    def from_dict(input_dict):
+        """Generate a validated model of a step in a model config files."""
+        fn_name, arg_dict = next(iter(input_dict.items()))
+        # TODO figure out where funcs are actually comming from
+        # fn = getattr(hydromt, fn_name)
+        # return HydromtStep(fn=getattr(hydromt, fn_name), args=arg_dict)
```
Jaapel, line 21: I do not understand this TODO yet :)

hydromt/validators/region.py
```diff
@@ -0,0 +1,239 @@
+"""Pydantic models for the validation of region specifications."""
```
Jaapel, line 1: Looks good, but you can check out geojson_pydantic from developmentseed to get some inspiration / reuse their work if you want to deal with geometries.

Jaapel (approved):
Small comments, LGTM

hydromt/validators/data_catalog.py
savente93: It might be interesting to explore this in the future, but for now, I'll stick with the current approach. Good comment!

PR #643, opened 2023-11-10
Model config validation

### LINKED ISSUE #637: Model config validation should be able to determine that functions are actually present
### Kind of request

### Enhancement Description

The model config validation doesn't currently work because I couldn't understand the dynamic loading of functions. This should be rectified before the validation can be to use. 

### Use case

The validation should be able to take in a config file and at least determine if all the functions in the config can be found. Argument validation would be really cool if possible, but that may or may not be realistic.

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #637 

## Explanation
I've copied what I can from how the model code actually does this. I've done input argument checking where I could but this is not possible everywhere especially where dictionary arguments are taken. I think this is a nice result for now though. 

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,6 +18,7 @@ Added
 - Pixi is now available as an additional task runner. (#634)
 - Support exporting data from catalogs from the CLI. (#627)
 - Support for validating data catalogs from the CLI. (#632)
+- Support for validating model configurations from the CLI. (#643)
 
 
 Changed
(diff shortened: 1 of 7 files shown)


savente93:
Wasn't quite sure whether docs would need to be updated on this and if so where so I'll leave that up to the reviewer.

Tjalling-dejong (approved):
Tests on wsl are running fine and coverage is also good 👍

Issue #647, opened 2023-11-14
release v0.9.1


DirkEilander:
### Kind of request

None

### Enhancement Description

We require a v0.9.1 release because of some minor bugs in v0.9 currently blocking HydroMT-Wflow and HydroMT-SFINCS to release with the latest HydroMT version. 
This release should include fixes for #641, #624 and possibly #648

savente93:
Just waiting for #649 to be merged, then we get this show on the road

PR #649, opened 2023-11-14
Fix meridian offset

### LINKED ISSUE #624: Transform/ boundary does not correspond with extent from data
### HydroMT version checks

### Reproducible Example

Use: `GridModel` (no underscore) branch from hydromt_wflow

Run this script in debug mode.
n:\Deltabox\Postbox\Dalmijn, Brendan\PS\setup_wflow_test.py
(maybe drag it to a local place with the accompanying geojson).

Break here:
- `setup_laimaps` (wflow.py; hydromt_wflow): https://github.com/Deltares/hydromt_wflow/blob/af63b5a62d3d1ede84232fd9b9ab003477032948/hydromt_wflow/wflow.py#L919

This then goes through `get_rasterdataset` (data_catalog.py) and `get_data` (rasterdataset.py) and `_slice_data` (rasterdataset.py).

Break here:
- `_slice_spatial_dimensions` (rasterdataset.py): https://github.com/Deltares/hydromt/blob/f3be55fb3aaf44e1aceec0fe9acac82d6515768a/hydromt/data_adapter/rasterdataset.py#L547
- `clip_bbox` (raster.py): https://github.com/Deltares/hydromt/blob/f3be55fb3aaf44e1aceec0fe9acac82d6515768a/hydromt/raster.py#L1226

Then in `clip` the actual clipping happens but it need the indices from `clip_bbox`.  

### Current behaviour

While building a model with hydromt_wflow I noticed that the 'LAI' (leaf area index) was full of nodata values. 

When going through the steps (mentioned above) I noticed that clip_bbox in and on itself uses indices (derived from the transform) instead of coordinates (like it did in 0.8).

This example concerns itself with 'modis_lai' (rasterdataset from 'deltares_data'). 
This dataset had an extent in the x direction from -180 to 180. However the transform and the bounds reflect an extent from -180 to 167.99058.

This causes the indexing to be off by something like 2 degrees. 
Which coupled with the `reproject_like` to the overal dataset (with correct coordinates) causes nodata values.

(EDIT)
It happens here:
https://github.com/Deltares/hydromt/blob/f3be55fb3aaf44e1aceec0fe9acac82d6515768a/hydromt/gis_utils.py#L306-L322

'modis_lai' slightly excedes 180 degrees in the east. This prompts this method to kick into action, which then assumes that the data must be from 0-360 degrees (which its not). The eventual `sortby` (return statement) method then solidifies the wrong extent .

### Desired behaviour

I guess that the transform should be correct in regards to the coordinates shown by the data.

### Additional context

_No response_

DirkEilander:
## Issue addressed
Fixes #624 

## Explanation
The error was because the the meridian_offset function created an irregular grid as some cells where incorrectly shifted.
The solution checks if the grid is global (x spans 360 degrees) before shifting which is required to keep a regular grid.
In addition we check if the grid is regular when computing the resolution to avoid these funny downstream errors.

## Checklist

### DIFF
--- a/hydromt/gis_utils.py
+++ b/hydromt/gis_utils.py
@@ -303,17 +303,18 @@ def axes_attrs(crs):
     return x_dim, y_dim, x_attrs, y_attrs
 
 
-def meridian_offset(ds, x_name="x", bbox=None):
+def meridian_offset(ds, bbox=None):
     """re-arrange data along x dim."""
-    if ds.raster.crs is None or ds.raster.crs.is_projected:
-        raise ValueError("The method is only applicable to geographic CRS")
+    w, _, e, _ = ds.raster.bounds
+    if ds.raster.crs is None or ds.raster.crs.is_projected or (e - w) < (360 - 1e-3):
+        raise ValueError("The method is only applicable to global geographic CRS")
+    x_name = ds.raster.x_dim
     lons = np.copy(ds[x_name].values)
-    w, e = lons.min(), lons.max()
     if bbox is not None and bbox[0] < w and bbox[0] < -180:  # 180W - 180E > 360W - 0W
         lons = np.where(lons > 0, lons - 360, lons)
     elif bbox is not None and bbox[2] > e and bbox[2] > 180:  # 180W - 180E > 0E-360E
         lons = np.where(lons < 0, lons + 360, lons)
-    elif e > 180:  # 0E-360E > 180W - 180E
+    elif e >= (360 - 1e-3):  # 0E-360E > 180W - 180E (allow for rounding errors)
         lons = np.where(lons > 180, lons - 360, lons)
     else:
         return ds
--- a/hydromt/gis_utils.py
+++ b/hydromt/gis_utils.py
@@ -306,7 +306,11 @@ def axes_attrs(crs):
 def meridian_offset(ds, bbox=None):
     """re-arrange data along x dim."""
     w, _, e, _ = ds.raster.bounds
-    if ds.raster.crs is None or ds.raster.crs.is_projected or (e - w) < (360 - 1e-3):
+    if (
+        ds.raster.crs is None
+        or ds.raster.crs.is_projected
+        or not np.isclose(e - w, 360)
+    ):
         raise ValueError("The method is only applicable to global geographic CRS")
     x_name = ds.raster.x_dim
     lons = np.copy(ds[x_name].values)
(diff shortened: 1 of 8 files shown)


savente93:
I'll have a look in just a sec, but also for my info, does this need to be included in v0.9.1?

DirkEilander:
> I'll have a look in just a sec, but also for my info, does this need to be included in v0.9.1?

Yes, see also the updated #647

hydromt/gis_utils.py
```diff
@@ -303,17 +303,22 @@ def axes_attrs(crs):
...
     """re-arrange data along x dim."""
-    if ds.raster.crs is None or ds.raster.crs.is_projected:
-        raise ValueError("The method is only applicable to geographic CRS")
+    w, _, e, _ = ds.raster.bounds
+    if (
+        ds.raster.crs is None
+        or ds.raster.crs.is_projected
+        or not np.isclose(e - w, 360)
+    ):
+        raise ValueError("The method is only applicable to global geographic CRS")
```
savente93, line 314: would be nice to mention in the error which f the conditions failed, but this is also acceptable

hydromt/raster.py
```diff
@@ -560,18 +555,25 @@ def res(self) -> tuple[float, float]:
...
-                dy = math.hypot(ddx0, ddy0)
+                dxs, dys = np.diff(xs[0, :]), np.diff(ys[:, 0])
+                ddx0 = xs[-1, 0] - xs[0, 0]
+                ddy0 = ys[-1, 0] - ys[0, 0]
+                ddx1 = xs[0, -1] - xs[0, 0]
+                ddy1 = ys[0, -1] - ys[0, 0]
+                dx = math.hypot(ddx1, ddy1) / (xs.shape[1] - 1)  # always positive!
+                dy = math.hypot(ddx0, ddy0) / (xs.shape[0] - 1)
+            # check if regular grid; not water tight for rotated grids, but close enough
+            # allow rtol 0.05% * resolution for rounding errors in geographic coords
```
savente93, line 572: should we document this tolerance somewhere? Also the tolerance below is set using `atol` not `rtol`, is this intentional?

hydromt/raster.py
```diff
@@ -2364,8 +2366,15 @@ def to_xyz_tiles(
...
                 if any((s > 0 for s in pad_sizes)):
-                    pad_sizes = {d: (0, s) for d, s in zip(dims, pad_sizes) if s > 0}
-                    temp = temp.pad(pad_sizes, mode="constant", constant_values=nodata)
+                    pad_dict = {d: (0, s) for d, s in zip(dims, pad_sizes) if s > 0}
+                    temp = temp.pad(pad_dict, mode="constant", constant_values=nodata)
+                    # pad nan coordinates; this is not done by xarray
+                    for d, s in zip(dims, pad_sizes):
+                        if s > 0:
+                            coords = temp[d].values
+                            res = dst_res if d == x_dim else -dst_res
```
savente93, line 2375: can resolutions be negative? this seems odd to me, but perhaps I'm not quite understanding what's happening here.

savente93 (commented):
couple of small comments, but mostly looks fine. Once the tests pass and Brendan has confirmed it solves his problem I'll be happy to merge

hydromt/raster.py
DirkEilander: This is intentional to work with resolution in projected units (metre) and geographic units degrees

hydromt/raster.py
DirkEilander: Yes, the orientation of the grid is also captured in the res. A grid which is is oriented north > south has a negative y res

dalmijn (approved):
lgtm and tested against own case.

PR #655, opened 2023-11-17
fix typing annotation for py39

DirkEilander:
## Issue addressed
Fixes typing issue when hydromt.gis_utils is loaded in py39

```
ImportError while loading conftest '/home/runner/work/hydromt_sfincs/hydromt_sfincs/tests/conftest.py'.
tests/conftest.py:7: in <module>
    from hydromt_sfincs.sfincs import SfincsModel
hydromt_sfincs/__init__.py:11: in <module>
    from .sfincs import *
hydromt_sfincs/sfincs.py:14: in <module>
    import hydromt
/usr/share/miniconda/envs/hydromt-sfincs-dev/lib/python3.[9](https://github.com/Deltares/hydromt_sfincs/actions/runs/6902656342/job/18779801237?pr=147#step:6:10)/site-packages/hydromt/__init__.py:17: in <module>
    from . import cli, flw, raster, stats, vector, workflows
/usr/share/miniconda/envs/hydromt-sfincs-dev/lib/python3.9/site-packages/hydromt/flw.py:[12](https://github.com/Deltares/hydromt_sfincs/actions/runs/6902656342/job/18779801237?pr=147#step:6:13): in <module>
    from . import gis_utils
/usr/share/miniconda/envs/hydromt-sfincs-dev/lib/python3.9/site-packages/hydromt/gis_utils.py:70: in <module>
    GPD_TYPES = gpd.GeoDataFrame | gpd.GeoSeries
E   TypeError: unsupported operand type(s) for |: 'type' and 'type'
```
https://github.com/Deltares/hydromt_sfincs/actions/runs/6902656342/job/18779801237?pr=147

## Explanation
Adding `from __future__ import annotations` to gis_utils.py
@savente93 @Jaapel Any idea why this error did not appear in our py39 tests? It woud be good to update the tests also in this PR so that we identify these bugs earlier..

## Checklist

### DIFF
--- a/hydromt/gis_utils.py
+++ b/hydromt/gis_utils.py
@@ -2,6 +2,8 @@
 # -*- coding: utf-8 -*-
 
 """GIS related convenience functions."""
+from __future__ import annotations
+
 import glob
 import logging
 import os
--- a/hydromt/gis_utils.py
+++ b/hydromt/gis_utils.py
@@ -9,7 +9,7 @@ import logging
 import os
 from io import IOBase
 from os.path import dirname
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Union
 
 import geopandas as gpd
 import numpy as np
@@ -69,8 +69,8 @@ GDAL_DRIVER_CODE_MAP = {
 }
 GDAL_EXT_CODE_MAP = {v: k for k, v in GDAL_DRIVER_CODE_MAP.items()}
 
-GPD_TYPES = gpd.GeoDataFrame | gpd.GeoSeries
-GEOM_TYPES = GPD_TYPES | BaseGeometry
+GPD_TYPES = Union[gpd.GeoDataFrame, gpd.GeoSeries]
+GEOM_TYPES = Union[GPD_TYPES, BaseGeometry]
 
 ## GEOM functions
 
@@ -569,7 +569,7 @@ def to_geographic_bbox(bbox, source_crs):
 
 def bbox_from_file_and_filters(
     fn: IOBase,
-    bbox: GEOM_TYPES | None = None,
+    bbox: Union[GEOM_TYPES, None] = None,
     mask: GEOM_TYPES | None = None,
     crs: CRS | None = None,
 ) -> Tuple[float, float, float, float] | None:
--- a/make_env.py
+++ b/make_env.py
@@ -47,7 +47,9 @@ parser.add_argument("profile", default="dev,test", nargs="?")
 parser.add_argument("--output", "-o", default="environment.yml")
 parser.add_argument("--channels", "-c", default=None)
 parser.add_argument("--name", "-n", default=None)
-parser.add_argument("--py-version", "-p", default=None)
+parser.add_argument(
+    "--py-version", "-p", default="3.11", choices=["3.9", "3.10", "3.11"]
+)
 parser.add_argument("-r", "--release", action="store_true")
 args = parser.parse_args()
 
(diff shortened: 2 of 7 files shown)


savente93:
So it turns out that using `|` instead of `Union` wasn't added until python 3.9, which is what is causing the error. The reason we didn't catch this in our tests is that there was a bug in my `make_env.py` which means that all of our tests were simply running on python 3.11 without us knowing, because python version specifications are very confusing. I'll update the make_env so that shouldn't happen anymore but we might have to fix this in a bunch of places first unless we want to go with the `__future__` solution. What exactly `__future__` does 'm not sure about. I'm ambivilent about which one we should take.

pixi.toml
```diff
@@ -14,7 +14,7 @@ dev = { depends_on = [
   "dev-install-hydromt",
   "dev-install-pre-commit",
 ] }
-dev-env-file = "python make_env.py full -n hydromt-dev"
+dev-env-file = "python make_env.py full -n hydromt-dev -p 3.9"
```
DirkEilander, line 17: I think this change should not be committed right?

savente93 (approved):
Looks like the tests run again, I'll check the cache when this lands on main. Dirk, I'm happy with it like this, take a look and feel free to merge when you're also satisfied

make_env.py
```diff
@@ -89,7 +91,7 @@ def _parse_profile(profile_str: str, opt_deps: dict, project_name: str) -> List[
     else:
         conda_deps.append(dep)
 if args.py_version is not None:
-    conda_deps.append(f"python~={args.py_version}")
+    conda_deps.append(f"python~={args.py_version}.0")
```
DirkEilander, line 94: From SO I understand that e.g. `~=3.9.0` means `>=3.9.0, ==3.9.*` And before it went wrong because without the final `.0` it means `>=3.9, ==3.*`?

It seems less confusing to me to simply use `==3.9` which also gives you the latest 3.9 release right? 

Either way, as long as it works, I'm fine with it.

DirkEilander (commented):
Thanks for looking into this @savente93! this is an important fix of the tests. 
The change in the pixi.toml file should be undone I think, but otherwise my comments are minor.

make_env.py
DirkEilander: I think @dalmijn  is also using the make_env.py in the hydromt_wflow repos and it is also in the cookie cutter. It would be good to update it in all these places to avoid future bugs.

savente93: mostly correct I think. I remember looking into this before, and the reason we didn't use `==3.9` is because then you get `3.9.0` which does not include security updates. That's at least how I understand it, but honestly I find the whole thing very confusing. So I suggest we keep it like this for now as I've verified that this gives you the latest 3.9, but I'll look a bit more into it to see if I can understand it better

hydromt/gis_utils.py
```diff
@@ -2,12 +2,14 @@
 # -*- coding: utf-8 -*-
 
 """GIS related convenience functions."""
+from __future__ import annotations
+
```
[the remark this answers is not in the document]
savente93: For now it doesn't matter that much as it seems to be okay as is, but I think we should make a decision at some point on what style we want and enforce it consistently. Which one we choose will depend on how long we wish to keep supporting 3.9, as active support for it has already ended, though security support will last for about 2 more years (source: https://endoflife.date/python)

pixi.toml
savente93: I'd like to keep it there, I want to start working on 3.9 again so that I can catch things earlier and this is what I use to recreate my env.

PR #660, opened 2023-11-24
Export improvements

### LINKED ISSUE #657: Improve export CLI
### Kind of request

### Enhancement Description

I had a quick look at what was done in PR #627. Nice start!
I have a few suggestions or things to discuss before we also officially add this function to our docs in #644

First a bug:

- region argument is not passed to the export_data function (and should be converted to bbox first)

Then comments:

- check typos in the docstrings 
- in the docstrings, I feel using '.' for the mandatory path argument instead of path/to/output_dir is less clear. Also compared to other command line examples for build/update
- CLI args option name: -f is used for config file while in build/update we use -i. Do you think we should harmonize and use -i here as well?
- CLI args option name: here might be completely my taste but would -s instead of -t for source names be possible? -s sounds maybe closer to data source compared to target?
- add CLI args option: would it be possible to add an option in CLI for time tuple? Then in really most cases, you don't need a config file, unless for advanced options.
- support all arguments of the export data functions from the config file (unit_add, meta, append)
- nice-to-have: support multiple sources option (current -t) from CLI like we support multiple data catalogs option (-d) in build/update
- format of the config file: might be taste again but I feel that separating args and sources does not feel so clear to me.. could maybe all arguments be put on the same level and naming following a little closer the arguments of the export_data function? Also not sure why dictionary is necessary for each new source? (also data_libs is used instead of catalog in build/update and I am not sure what is the tuple syntax in yaml). My idea would be something like:

```
export:
  region: 
    - subbasin: [-7.24, 62.09]
    - uparea: 50
  data_libs:
    - tests/data/test_sources.yml
  time_tuple: (2010-01-01, 2020-12-31)
  source_names:
    - hydro_lakes[precip]
    - gtsmv3_eu_era5
  unit_conversion: False
  append: False
  meta: 
    version: 0.1
```

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #657 

## Explanation
- I ended up not allowing multiple instances of time tuples because that would end up either overwriting the previous exports, or changing the file names, which I didn't want to do. 
- Other changes are based on the discussion Hélène and I had. 

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,7 +13,7 @@ Unreleased
 
 Added
 -----
-- Export CLI now also accetps time tuples (#660)
+- Export CLI now also accepts time tuples (#660)
 
 Changed
 -------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -13,13 +13,16 @@ Unreleased
 
 Added
 -----
+- Export CLI now also accetps time tuples (#660)
 
 Changed
 -------
+- Export CLI now uses '-s' for source, '-t' for time and '-i' for config. (#660)
 
 Fixed
 -----
 - Bug in `raster.set_crs` if input_crs is of type CRS. (#659)
+- Export CLI now actually parses provided geoms. (#660)
 
 Deprecated
 ----------
--- a/docs/user_guide/data_overview.rst
+++ b/docs/user_guide/data_overview.rst
@@ -61,6 +61,23 @@ A special exception is made for the Deltares data catalog which can be accessed 
     data sources with the same names, the sources from the last catalog in the list are used.
     If the ``--dd (--deltares-data)`` flag is used the deltares_data catalog is read first.
 
+The CLI also provides two other utility subcommands: `export` and `check`. THe export function can be used
+to export a subset of the data avaialble in your data catalogs for e.g. collaboration. This sub command supports
+providing a region (bounding box or path to geom only), data catalogs, data catalgo entry names, and time ranges
+for example:
+
+.. code-block:: console
+
+    hydromt export -r "{'bbox': [4.6891,52.9750,4.9576,53.1994]}" -s era5_hourly -d ../hydromt/data/catalogs/artifact_data.yml -t '["2010-01-01", "2022-12-31"]' path/to/output_dir
+
+
+Finally the CLI also provides a lightweight command to check your files for compliance without having to build the model.
+Note that this does not yet validate the data specified, but only the syntax of the files themselfes.
+However, we do plan to expand it's functionality over time. It can be use for example like so:
+
+.. code-block:: console
+
+    hydromt check grid -d /path/to/data_catalog.yml -i /path/to/model_config.yml
 
 .. _get_data_python:
 
--- a/docs/user_guide/data_overview.rst
+++ b/docs/user_guide/data_overview.rst
@@ -61,9 +61,9 @@ A special exception is made for the Deltares data catalog which can be accessed 
     data sources with the same names, the sources from the last catalog in the list are used.
     If the ``--dd (--deltares-data)`` flag is used the deltares_data catalog is read first.
 
-The CLI also provides two other utility subcommands: `export` and `check`. THe export function can be used
-to export a subset of the data avaialble in your data catalogs for e.g. collaboration. This sub command supports
-providing a region (bounding box or path to geom only), data catalogs, data catalgo entry names, and time ranges
+The CLI also provides two other utility subcommands: `export` and `check`. The export function can be used
+to export a subset of the data available in your data catalogs for e.g. collaboration. This sub command supports
+providing a region (bounding box or path to geom only), data catalogs, data source names, and time ranges
 for example:
 
 .. code-block:: console
@@ -72,8 +72,8 @@ for example:
 
 
 Finally the CLI also provides a lightweight command to check your files for compliance without having to build the model.
-Note that this does not yet validate the data specified, but only the syntax of the files themselfes.
-However, we do plan to expand it's functionality over time. It can be use for example like so:
+Note that this does not yet validate the data specified, but only the syntax of the files themselves.
+However, we do plan to expand its functionality over time. It can be use for example like so:
 
 .. code-block:: console
 
--- a/hydromt/cli/main.py
+++ b/hydromt/cli/main.py
@@ -2,17 +2,18 @@
 """command line interface for hydromt models."""
 
 import logging
+from ast import literal_eval
+from datetime import datetime
 from json import loads as json_decode
 from os.path import join
 from pathlib import Path
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional
 
 import click
 import numpy as np
 from pydantic import ValidationError
 
 from hydromt.data_catalog import DataCatalog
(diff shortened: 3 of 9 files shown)


docs/changelog.rst
```diff
@@ -13,13 +13,16 @@ Unreleased
 
 Added
 -----
+- Export CLI now also accetps time tuples (#660)
```
hboisgon, line 16:
```suggestion
- Export CLI now also accepts time tuples (#660)
```

docs/user_guide/data_overview.rst
```diff
@@ -61,6 +61,23 @@ A special exception is made for the Deltares data catalog which can be accessed
     data sources with the same names, the sources from the last catalog in the list are used.
     If the ``--dd (--deltares-data)`` flag is used the deltares_data catalog is read first.
 
+The CLI also provides two other utility subcommands: `export` and `check`. THe export function can be used
```
hboisgon, line 64:
```suggestion
The CLI also provides two other utility subcommands: `export` and `check`. The export function can be used
```

docs/user_guide/data_overview.rst
```diff
@@ -61,6 +61,23 @@ A special exception is made for the Deltares data catalog which can be accessed
...
+to export a subset of the data avaialble in your data catalogs for e.g. collaboration. This sub command supports
```
hboisgon, line 65:
```suggestion
to export a subset of the data available in your data catalogs for e.g. collaboration. This sub command supports
```

docs/user_guide/data_overview.rst
```diff
@@ -61,6 +61,23 @@ A special exception is made for the Deltares data catalog which can be accessed
...
+providing a region (bounding box or path to geom only), data catalogs, data catalgo entry names, and time ranges
```
hboisgon, line 66:
```suggestion
providing a region (bounding box or path to geom only), data catalogs, data source names, and time ranges
```

docs/user_guide/data_overview.rst
```diff
@@ -61,6 +61,23 @@ A special exception is made for the Deltares data catalog which can be accessed
...
+for example:
+
+.. code-block:: console
+
+    hydromt export -r "{'bbox': [4.6891,52.9750,4.9576,53.1994]}" -s era5_hourly -d ../hydromt/data/catalogs/artifact_data.yml -t '["2010-01-01", "2022-12-31"]' path/to/output_dir
+
+
+Finally the CLI also provides a lightweight command to check your files for compliance without having to build the model.
+Note that this does not yet validate the data specified, but only the syntax of the files themselfes.
```
hboisgon, line 75:
```suggestion
Note that this does not yet validate the data specified, but only the syntax of the files themselves.
```

docs/user_guide/data_overview.rst
```diff
@@ -61,6 +61,23 @@ A special exception is made for the Deltares data catalog which can be accessed
...
+However, we do plan to expand it's functionality over time. It can be use for example like so:
```
hboisgon, line 76:
```suggestion
However, we do plan to expand its functionality over time. It can be use for example like so:
```

hydromt/cli/main.py
```diff
@@ -440,14 +440,15 @@ def export(
     Example usage:
     --------------
 
-    export the data of in a single source, in a pertcular region
-    hydromt export -r "{'subbasin': [-7.24, 62.09], 'uparea': 50}" -t era5_hourly -d ../hydromt/data/catalogs/artifact_data.yml .
+    export the data of in a single source, in a particular region, in a particular time range
```
hboisgon, line 443:
```suggestion
    export the data of a single data source, in a particular region, for a particular time range
```

hydromt/cli/main.py
```diff
@@ -440,14 +440,15 @@ def export(
...
+    hydromt export -r "{'bbox': [4.6891,52.9750,4.9576,53.1994]}" -s era5_hourly -d ../hydromt/data/catalogs/artifact_data.yml -t '["2010-01-01", "2022-12-31"]' path/to/output_dir
+    note to sam: only support BBOX and GEOM
```
hboisgon, line 445:
```suggestion

```

hydromt/cli/main.py
```diff
@@ -440,14 +440,15 @@ def export(
...
 
-    export all data of in a single source
-    hydromt export --dd -t era5_hourly .
+    export all data of in a single source of the deltares data catalog
```
hboisgon, line 447:
```suggestion
    export a single data source from the deltares data catalog without time/space slicing
```

hydromt/cli/main.py
```diff
@@ -440,14 +440,15 @@ def export(
...
+    hydromt export --dd -s era5_hourly path/to/output_dir
```
hboisgon, line 448: I think we want to deprecate the --dd flag:
```suggestion
    hydromt export -d deltares_data -s era5_hourly path/to/output_dir
```

hydromt/cli/main.py
```diff
@@ -464,31 +465,58 @@ def export(
...
+
+    if source:
+        if isinstance(source, str):
+            sources = [source]
+        else:
+            sources = list(source)
+
+    if config:
+        config_dict = cli_utils.parse_config(config)["export_data"]
+        args = config_dict.pop("args", {})
```
hboisgon, line 478: Is 'args' still in the export_data config? If not need to update below

hydromt/cli/main.py
```diff
@@ -464,31 +465,58 @@ def export(
...
         if "catalog" in args.keys():
```
hboisgon, line 479: I suggested data_libs in the issue instead of catalog in the template (not as a bullet point). I leave up to you to change but data_libs is used in the build and update command to specify catalogs from config file ([global] section of the config corresponding to Model.__init__ keys)

hydromt/cli/main.py
```diff
@@ -464,31 +465,58 @@ def export(
...
 
     try:
-        data_catalog = DataCatalog(data_libs=data_libs)
         data_catalog.export_data(
             export_dest_path,
-            source_names=export_targets,
-            time_tuple=time_tuple,
+            source_names=sources,
+            bbox=bbox,
+            time_tuple=time_tup,
```
hboisgon, line 519: I miss passing the other arguments from config_dict like unit_conversion etc. (especially if you pop the others, can easily added here as **config_dict)

hboisgon (changes requested):
Almost there @savente93, thanks for looking into it! 
I think the multiple -t options is actually implemented (it was support multiple sources which at the time were using -t instead of -s now :) )
What I miss is passing the extra arguments from config and for you to check if the config is properly parsed by your function still with the changes implemented

savente93:
> What I miss is passing the extra arguments from config and for you to check if the config is properly parsed by your function still with the changes implemented

Sorry I don't understand what you are asking here, could you rephrase it for me? other than that, all of the comments should be resolved now.

hydromt/cli/main.py
hboisgon: Oh sorry, what you did also works perfectly :) I meant pass here export_data(source_names = sources, ..., **config_dict) and instead of defining unit_conversion = True etc. just define config_dict = dict() in the case there is no config file provided.
But his also works perfectly well

PR #664, opened 2023-11-28
Enable simple region checking for cli

### LINKED ISSUE #638: Model config should be able to validate region arguments
### Kind of request

None

### Enhancement Description

The current implementation of the region validation is incomplete. Mainly, some options regarding subbasin id are missing. The code could also do with some refactoring as there is a lot of duplication at the moment. If possible it would be great if we could also validate the regions even ones that were loaded from files, but we'll have to see how realisitic this is. 

### Use case

I would like to use `hydromt check -r <region>` and have it tell me if the region is invalid. A nice bonus would be to tell if we, for example, end up with an empty region, but that is secondary. 

### Additional Context

for additional inspiration see: https://pypi.org/project/geojson-pydantic/

savente93:
## Issue addressed
Fixes #638 

## Explanation
Given that we only support bbox and geom regions in the export CLI, I thought it was appropriate to do the same here. We can add to this over time but then we will need a more concrete proposal for who it should look and what we should support so I think this is okay for now. Then at least people can use the most simple variants. 

## Checklist


savente93:
Things we want to still add:
- [ ] Add all options for drivers, even if we can't yet check that the data type and driver combo is correct
- [ ] Check that crs is a valid crs
- [ ] Check that the region kind is one that hydromt supports even if we cannot validate it yet, and give better errors there
- [ ] See what is up with the setup precip forcing on grid model not being caught

savente93:
@hboisgon I added your `setup_precip_forcing` as a test case and it passes here so I am not quite sure what is going on here. Perhaps this indicates that the cli is the problem and not the checking, I'll try and investigate that a bit further if I have some time left today or tomorrow, but the other points we discussed should now be fixed

savente93:
I figured out what the issue was with the grid model precip example Hélène found. The validator was actually only validating the first step in the file provided, not all of them. This is now rectified and it should all work. Very good catch @hboisgon !

Issue #537, opened 2023-09-29
Improve reproducibility of "deltares_data" catalog sources


DirkEilander:
### Enhancement Description

Updating the meta data section in deltares_data.yml & documentation according to:

```yaml
  meta:
    source_url: zenodo.org/my_dataset # should point to processed data OR original in combi with processing_notes/script
    source_license: CC-BY-3.0
    source_version: vX.X
    paper_ref: Author et al. (year)
    paper_doi: doi
    processing_notes:  <description of process in script OR simple processing steps (e.g. filter / gdalbuildvrt)>
    processing_script: <url to script>
    category: category
```

It should be checked case by case what is required for reproducibility. there are several options:
- publish pre-processed data together with the script on Zenodo (e.g. MODIS_LAI/ MERIT Hydro basins map) and point to this data in `source_url`
- point to scripts in `processing_script` to download and/or process (e.g. ERA5)
- add `processing_notes` for simple processing to filter data (e.g. hydro_lakes) or create a vrt (merit)
- documentation of required data (e.g. bounds is required in the hydrographic region argument unless the basin map and index are present)
- check used data sources in examples (e.g. replace merit_hydro with merit_hydro_ihu).

### ToDo
- [x] identify data sources in deltares_data.yml that need updating
- [x] update meta section
- [ ] if the dataset requires publishing / new scripts (more work) -> create separate issues
- [ ] update docs and examples (also in plugins)

### Additional Context

this is a continuation of #356

Tjalling-dejong:
The following datasets need to be updated:
- CHELSA 1.2 -> 2.1
- COPDEM 2021_1 -> 2022_1
- E-OBS 25.0e -> 28.0e
- Esa world cover v100 -> v200
- FABDEM v1.0 -> v1.2
- GADM 3.6 -> 4.1
- GEBCO 2020 -> 2023
- GHS_POP R2019A_v1.0 -> R2023A
- GHS_SMOD  R2016A_v1.0 -> R2023A
- GLOFAS v31? -> v4.0
- GLW v3 -> v4
- GSWO v1_1_2019 -> v2021
- Mdt_cnes_cls18 -> cls22
- MODIS LAI MCD15A3H V006- > MCD15A3H V061
- SM2RAIN_ASCAT v1.4 -> v2.1
- Worldclim 2.0 -> worldclim 2.1
- World settlement footprint 2015 -> 2019 version

Issue #521, opened 2023-09-21
Data catalog should be updated to take advantage of new features


savente93:
### Kind of request

Adding new functionality

### Enhancement Description

Hydromt has recently added support for a bunch of features in the data catalog that are currently not use. Most notably the use of variants and compatability checking. The data catalog should be brought up to date with these chagnes

savente93:
fixed in #667

PR #667, opened 2023-11-30
Improve reproducibility dd

### LINKED ISSUE #356: DOC: better document in the data catalogs if datasets were pre-processed
### HydroMT version checks

In the docs here: https://deltares.github.io/hydromt/latest/user_guide/data_existing_cat.html
Maybe also here (meta section) depending on implementation: https://deltares.github.io/hydromt/latest/user_guide/data_prepare_cat.html

### Documentation problem

Some of the datasets in the pre-defined catalogs are actually not original but pre-processed data (eg. modis_lai, merit_hydro for some of the layers etc).
Maybe we should find a standard way of letting the user know about this?
See also issue in [hydromt-wflow:#157](https://github.com/Deltares/hydromt_wflow/issues/157)

Known issues:

- merit_hydro
- merit_hydro_patch
- modis_lai
- era5 other than hourly (daily, zarr)
- chirps

Possibly related:
- hydro_lakes
- hydro_reservoirs

I think so far we tried to use source_url and notes in meta to say if processing was done. For some data sources it's missing but I also wonder if this way is clear to the user or if we should do it differently ?

For example only add source_url if no processing was done.
In case of processing, use new keywords processing_from_url, processing_from_doi, processing_steps?

Tjalling-dejong:
## Issue addressed
Fixes #356  #521 and partially #537

## Explanation
This PR updates the deltares data catalog to include new features such as versions/variants of data sources. Also if a data source has been pre-processed the path to the pre-processing script or the notes describing the pre-processing is included. 
In addition I added the temporal extent of the datasets with a temporal dimension.

## Checklist


savente93 (approved):
Nice work Tjalling. It would be nice if we can also add the spatial extent and also host the pre-processing scripts we now have locally (I assume on the P drive) somewhere but these can be separate issues I think, so we can merge this for now.

DirkEilander:
Thanks for this work @Tjalling-dejong! Unfortunately this PR got merged too soon I think seeing some comments on Teams. This change directly affects all users and we need to think about how to properly test it.

Tjalling-dejong:
I replicated the error with the soilgrids dataset that people have raised in the HydroMT chat. The error persists in earlier versions of the data catalog as well. What has changed with this PR is that I added versions to data catalog items. With this change the users are served the latest version of a dataset by default. Thus, the soilgrids dataset given to the users is the latest, namely soilgrids_v2.0. This version of the dataset also breaks in earlier versions of the data catalog.

Error can be replicated with `datacatalog.get_rasterdataset("soilgrids_v2.0")`.

Tjalling-dejong:
I have made an issue regarding the error with soilgrids_v2.0 dataset (#706).

DirkEilander:
Thanks for checking this issue @Tjalling-dejong 👍 

The issue was indeed not caused by this PR, but we do change the default dataset as you already mention. I think it would be good to:
1) have a small test that only runs locally to actually reads (a random slice of) all the datasets in the catalog
2) keep track and communicate for which datasets the default version changes when we publish. Perhaps we can also add a note in the meta data of the items like `updated_in_version: xx`
3) make it common practice to add the version of your dataset in your hydromt scripts / model builder yml files (for discussion in the next plugin meeting @alimeshgi)

Would it be possible to add point 1 to this PR? And let's discuss about point 2.

Issue #432, opened 2023-06-30
Support custom drivers to connect to data APIs


DirkEilander:
### Kind of request

Adding new functionality

### Enhancement Description

We want to enable users to access data directly from custom APIs through the data catalog.  
This is especially relevant for observational data that gets updated frequently, but can be used in a broader context.

### Use case

There are two known use cases in plugins

- HydroMT-Wflow does uses the global water watch api: https://github.com/Deltares/hydromt_wflow/blob/74812afe1ad163258c1afe63bfecdb6a14e7900d/hydromt_wflow/workflows/waterbodies.py#L256
- HydroMT-FIAT uses an api for NSI data: https://github.com/Deltares/hydromt_fiat/blob/fiat_integrator/hydromt_fiat/workflows/exposure_vector.py#L125

### Additional Context

My initial thoughts:
-	Downloading data can be supported through custom “driver” functions which can be made available to the data adaptors via the entrypoints (similar to plugins). We need to provide a template for these functions. The data catalog file can be used to pass possible kwargs to these driver functions similar to the current drivers.
-	The actual functions should live in plugins and/or a separate package to avoid having to maintain these and related dependencies in HydroMT.
-	I think it would be good practice to cache the downloaded data in a subfolder of the model root with a yml file for reproducibility.

DirkEilander:
@frederique-hub @roeldegoede @hboisgon Adding this issue based on our discussion yesterday. Let me know if this would work for you / if you have additional ideas.

@veenstrajelmer I understood that you may also have some thoughts on this?

veenstrajelmer:
Indeed, we would like to connect to data from e.g. the CMEMS ocean model via an API and other datasets via OPeNDAP. Would be nice to be able to access them via hydromt directly. I would like to contribute to a discussion, or alternatively can provide you with a textual overview, or both. Let me know what you prefer.

DirkEilander:
> Indeed, we would like to connect to data from e.g. the CMEMS ocean model via an API and other datasets via OPeNDAP. 

I think that OPeNDAP should already be possible as we use xarray.open_dataset for our netcdf driver. see https://docs.xarray.dev/en/stable/user-guide/io.html#opendap It would be good to verify this with a test.

veenstrajelmer:
[...]

PR #676, opened 2023-12-12
Replace Deprecated Entrypoints

### LINKED ISSUE #454: use of `entrypoints` should be deprecated
### HydroMT version checks

### Reproducible Example

https://pypi.org/project/entrypoints/

### Current behaviour

Currently HydroMT uses the `entrypoints` packages to advertise entry points for the plugins to use. However the PyPi page of `entrypoints` lists the current paragraph: 

> This package is in maintenance-only mode. New code should use the [importlib.metadata module](https://docs.python.org/3/library/importlib.metadata.html) in the Python standard library to find and load entry points.  

### Desired behaviour

Whenever possible we should prefer dependencies that are actively maintained, therefore we should investigate the possibility of moving away from `entrypoints` and use the new standard library functionality for this if possible

### Additional context

_No response_

Jaapel:
## Issue addressed
Fixes #454

## Explanation
Moved to importlib

## Checklist


Jaapel:
Nice to fix this already and tick off an issue. Will branch from here to work on https://github.com/Deltares/hydromt/issues/432.

hydromt/models/model_plugins.py
```diff
@@ -156,8 +162,8 @@ def __str__(self):
         """Generate string representation containing the registered entrypoints."""
         plugins = "".join(
             [
-                f" - {name} ({self.eps[name].distro.name}"
-                f" {self.eps[name].distro.version})\n"
+                f" - {name} ({self.eps[name].dist.name}"
+                f" {self.eps[name].dist.version})\n"
```
DirkEilander, line 166: found and fixed a small issue here. The new EntryPoint class does not have a distro, but dist property. To test these lines I had to make some changes in the test as well as when loading the general hydromt model classes as entry points.

DirkEilander (approved):

DirkEilander:
nice job @Jaapel !

PR #684, opened 2023-12-14
Write geojson file with WGS84 coordinates in Model.write_geoms()

### LINKED ISSUE #510: GeoJSON files should have WGS84 CRS (2016 specs)
### HydroMT version checks

### Current behaviour

With the standard `GeoDataFrame.to_file(driver='GeoJSON')` a crs key is added to the file which can be different from WGS84. This is not compatible with (or at least not preferred by) the [2016 geojson specs](https://datatracker.ietf.org/doc/html/rfc7946):

> The coordinate reference system for all GeoJSON coordinates is a geographic coordinate reference system, using the World Geodetic System 1984 (WGS 84) [WGS84] datum. [.....] However, where all involved parties have a prior arrangement, alternative coordinate reference systems can be used without risk of data being misinterpreted.

@aweerts found that the crs key in GeoJSON is not supported by Delft-FEWS

### Desired behaviour

To be sure the written files are compatible with all other software (e.g., Delft-FEWS), we could enforce writing these in WGS84 only  in the `Model.write_geoms`. This can be implemented with `GeoDataFrame.to_file(driver='GeoJSON', to_wgs84=True)`. When reading these files in `read_geoms` and then setting these in `set_geoms`, we should then by default reproject the geom to the model CRS to make sure internally the CRS are matching.  Note that this could create a loop as the crs might be based on geoms.

Note that plugins which implement their own write_geoms (e.g., HydroMT-SFINCS) would need to do the same in the write_geoms method.

### Additional context
The [GDAL GeoJSON driver](https://gdal.org/drivers/vector/geojson.html) implements the alternative [v1 specs](https://geojson.org/geojson-spec.html) which does allow for a different crs

Tjalling-dejong:
GeoJSON should be written with WGS84 coordinates according to the [2016 GeoJSON specs](https://datatracker.ietf.org/doc/html/rfc7946). Model.write_geoms() writes the geometries in model CRS normally. I added a option so that write_geoms can write GeoJSONs with WGS84 coordinates.

## Issue addressed
Fixes #510 

## Explanation
Added an to_wgs84 option in Model.write_geoms(). If set to True and the driver is GeoJSON the geometry will be set to EPSG:4326.

Also added a check in set_geoms so that new geometries are set to the model crs if the model has a crs. 

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -27,6 +27,7 @@ Fixed
 - Bug in `raster.set_crs` if input_crs is of type CRS. (#659)
 - Export CLI now actually parses provided geoms. (#660)
 - Bug in stats.skills for computation of pbias and MSE / RMSE. (#666)
+- `Model.write_geoms` now enforces GeoJSON to write coordinates in WGS84 if specified (#510)
 
 Deprecated
 ----------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -27,7 +27,7 @@ Fixed
 - Bug in `raster.set_crs` if input_crs is of type CRS. (#659)
 - Export CLI now actually parses provided geoms. (#660)
 - Bug in stats.skills for computation of pbias and MSE / RMSE. (#666)
-- `Model.write_geoms` now enforces GeoJSON to write coordinates in WGS84 if specified (#510)
+- `Model.write_geoms` ow has an option to write GeoJSON coordinates in WGS84 if specified (#510)
 
 Deprecated
 ----------
--- a/hydromt/models/model_api.py
+++ b/hydromt/models/model_api.py
@@ -1259,6 +1259,8 @@ class Model(object, metaclass=ABCMeta):
         fn : str, optional
             filename relative to model root and should contain a {name} placeholder,
             by default 'geoms/{name}.geojson'
+        to_wgs84: bool, optional
+            Option to enforce writing GeoJSONs with WGS84(EPSG:4326) coordinates.
         \**kwargs:
             Additional keyword arguments that are passed to the
             `geopandas.to_file` function.
--- a/hydromt/models/model_api.py
+++ b/hydromt/models/model_api.py
@@ -1220,6 +1220,10 @@ class Model(object, metaclass=ABCMeta):
             )
         if name in self.geoms:  # trigger init / read
             self.logger.warning(f"Replacing geom: {name}")
+        if hasattr(self, "crs"):
+            # Verify if a geom is set to model crs and if not sets geom to model crs
+            if self.crs.to_epsg() != geom.crs.to_epsg():
+                geom.to_crs(self.crs.to_epsg(), inplace=True)
         self._geoms[name] = geom
 
     def read_geoms(self, fn: str = "geoms/*.geojson", **kwargs) -> None:
@@ -1243,7 +1247,9 @@ class Model(object, metaclass=ABCMeta):
             self.logger.debug(f"Reading model file {name}.")
             self.set_geoms(gpd.read_file(fn, **kwargs), name=name)
 
-    def write_geoms(self, fn: str = "geoms/{name}.geojson", **kwargs) -> None:
+    def write_geoms(
+        self, fn: str = "geoms/{name}.geojson", to_wgs84: bool = False, **kwargs
+    ) -> None:
         r"""Write model geometries to a vector file (by default GeoJSON) at <root>/<fn>.
 
         key-word arguments are passed to :py:meth:`geopandas.GeoDataFrame.to_file`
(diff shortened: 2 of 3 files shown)


hydromt/models/model_api.py
```diff
@@ -1273,6 +1281,8 @@ def write_geoms(self, fn: str = "geoms/{name}.geojson", **kwargs) -> None:
             _fn = join(self.root, fn.format(name=name))
             if not isdir(dirname(_fn)):
                 os.makedirs(dirname(_fn))
+            if kwargs.get("driver") == "GeoJSON" and to_wgs84:
```
DirkEilander, line 1284:
```suggestion
            if to_wgs84 and (kwargs.get("driver") == "GeoJSON" or str(fn).lower().endswith(".geojson")):
```

hydromt/models/model_api.py
```diff
@@ -1273,6 +1281,8 @@ def write_geoms(self, fn: str = "geoms/{name}.geojson", **kwargs) -> None:
...
+                gdf.to_crs(4326, inplace=True)
```
DirkEilander, line 1285: Avoid that the original gdf in self.geoms get modified
```suggestion
                gdf = gdf.to_crs(4326)
```

hydromt/models/model_api.py
DirkEilander: I would be good to also remove lines 1272-73. Geopandas itself infers the driver based on the file path extension, see https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.to_file.html

docs/changelog.rst
```diff
@@ -27,6 +27,7 @@ Fixed
 - Bug in `raster.set_crs` if input_crs is of type CRS. (#659)
 - Export CLI now actually parses provided geoms. (#660)
 - Bug in stats.skills for computation of pbias and MSE / RMSE. (#666)
+- `Model.write_geoms` now enforces GeoJSON to write coordinates in WGS84 if specified (#510)
```
DirkEilander, line 30: I suggest to rephrase to make clear it is an option.
```suggestion
- `Model.write_geoms` now has an option to write GeoJSON coordinates in WGS84 if specified (#510)
```

tests/test_model.py
```diff
@@ -370,6 +370,40 @@ def test_setup_region(model, demda, tmpdir):
...
+    model.data_catalog.from_dict(
+        {
+            "region": {
+                "path": fn_region,
+                "data_type": "GeoDataFrame",
+                "driver": "vector",
+            }
+        }
+    )
+    model.setup_region({"geom": "region"})
```
DirkEilander, line 389: this should also work:
```suggestion
    model.setup_region({"geom": geom})
```

or even more simplified:
```py    
    model.set_geoms({"region": geom})
```

tests/test_model.py
```diff
@@ -370,6 +370,40 @@ def test_setup_region(model, demda, tmpdir):
...
+    model.write_geoms(to_wgs84=True)
+    region_geom = gpd.read_file(str(join(tmpdir, "geoms/region.geojson")))
+    assert region_geom.crs.to_epsg() == 4326
+
+
+def test_model_set_geoms(tmpdir):
```
DirkEilander, line 395: not sure what this method is testing? can you add some comments to explain?

DirkEilander (changes requested):
thanks for picking this up @Tjalling-dejong I've made some suggestions. Let me know if you have questions.

tests/test_model.py
Tjalling-dejong: Its testing if set_geoms converts the geom to the model crs if the geom crs is not the same as the model crs.

tests/test_model.py
Tjalling-dejong: I changed it to `model.set_geoms(geom=geom, name="region")`

tests/test_model.py
```diff
@@ -370,6 +370,29 @@ def test_setup_region(model, demda, tmpdir):
...
+    bbox = box(*[4.221067, 51.949474, 4.471006, 52.073727])
+    geom = gpd.GeoDataFrame(geometry=[bbox], crs=4326)
+    geom_28992 = geom.to_crs(epsg=28992)
+    # fn_region = str(join(tmpdir, "region.gpkg"))
+    # geom_28992.to_file(fn_region, driver="GPKG")
+    model = Model(root=str(tmpdir), mode="w")
+    model.setup_region({"geom": geom_28992})
+    model.set_geoms(geom, "geom_wgs84")
```
DirkEilander, line 392:
```suggestion
    model = Model(root=str(tmpdir), mode="w")
    model.setup_region({"geom": geom_28992})  # set model crs based on epsg28992
    model.set_geoms(geom, "geom_wgs84")  # this should convert the geom crs to epsg28992
```

tests/test_model.py
DirkEilander: check. I've added some comments to clarify.

DirkEilander (approved):
Thanks @Tjalling-dejong, LGTM!

PR #689, opened 2023-12-19
Updated deltares data catalog with new GADM fgb files

Tjalling-dejong:
## Issue addressed
Fixes #686

## Explanation
Explain how you addressed the bug/feature request, what choices you made and why.

## Checklist
- [ ] Updated tests or added new tests
- [ ] Branch is up to date with `main`
- [ ] Tests & pre-commit hooks pass
- [ ] Updated documentation if needed
- [ ] Updated changelog.rst if needed

## Additional Notes (optional)
Add any additional notes or information that may be helpful.

PR #690, opened 2023-12-19
Added csv version of grdc

Tjalling-dejong:
## Explanation
Added csv version of GRDC stations so that the optional openpyxl dependency does not have to be used.

## Checklist

### DIFF
--- a/data/catalogs/deltares_data.yml
+++ b/data/catalogs/deltares_data.yml
@@ -1,7 +1,7 @@
 ---
 meta:
   root: p:/wflow_global/hydromt
-  version: 2023.3
+  version: 2023.12
 
 
 basin_atlas_level12_v10:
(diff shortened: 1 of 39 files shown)


data/catalogs/deltares_data.yml
```diff
@@ -1052,6 +1052,21 @@ grdc:
   rename:
     area: uparea
 
+grdc_csv:
```
JoostBuitink, line 1055: Perhaps better to name this one `grdc` and the excel version `grdc_excel`. If no issues arise (I don't expect any, but just to be sure), we can remove the excel version, but perhaps safer to keep this one for now as a fallback option.

JoostBuitink (changes requested):
Thanks for the quick changes! I have one comment regarding the name the dataset. I would opt to replace the csv version with the excel version (to prevent people from getting the openpyxl dependency error).

data/catalogs/deltares_data.yml
Tjalling-dejong: Yes that is better option. I might as well remove the excel version since the column info is also available at the source url.

DirkEilander:
@Tjalling-dejong  see also my comment in #689

JoostBuitink (approved):
LGTM!

DirkEilander (changes requested):
Generally looks good, but I'd like to see two changes:
- the changes should be logged in data/catalogs/readme.rst rather than docs/changelog.rst
- the version of the deltares_data.yml needs to be updated at the top of the file and registered in data/predefined_catalogs.yml

savente93 (approved):
LGTM but as @DirkEilander requested changes I think he'll haveto give the final OK

DirkEilander (approved):
LGTM!

PR #738, opened 2024-01-16
Handle nodata strategy in adapters more consistently

### LINKED ISSUE #731: Inconsistent handling of handle_nodata
### HydroMT version checks

### Reproducible Example

```python
import hydromt
from hydromt.nodata import NoDataStrategy

data_catalog = hydromt.DataCatalog("deltares_data")

ds = data_catalog.get_geodataframe(
    "hydro_lakes",
    bbox=[12.5, 46.5, 12.7, 46.7],
    predict="intersects",
    handle_nodata=NoDataStrategy.IGNORE,
)
```

### Current behaviour

When requesting data from the data_catalog (using the name), and when passing a NoDataStrategy, still an error is thrown, rather than a warming.

### Desired behaviour

Throw a warning in stead of an error. In the `.get_data()` functions (e.g., https://github.com/Deltares/hydromt/blob/main/hydromt/data_catalog.py#L1480), the `handle_nodata` argument should get passed as well. 

### Additional context

_No response_

savente93:
## Issue addressed
Fixes #731 

## Explanation
the no data strategy wasn't being passed or handled everywhere. this PR fixes that. 

## Checklist

### DIFF
--- a/hydromt/data_adapter/geodataset.py
+++ b/hydromt/data_adapter/geodataset.py
@@ -303,7 +303,7 @@ class GeoDatasetAdapter(DataAdapter):
             if handle_nodata == NoDataStrategy.RAISE:
                 raise NoDataException(f"No data was read from files {fns}")
             elif handle_nodata == NoDataStrategy.IGNORE:
-                logger.warn(f"no data was found at {fns}. skipping...")
+                logger.warning(f"no data was found at {fns}. skipping...")
 
         return ds
 
@@ -425,7 +425,7 @@ class GeoDatasetAdapter(DataAdapter):
                     "No data within spatial domain.", handle_nodata, logger
                 )
             elif handle_nodata == NoDataStrategy.IGNORE:
-                logger.warn("No data within spacial domain for")
+                logger.warning("No data within spacial domain. skipping...")
         return ds
 
     def _shift_time(self, ds, logger=logger):
@@ -451,7 +451,7 @@ class GeoDatasetAdapter(DataAdapter):
                         logger=logger,
                     )
                 elif handle_nodata == NoDataStrategy.IGNORE:
-                    logger.warn("GeoDataset: Time slice out of range.")
+                    logger.warning("GeoDataset: Time slice out of range.")
         return ds
 
     def _set_metadata(self, ds):
--- a/hydromt/data_adapter/geodataset.py
+++ b/hydromt/data_adapter/geodataset.py
@@ -15,6 +15,7 @@ from pystac import Catalog as StacCatalog
 from pystac import Item as StacItem
 from pystac import MediaType
 
+from hydromt.exceptions import NoDataException
 from hydromt.typing import ErrorHandleMethod, GeoDatasetSource, TimeRange, TotalBounds
 
 from .. import gis_utils, io
@@ -165,6 +166,7 @@ class GeoDatasetAdapter(DataAdapter):
         time_tuple=None,
         variables=None,
         driver=None,
+        handle_nodata=NoDataStrategy.RAISE,
         logger=logger,
         **kwargs,
     ):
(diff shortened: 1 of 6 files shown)


hydromt/data_adapter/geodataset.py
```diff
@@ -400,7 +420,12 @@ def _slice_spatial_dimension(
         logger.debug(f"Clip {predicate} [{bbox_str}] (EPSG:{epsg})")
         ds = ds.vector.clip_geom(geom, predicate=predicate)
         if ds.vector.index.size == 0:
-            _exec_nodata_strat("No data within spatial domain.", handle_nodata, logger)
+            if handle_nodata == NoDataStrategy.RAISE:
+                _exec_nodata_strat(
+                    "No data within spatial domain.", handle_nodata, logger
+                )
+            elif handle_nodata == NoDataStrategy.IGNORE:
+                logger.warn("No data within spacial domain for")
```
JoostBuitink, line 428: Typo (spacial vs spatial). Also, this warning message feels incomplete due to the "for". Where you planning to add any information here? Otherwise just add something like "No data within spatial domain. skipping..." (similar to above)

hydromt/data_adapter/geodataset.py
```diff
@@ -285,6 +299,12 @@ def _read_data(self, fns, logger=logger):
         else:
             raise ValueError(f"GeoDataset: Driver {self.driver} unknown")
 
+        if len(ds) == 0:
+            if handle_nodata == NoDataStrategy.RAISE:
+                raise NoDataException(f"No data was read from files {fns}")
+            elif handle_nodata == NoDataStrategy.IGNORE:
+                logger.warn(f"no data was found at {fns}. skipping...")
```
JoostBuitink, line 306: Just a general question, I thought it was better to use `logger.warning` in stead of `logger.warn`. Is there a difference or preference here?

tests/test_data_adapter.py
```diff
@@ -267,6 +284,22 @@ def test_geodataset(geoda, geodf, ts, tmpdir):
...
+
+    with pytest.raises(NoDataException):
+        da3 = data_catalog.get_geodataset(
+            "test.nc",
+            # only really care that the bbox doesn't intersect with anythign
+            bbox=[12.5, 12.6, 12.7, 12.8],
+            handle_nodata=NoDataStrategy.RAISE,
+        )
+
+    assert len(da3) == 0
```
JoostBuitink, line 302: Nice to see some tests added to this as well! Just for this line, it makes more sense to move it up (to 293), to keep it closer to code where the test belongs to.

JoostBuitink (commented):
Looks good to me, thanks for picking this up! Just a couple of small, general comments; let me know if anything is not clear!

JoostBuitink (approved):
Just one additional comment (that I forgot to highlight in the previous review), but other than that: looks good to me!

hboisgon:
@savente93 is there a reason you added ``handle_nodata`` only to ``geodataset`` and ``rasterdataset`` adapters and not the others (``dataset`` and ``dataframe``, ``geodataframe`` was already done before)? I see you added the argument to the general abstract method of the driver.

And for the ``to_file``, you only added it to ``geodataset`` but not the others?

Additionally but this can be a new issue, the default behavior is raise and for the export data catalog method, I would like it to be a warning/ignore (case: I would like to export data from deltares_data catalog but my region may not contain lake, glacier etc).

savente93:
The reason that I didn't add it to dataset and data frame is that those don't have spatial or temporal dimensions. I could do it for the reading of the files alone that's true. The reason I didn't add it to the `to_file` implementations is that I simply forgot. I'll make an issue for adjusting the default, and I'll add it to the other places at the same time. Do we want ignore to be the default everywhere?

PR #741, opened 2024-01-18
Added hydromt version to predefined data catalogs 

### LINKED ISSUE #708: Add hydromt_version in predefined data catalog files
### HydroMT version checks

### Reproducible Example

Apart from the `artifact_data` catalog, all predefined data catalogs do not yet use the `hydromt_version` meta key. This should be added to raise better errors when loading unsupported datasets.

This is not documented, but already supported in the code, see:
https://github.com/Deltares/hydromt/blob/074020ef122eada11bcb42219d554e266277d5cd/hydromt/data_catalog.py#L870

### Current behaviour

when loading deltares_data with hydromt v0.8 you get a random ValueError

```
hydromt.DataCatalog('deltares_data')

>> ValueError: pcr_globwb: Missing required path argument
```

### Desired behaviour

```
RunTimeError: Data catalog requires Hydromt Version ...
```

### Additional context

_No response_

Tjalling-dejong:
## Issue addressed
Fixes #708

## Explanation
Added HydroMT version to predefined data catalogs

## Checklist

### DIFF
--- a/data/catalogs/aws_data.yml
+++ b/data/catalogs/aws_data.yml
@@ -1,6 +1,8 @@
 ---
 meta:
-  version: 2023.2
+  version: 2024.1
+  hydromt_version: '>=0.9'
+
 esa_worldcover_2020_v100:
   crs: 4326
   data_type: RasterDataset
--- a/data/catalogs/aws_data.yml
+++ b/data/catalogs/aws_data.yml
@@ -1,7 +1,8 @@
 ---
 meta:
   version: 2024.1
-  hydromt_version: '<1.0'
+  hydromt_version: '>=0.9, <1.0'
+
 
 esa_worldcover_2020_v100:
   crs: 4326
--- a/data/catalogs/aws_data.yml
+++ b/data/catalogs/aws_data.yml
@@ -1,7 +1,7 @@
 ---
 meta:
   version: 2024.1
-  hydromt_version: '>=0.9'
+  hydromt_version: '<1.0'
 
 esa_worldcover_2020_v100:
   crs: 4326
--- a/data/catalogs/aws_data.yml
+++ b/data/catalogs/aws_data.yml
@@ -1,6 +1,6 @@
 ---
 meta:
-  version: 2024.1
+  version: 2024.1.30
   hydromt_version: '>=0.9, <1.0'
 
 
--- a/.github/workflows/check-data-catalogs.yml
+++ b/.github/workflows/check-data-catalogs.yml
@@ -28,6 +28,7 @@ jobs:
       cancel-in-progress: true
     steps:
 
+      - uses: actions/checkout@v3
       # we need environment.yml to see if we have a cache hit
       - name: Generate env spec
         run: pip install tomli && python make_env.py test,io,extra,doc -p ${{ env.DOC_PYTHON_VERSION }} -n hydromt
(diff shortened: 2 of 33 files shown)


Tjalling-dejong:
@savente93 Unfortunately the validate data catalog workflow is breaking :(

savente93:
will have a look

data/catalogs/aws_data.yml
```diff
@@ -1,6 +1,8 @@
 ---
 meta:
-  version: 2023.2
+  version: 2024.1
+  hydromt_version: '>=0.9'
```
savente93, line 4: So I'm not sure what version spec we want here. because if we release a new version of the catalog then this will still try to read that, but I'm also not sure how we should bound this version from above. We could add a `<0.10` for example, but it's not necessarifly true that 0.10 whill have braking changes for the catalog. @hboisgon what do you think?

savente93 (commented):
I think we need to decide which version we want to specifiy. For furute correctness we also need to bound it from above but I'm not sure what version we want to bound it at. Perhaps we can discuss this amoung the team

DirkEilander:
As upper bound we can use '<1.0' I think. I don't expect we will have breaking changes before and we can release a new version with v1

Tjalling-dejong:
@DirkEilander's comment closes the discussion right? Shall we merge it? @savente93

savente93:
the upper bound still needs to be added

savente93:
You'll need both an upper and a lower bound. As specified, currently any version of hydromt matches this version specifier

savente93 (changes requested):
You'll need both an upper and a lower bound. As specified, currently any version of hydromt matches this version specifier

DirkEilander (changes requested):
- The version numbers in the aws_data and gcs_cmip6_data do not match with the version numbers in the predefined_catalog.yml and changelog. Could this be checked automatically?
- Perhaps for another issue, but instead of YY.M.D versioning I would also prefer YY.MINOR.MICRO versioning (used by e.g. pip). With the YY.M.D it is impossible for a user to know or what the previous was. Furthermore our versioning is not time sensitive to the day, adding days doesn't really make sense to me.  
- Currently we also don't document the versions anywhere outside of the changelog. It would be good if this is visible in the docs, e.g. a link to the changelog [here](https://deltares.github.io/hydromt/latest/user_guide/data_existing_cat.html) would be a good start.

Tjalling-dejong:
@DirkEilander with YY.M.D it is clear what the previous version is right? What do you mean?
I am adding days because there are multiple PRs this month making changes to the data catalog. 
We can of course do year and month but then the data catalog version can contain multiple changes done over the month.

DirkEilander:
> @DirkEilander with YY.M.D it is clear what the previous version is right? What do you mean? I am adding days because there are multiple PRs this month making changes to the data catalog. We can of course do year and month but then the data catalog version can contain multiple changes done over the month.

Correct, this issue was already there. Now that we start adding days I realize that this type of versioning will be very difficult for our users:  it is impossible to know based on the current version what the previous version was and also we don't distinguish between MAJOR (changing a default version or source) and MINOR changes (adding new sources or changing the version number). I will make a new issue to propose YY.MAJOR.MINOR versioning. I guess it might be best to change this with V1.

DirkEilander (approved):
LGTM!

savente93 (approved):
LGTM

PR #743, opened 2024-01-19
Export no data

### LINKED ISSUE #739: CLI users should have some control over the nodata strategy used
### Kind of request

Adding new functionality

### Enhancement Description

in #738 I forgot to add the nodata strategy handeling to a couple of places, (e.g. the `to_file` functions and in general in the dataframe and dataset adapters) so that should be taken care of. Secondly, we should have a good discussion about what we want the default to be. I would advocate for having one default everywhere for consistency, but for example cli export users do need some control over this. therefor I have come up with the following proposal:
1. Raise is the default
1. add a cli option to change the default behaviour to ignore
1. options in the config should override basically everything else

Additionally the use of the nodata strategy should be tested from the configuration files as well .

### Use case

_No response_

### Additional Context

(sorry for cramming multiple issues in one ticket, but they are all related and fairly small, so I think it's okay)

savente93:
## Issue addressed
Fixes #739 

## Explanation

## Checklist

### DIFF
--- a/hydromt/cli/main.py
+++ b/hydromt/cli/main.py
@@ -15,6 +15,7 @@ from geopandas import GeoDataFrame
 from pydantic import ValidationError
 
 from hydromt.data_catalog import DataCatalog
+from hydromt.nodata import NoDataStrategy
 from hydromt.validators.data_catalog import DataCatalogValidator
 from hydromt.validators.model_config import HydromtModelSetup
 from hydromt.validators.region import validate_region
@@ -117,6 +118,12 @@ overwrite_opt = click.option(
     is_flag=True,
     default=False,
     help="Flag: If provided overwrite existing model files",
+)
+ignore_empty_opt = click.option(
+    "--ignore-empty",
+    is_flag=True,
+    default=False,
+    help="Flag: Simply ignore empty data sets when exporting",
 )
 
 cache_opt = click.option(
@@ -444,6 +451,7 @@ def check(
 @data_opt
 @deltares_data_opt
 @overwrite_opt
+@ignore_empty_opt
 @quiet_opt
 @verbose_opt
 @click.pass_context
@@ -457,6 +465,7 @@ def export(
     data: Optional[List[Path]],
     dd: bool,
     fo: bool,
+    ignore_empty: bool,
     quiet: int,
     verbose: int,
 ):
@@ -480,6 +489,11 @@ def export(
         "export", join(export_dest_path, "hydromt.log"), log_level=log_level
     )
     logger.info(f"Output dir: {export_dest_path}")
+
+    if ignore_empty:
+        handle_nodata = NoDataStrategy.IGNORE
+    else:
+        handle_nodata = NoDataStrategy.RAISE
 
     if data:
         data_libs = list(data)  # add data catalogs from cli
@@ -557,6 +571,7 @@ def export(
             unit_conversion=unit_conversion,
             meta=meta,
             append=append,
+            handle_nodata=handle_nodata,
         )
 
     except Exception as e:
--- a/hydromt/data_adapter/dataframe.py
+++ b/hydromt/data_adapter/dataframe.py
@@ -11,9 +11,10 @@ from pystac import Asset as StacAsset
 from pystac import Catalog as StacCatalog
 from pystac import Item as StacItem
 
+from hydromt.nodata import NoDataException, NoDataStrategy, _exec_nodata_strat
 from hydromt.typing import ErrorHandleMethod, StrPath, TimeRange, Variables
+from hydromt.utils import has_no_data
 
-from ..nodata import NoDataStrategy, _exec_nodata_strat
 from .data_adapter import DataAdapter
 
 logger = getLogger(__name__)
@@ -177,34 +178,37 @@ class DataFrameAdapter(DataAdapter):
 
 
         """
-        kwargs.pop("bbox", None)
-        obj = self.get_data(time_tuple=time_tuple, variables=variables, logger=logger)
-
-        if len(obj) == 0:
-            _exec_nodata_strat(
-                f"no data to export for {self.name}.",
-                strategy=handle_nodata,
-                logger=logger,
+        try:
+            kwargs.pop("bbox", None)
+            obj = self.get_data(
+                time_tuple=time_tuple, variables=variables, logger=logger
             )
-            return
 
-        read_kwargs = dict()
-        if driver is None or driver == "csv":
-            # always write as CSV
-            driver = "csv"
-            fn_out = join(data_root, f"{data_name}.csv")
-            obj.to_csv(fn_out, **kwargs)
-            read_kwargs["index_col"] = 0
-        elif driver == "parquet":
-            fn_out = join(data_root, f"{data_name}.parquet")
-            obj.to_parquet(fn_out, **kwargs)
-        elif driver == "excel":
-            fn_out = join(data_root, f"{data_name}.xlsx")
-            obj.to_excel(fn_out, **kwargs)
-        else:
-            raise ValueError(f"DataFrame: Driver {driver} is unknown.")
+            if len(obj) == 0:
+                raise NoDataException()
 
-        return fn_out, driver, read_kwargs
+            read_kwargs = dict()
+            if driver is None or driver == "csv":
+                # always write as CSV
+                driver = "csv"
+                fn_out = join(data_root, f"{data_name}.csv")
+                obj.to_csv(fn_out, **kwargs)
+                read_kwargs["index_col"] = 0
+            elif driver == "parquet":
+                fn_out = join(data_root, f"{data_name}.parquet")
+                obj.to_parquet(fn_out, **kwargs)
+            elif driver == "excel":
+                fn_out = join(data_root, f"{data_name}.xlsx")
+                obj.to_excel(fn_out, **kwargs)
+            else:
+                raise ValueError(f"DataFrame: Driver {driver} is unknown.")
+
+            return fn_out, driver, read_kwargs
+        except NoDataException:
+            _exec_nodata_strat(
+                "No data to export", strategy=handle_nodata, logger=logger
+            )
+            return None
 
     def get_data(
         self,
@@ -212,43 +216,48 @@ class DataFrameAdapter(DataAdapter):
         time_tuple: Optional[TimeRange] = None,
         handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
         logger: Logger = logger,
-    ) -> pd.DataFrame:
+    ) -> Optional[pd.DataFrame]:
         """Return a DataFrame.
 
         Returned data is optionally sliced by time and variables,
         based on the properties of this DataFrameAdapter. For a detailed
         description see: :py:func:`~hydromt.data_catalog.DataCatalog.get_dataframe`
         """
-        # load data
-        fns = self._resolve_paths(variables=variables)
-        df = self._read_data(fns, logger=logger)
-        if len(df) == 0:
-            _exec_nodata_strat(
-                "No data was read from source", strategy=handle_nodata, logger=logger
+        try:
+            # load data
+            fns = self._resolve_paths(variables=variables)
+            df = self._read_data(fns, handle_nodata=NoDataStrategy.RAISE, logger=logger)
+            if has_no_data(df):
+                raise NoDataException()
+            # rename variables and parse nodata
+            df = self._rename_vars(df)
+            df = self._set_nodata(df)
+            # slice data
+            df = DataFrameAdapter._slice_data(
+                df,
+                variables,
+                time_tuple,
+                handle_nodata=NoDataStrategy.RAISE,
+                logger=logger,
             )
+
+            # only if data is actually returned
+            self.mark_as_used()  # mark used
+
+            # uniformize data
+            df = self._apply_unit_conversion(df, logger=logger)
+            df = self._set_metadata(df)
             return df
-        # rename variables and parse nodata
-        df = self._rename_vars(df)
-        df = self._set_nodata(df)
-        # slice data
-        df = DataFrameAdapter._slice_data(df, variables, time_tuple, logger=logger)
-        if len(df) == 0:
+        except NoDataException:
             _exec_nodata_strat(
-                "No data was left after slicing", strategy=handle_nodata, logger=logger
+                "No data to export", strategy=handle_nodata, logger=logger
             )
-            return df
-
-        # only if data is actually returned
-        self.mark_as_used()  # mark used
-
-        # uniformize data
-        df = self._apply_unit_conversion(df, logger=logger)
-        df = self._set_metadata(df)
-        return df
+            return None
 
     def _read_data(
         self,
         fns: List[StrPath],
+        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
         logger: Logger = logger,
     ) -> pd.DataFrame:
         if len(fns) > 1:
@@ -298,6 +307,7 @@ class DataFrameAdapter(DataAdapter):
         df: pd.DataFrame,
         variables: Optional[Variables] = None,
         time_tuple: Optional[TimeRange] = None,
+        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
         logger: Logger = logger,
     ) -> pd.DataFrame:
         """Return a sliced DataFrame.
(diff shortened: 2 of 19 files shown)


savente93:
Opening this up for review as it's the seg fault that we can't do much about that is causing the tests to fail, and the docs will be taken care of in a seperate issue .

hydromt/cli/main.py
```diff
@@ -118,6 +119,12 @@ def print_models(ctx, param, value):
     default=False,
     help="Flag: If provided overwrite existing model files",
 )
+ignore_empty_opt = click.option(
+    "--ignore-empty",
+    is_flag=True,
+    default=False,
```
hboisgon, line 125: Can we make for export_data that the default behavior is to ignore empty data (default = IGNORE)?

hydromt/data_catalog.py
```diff
@@ -1136,6 +1135,7 @@ def export_data(
         unit_conversion: bool = True,
         meta: Optional[Dict] = None,
         append: bool = False,
+        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
```
hboisgon, line 1138: Can the default be IGNORE and not RAISE?

hydromt/data_catalog.py
```diff
@@ -1342,53 +1347,67 @@ def get_rasterdataset(
...
+                time_tuple=time_tuple,
+                single_var_as_array=single_var_as_array,
+                cache_root=self._cache_dir if self.cache else None,
+                handle_nodata=NoDataStrategy.RAISE,
                 logger=self.logger,
             )
-            return RasterDatasetAdapter._single_var_as_array(
-                data_like, single_var_as_array, variables
+            return obj
+        except NoDataException:
```
hboisgon, line 1404: Is it needed to handle nodata both in get_rasterdataset and rasterdataset.get_data? Or are we doing things twice? Ie you should pass handle_nodata in get_rasterdataset but the result of rasterdataset.get_data should already be either data, None or Error depending on the handle_nodata argument passed?
Same remark for all get_* methods in this file.

hydromt/utils.py
```diff
@@ -65,3 +66,13 @@ def _dict_pprint(d):
     import json
 
     return json.dumps(d, indent=2)
+
+
+def has_no_data(data) -> bool:
```
hboisgon, line 71: Can you specify the type of data that can be passed here? 
xr.Dataset only and not xr.DataArray? Can pd.DataFrame, gpd.GeoDataFrame and numpy.array be passed too?

hydromt/data_adapter/dataframe.py
```diff
@@ -176,54 +178,88 @@ def to_file(
...
-            fn_out = join(data_root, f"{data_name}.parquet")
-            obj.to_parquet(fn_out, **kwargs)
-        elif driver == "excel":
-            fn_out = join(data_root, f"{data_name}.xlsx")
-            obj.to_excel(fn_out, **kwargs)
-        else:
-            raise ValueError(f"DataFrame: Driver {driver} is unknown.")
+        try:
+            kwargs.pop("bbox", None)
+            obj = self.get_data(
```
hboisgon, line 183: Could you here decide to pass the ``handle_nodata`` argument? Then get_data will take care or returning either error that you want or data then you write or None and then you don't write? Instead of again doing try/except?

hydromt/data_adapter/dataframe.py
```diff
@@ -176,54 +178,88 @@ def to_file(
...
-        # slice data
-        df = DataFrameAdapter._slice_data(df, variables, time_tuple, logger=logger)
-        # uniformize data
-        df = self._apply_unit_conversion(df, logger=logger)
-        df = self._set_metadata(df)
-        return df
+        try:
+            # load data
+            fns = self._resolve_paths(variables=variables)
+            df = self._read_data(fns, handle_nodata=NoDataStrategy.RAISE, logger=logger)
```
hboisgon, line 229: do you need to pass ``handle_nodata`` here? It is not used in ``_read_data``

hydromt/data_adapter/dataframe.py
```diff
@@ -176,54 +178,88 @@ def to_file(
...
+            if has_no_data(df):
+                raise NoDataException()
```
hboisgon, line 231: Would it work if here you do ``_exec_nodata_strat`` instead or raise NoDataException? Just to avoid too many try / except

hydromt/data_adapter/dataframe.py
```diff
@@ -176,54 +178,88 @@ def to_file(
...
+            # rename variables and parse nodata
+            df = self._rename_vars(df)
+            df = self._set_nodata(df)
+            # slice data
+            df = DataFrameAdapter._slice_data(
+                df,
+                variables,
+                time_tuple,
+                handle_nodata=NoDataStrategy.RAISE,
```
hboisgon, line 240: Same here, handle_nodata is used but empty df is taken care of.
So maybe move your check on ``has_no_data`` before your current except? I think it not a probllem in case of IGNORE if mark_as_used is True, because the data was used and said there was no instance in your case (eg no lake in your model domain according to one dataset but another may tell a different story).

hboisgon (commented):
Thanks for looking into this @savente93 
It starts to be a little tough for me to follow this NoDataStrategy (I don't know so well how it works) so I still tried to understand and leave I hope still useful comments.

Main remarks:

- All typing changes look good.
- I think for export_data the default should be IGNORE rather than RAISE.
- I only reviewed dataframe and not other adapters yet but some things I noticed is that I have the feeling it gets confusing when the handle_nodata is actually handled. I think the only places that it should do something is in the lower ``get_data`` methods (eg dataframe.get_data). Then the to_file functions in the adapters or get_dataframe in the data_catalog, should only be able to propagate the ``handle_nodata`` argument all the way to ``get_data`` which takes care of returning error, data or None. Then in ``to_file`` or others you only have to maybe do a check if None and then do something or not but not again raise the error or execute the no data strategy? What do you think? Did I get it right or you think those try/except are needed everywhere?

hydromt/cli/main.py
savente93: sure!

hydromt/utils.py
savente93: good point! I forgot about it.

hydromt/data_adapter/dataframe.py
savente93: so I prefer the try except because it means we do the handeling of this in only one place, so that makes it less likely to cause inconsisten behaviour later due to e.g. accidentally changing parameters. Do you mind the try/excepts? (I've reduced them a lot since our discussion)

savente93:
@hboisgon I think I've resolved everything we talked about today. Could you take another look when you have a moment? The tests are failing because of a problem with xarray that is causing segementaion faults, for now this is not something we can do anything about, but the rest pass no problem

hydromt/data_adapter/dataframe.py
```diff
@@ -177,7 +178,15 @@ def to_file(
...
         kwargs.pop("bbox", None)
-        obj = self.get_data(time_tuple=time_tuple, variables=variables, logger=logger)
+        obj = self.get_data(
+            time_tuple=time_tuple,
+            variables=variables,
+            handle_nodata=handle_nodata,
+            logger=logger,
+        )
+
+        if obj is None:
```
hboisgon, line 188: In data_catalog,export_data, the output fn_out, etc are explicitly asked and passed. If you return None here, it needs to be handled in that function I think:

 https://github.com/Deltares/hydromt/blob/3ef7af4d4a6c0fc24f0aaa08dc5d5def2d2d8227/hydromt/data_catalog.py#L1225-L1241

hydromt/data_adapter/dataframe.py
```diff
@@ -136,14 +136,15 @@ def __init__(
...
-        driver=None,
-        variables=None,
-        time_tuple=None,
-        logger=logger,
+        data_root: StrPath,
+        data_name: str,
+        driver: Optional[str] = None,
+        variables: Optional[Variables] = None,
+        time_tuple: Optional[TimeRange] = None,
+        handle_nodata: NoDataStrategy = NoDataStrategy.IGNORE,
```
hboisgon, line 144: I think the default can still be RAISE no? Only for data_catalog.export_data I prefer to have IGNORE

hydromt/data_adapter/dataframe.py
```diff
@@ -268,12 +309,11 @@ def _set_nodata(self, df):
...
-        variables=None,
-        time_tuple=None,
-        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
-        logger=logger,
-    ):
+        df: pd.DataFrame,
+        variables: Optional[Variables] = None,
+        time_tuple: Optional[TimeRange] = None,
+        logger: Logger = logger,
+    ) -> Optional[pd.DataFrame]:
```
hboisgon, line 316: It's not shown here but handle_nodata was not removed from the docstrings

hydromt/data_adapter/geodataset.py
```diff
@@ -348,16 +381,15 @@ def _set_crs(self, ds, logger=logger):
...
-    ):
+        ds: Data,
+        variables: Optional[Variables] = None,
+        geom: Optional[Geom] = None,
+        bbox: Optional[Bbox] = None,
+        buffer: Optional[GeomBuffer] = 0,
+        predicate: Predicate = "intersects",
+        time_tuple: Optional[TimeRange] = None,
+        logger: Logger = logger,
+    ) -> Optional[Data]:
```
hboisgon, line 392: Also here remove handle_nodata from docstrings

hydromt/data_adapter/rasterdataset.py
```diff
@@ -271,69 +282,81 @@ def to_file(
...
+            if has_no_data(ds):
+                raise NoDataException()
+
+            # uniformize data
+            ds = self._apply_unit_conversions(ds, logger)
+            ds = self._set_metadata(ds)
+            self.mark_as_used()  # mark used
+            # return array if single var and single_var_as_array
+            return self._single_var_as_array(ds, single_var_as_array, variables)
+        except (NoDataException, FileNotFoundError):
```
hboisgon, line 343: I passed it for other adapters but actually I think it is best to still have two separate error messages for NoDataExecption and FileNotFoundError.
If FileNotFoundError, it could be that the user made a mistake in the path in the data catalog, so I still think it's good to return an error message that let's them know the file was not found rather than just ignore in that case.

hydromt/data_adapter/rasterdataset.py
```diff
@@ -448,15 +467,14 @@ def _set_crs(self, ds, logger=logger):
...
-        handle_nodata=NoDataStrategy.RAISE,
-        logger=logger,
+        ds: Data,
+        variables: Optional[Variables] = None,
+        geom: Optional[Geom] = None,
+        bbox: Optional[Bbox] = None,
+        buffer: GeomBuffer = 0,
+        align: Optional[bool] = None,
+        time_tuple: Optional[TimeRange] = None,
+        logger: Logger = logger,
```
hboisgon, line 477: Remove handle_nodata from docstrings

hydromt/data_adapter/rasterdataset.py
```diff
@@ -534,7 +554,10 @@ def _shift_time(self, ds, logger=logger):
 
     @staticmethod
     def _slice_temporal_dimension(
-        ds, time_tuple, handle_nodata=NoDataStrategy.RAISE, logger=logger
+        ds: Data,
+        time_tuple: Optional[TimeRange],
+        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
```
hboisgon, line 559: Do we still need handle_nodata here?

hydromt/data_adapter/rasterdataset.py
```diff
@@ -544,26 +567,20 @@ def _slice_temporal_dimension(
...
-        buffer,
-        align,
-        handle_nodata=NoDataStrategy.RAISE,
-        logger=logger,
+        ds: Data,
+        geom: Optional[Geom],
+        bbox: Optional[Bbox],
+        buffer: GeomBuffer,
+        align: Optional[bool],
+        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
```
hboisgon, line 582: And here?

hydromt/data_catalog.py
```diff
@@ -1229,10 +1229,15 @@ def export_data(
                                 variables=source_vars.get(key, None),
                                 bbox=bbox,
                                 time_tuple=time_tuple,
+                                handle_nodata=NoDataStrategy.RAISE,
```
hboisgon, line 1232: See my comment in dataframe. If NoDataStrategy is IGNORE, None is returned and I am not sure that case is handled well in this function?
Oh is this actually why you spcifically pass here RAISE and use handle_nodata and continue in case of IGNORE?
Then it's actually good, ignore my comment :)

hydromt/data_catalog.py
```diff
@@ -1286,7 +1291,7 @@ def get_rasterdataset(
         provider: Optional[str] = None,
         version: Optional[str] = None,
         **kwargs,
-    ) -> xr.Dataset:
+    ) -> Optional[xr.Dataset]:
         """Return a clipped, sliced and unified RasterDataset.
 
         To clip the data to the area of interest, provide a `bbox` or `geom`,
```
hboisgon, line 1297: I think it good in this docstring (and all other get_* methods in this file) to mention that this function returns either data, error if RAISE and None if IGNORE. This is the main doc for the users.

hydromt/data_catalog.py
```diff
@@ -1472,7 +1477,6 @@ def get_geodataframe(
                 bbox,
                 buffer,
                 predicate,
-                handle_nodata,
```
hboisgon, line 1475: Because handle_nodata is not done here, function should return gdf rather than already return out of the function?
ie gdf = GeoDataFrameAdapter._slice_data? Or at least the exex_nodata_strategy before return?
I'm guessing the following get_data call could go in the if isinstance(data_like, (str. Path)) above for more clarity?

hboisgon: I think this comment is the same for a lot of get_* methods (at least not true for get_dataframe but true for get_rasterdataset). That in case a python object is passed, the NoDataStrategy is not applied.

hydromt/utils.py
```diff
@@ -65,3 +70,15 @@ def _dict_pprint(d):
     import json
 
     return json.dumps(d, indent=2)
+
+
+def has_no_data(
+    data: Optional[Union[pd.DataFrame, gpd.GeoDataFrame, xr.Dataset, xr.DataArray]],
+) -> bool:
+    """Check whether various data containers are empty."""
```
hboisgon, line 78: Also in the docstings :) Not sure if this part of exposed API or just internal though, if internal then typehint is enough for other hydromt developpers.

tests/test_data_adapter.py
```diff
@@ -78,9 +78,9 @@ def test_rasterdataset(rioda, tmpdir):
     geom = rioda.raster.box
     da1 = data_catalog.get_rasterdataset("test.tif", geom=geom)
     assert np.all(da1 == rioda_utm)
-    with pytest.raises(FileNotFoundError, match="No such file"):
+    with pytest.raises(NoDataException):
```
hboisgon, line 81: See my comment on unmerging the error messages for FIleNotFound and NoDataException

hboisgon (commented):
Nice work @savente93 , looks much better already I have the feeling! Still a couple of things to solve before we are good to go but almost there!
For the default strategy for to_file, I guess actually IGNORE is good so you can keep.

hydromt/data_catalog.py
savente93: That is indeed why :)

hydromt/data_catalog.py
savente93: I don't think I understand this comment. I thought that was what we described before, that if the user passes in a data container themselves then we don't have to do the no_data handeling? so that we only have to handle the no data in the get data. Could you elaborate on what you mean here? (we can also quickly call about it if you think that's easier)

hboisgon: So if the user passes a geodataframe, it could still be that after slicing an empty gdf is returned. So after slicing, you may still want to do the if has_nodata check and then return None or error in case the data is empty.

hboisgon:
Hi @savente93 ! Looking good! I checked the changes, and once the last item in the conversation above is solved then we are good to go!

savente93:
@hboisgon should be all good now!

hboisgon (approved):
Yes looks all good now! Thanks a lot @savente93

PR #747, opened 2024-01-22
geodataframe reading stop opening

### LINKED ISSUE #742: Building a model results not supported file format error
### HydroMT version checks

### Reproducible Example

[wflow_build_1km_europe_pfafstetter_level_6.ini.txt](https://github.com/Deltares/hydromt/files/13988004/wflow_build_1km_europe_pfafstetter_level_6.ini.txt)
[deltares_data.yml.txt](https://github.com/Deltares/hydromt/files/13988013/deltares_data.yml.txt)
[individual_basins_pfafstetter_level6_filtered.zip](https://github.com/Deltares/hydromt/files/13988020/individual_basins_pfafstetter_level6_filtered.zip)

See the attached shapefile, deltares_data.yml file used on the HPC cluster and used ini-file to reproduce the case.

### Current behaviour

[slurm-739735.txt](https://github.com/Deltares/hydromt/files/13987975/slurm-739735.txt)

See attached file for the error (`ERROR - '/vsimem/cb24f46d5b8f405a8e51edc230ba61f6' not recognized as a supported file format.`). 

### Desired behaviour

merit_hydro_ihu should be found without problems during the step `build - rasterdataset - INFO - Reading merit_hydro_ihu raster data from /mnt/proj3/dd-23-93/hydromt/topography/merit_hydro_ihu/30sec/*.tif`

### Additional context

_No response_

Jaapel:
## Issue addressed
Fixes #742

## Explanation
Stopped using the `.open` functionality on UPath. 

## Checklist

May break specific UPath options on other filesystems, but UPath breaks with geopandas. Should wait for v1 fix

### DIFF
--- a/hydromt/gis_utils.py
+++ b/hydromt/gis_utils.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 import glob
 import logging
 import os
-from io import IOBase
 from os.path import dirname
 from typing import Optional, Tuple, Union
 
@@ -200,7 +199,9 @@ def filter_gdf(gdf, geom=None, bbox=None, crs=None, predicate="intersects"):
         raise ValueError("Either geom or bbox is required.")
     if not isinstance(geom, BaseGeometry):
         # reproject
-        if gdf.crs is not None and geom.crs != gdf.crs:
+        if geom.crs is None and gdf.crs is not None:
+            geom = geom.set_crs(gdf.crs)
+        elif gdf.crs is not None and geom.crs != gdf.crs:
             geom = geom.to_crs(gdf.crs)
         # convert geopandas to geometry
         geom = geom.unary_union
@@ -595,7 +596,7 @@ def to_geographic_bbox(bbox, source_crs):
 
 
 def bbox_from_file_and_filters(
-    fn: IOBase,
+    fn: str,
     bbox: Union[GEOM_TYPES, None] = None,
     mask: GEOM_TYPES | None = None,
     crs: CRS | None = None,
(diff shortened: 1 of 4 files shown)


savente93 (commented):
Not totally sure what the problem was but the tests seem to be adjusted, so if they catch that issue then LGTM, will be happy to approve once it's confirmed this fixes the issue

DirkEilander:
@Jaapel Would it be an idea to use fiona (i.e., default `geopandas.read_file`) as default again in `io.open_vector` method and only use `pyogrio` if a user states `driver=pyogrio` in the data catalog? This would potentially avoid some of the issues we are facing with pyogrio.

Jaapel:
> @Jaapel I agree with the approach to use fiona again as default in `io.open_vector` method. We could still use `pyogrio` if a user states `driver='pyogrio'` in the data catalog. Could the `read_info` that is used in the `gis_utils.bbox_from_file_and_filters` still cause this issue in #742? This method is only needed in combination with pyogrio right? We didn't need this method before using pyogrio?

I like to still have this method, as is does some CRS checks on the bbox and the mask, and constructs a bbox to filter the data with while reading. The data will be filtered further afterwards when using a mask.

DirkEilander:
I agree! Was just thinking which change might have caused the issue. If caused by the pyogrio 'read_info' method we could use this utility method only when also reading with pyogrio if a user sets 'driver=pyogrio' for now.

savente93 (approved):
OP of the issue reported this solved the issue, and I see nothing out of the ordenary which is good enough for me

DirkEilander (approved):
LGTM!

PR #749, opened 2024-01-23
Add datacatalog test script

### LINKED ISSUE #735: Add a testing script for testing predefined data catalogs
### Kind of request

Adding new functionality

### Enhancement Description

Currently there is no testing in place for changing or adding a dataset to the predefined data catalogs. This has recently lead to some bugs popping up. It is a good idea to make a script that checks if the dataset can be properly opened and interacted with, maybe clipping it to a small area. 

### Use case

_No response_

### Additional Context

_No response_

Tjalling-dejong:
## Issue addressed
Fixes #735 

## Explanation
I made a script for testing the predefined data catalogs. It works with command line arguments. For instance this call will test the chelsa dataset source of version 1.2:

`python path/to/datacatalog.yml -d chelsa -v 1.2`

This call will open the dataset and clip the dataset to a small bounding box if the dataset contains geo data. This can take some time. It also possible to supply your own bounding box with the -r flag. Bounding box must be a dict containing a list of xmin, ymin, xmax, ymax coordinates in EPSG 4326.

The script can also check if all the paths mentioned in the data catalog exist. This is done by calling the script and giving only the path to the data catalog.
The script will also always check if the given .yml is a valid data catalog.

## Checklist

Let me know if you have more ideas for testing the dataset sources.

### DIFF
--- a/data/catalogs/test_data_catalog.py
+++ b/data/catalogs/test_data_catalog.py
@@ -23,6 +23,7 @@ import argparse
 import json
 
 from dask.distributed import Client
+from pydantic_core import ValidationError
 
 from hydromt import DataCatalog
 from hydromt.log import setuplog
@@ -114,5 +115,9 @@ if __name__ == "__main__":
         test_data_catalog(args, datacatalog)
     # Validate data catalog yaml
     logger.info("Validating data catalog")
-    DataCatalogValidator().from_yml(args.data_catalog)
-    logger.info("Data catalog is valid!")
+    try:
+        DataCatalogValidator().from_yml(args.data_catalog)
+        logger.info("Data catalog is valid!")
+    except ValidationError as e:
+        logger.error("Data catalog is not valid!")
+        logger.error(e)
--- a/data/catalogs/test_data_catalog.py
+++ b/data/catalogs/test_data_catalog.py
@@ -1,10 +1,13 @@
 """Script for testing predefined data catalogs."""
 import argparse
-import os
+
+from dask.distributed import Client
 
 from hydromt import DataCatalog
 from hydromt.log import setuplog
+from hydromt.validators.data_catalog import DataCatalogValidator
 
+client = Client(processes=False)
 logger = setuplog()
 
 parser = argparse.ArgumentParser("Test predefined data catalog")
@@ -13,8 +16,15 @@ parser.add_argument(
     "-ds", "--dataset", help="The name of the dataset to test", required=False
 )
 parser.add_argument(
-    "-dsv" "--dataset_version",
+    "-dsv",
+    "--dataset_version",
     help="Optional version of the dataset to test for",
+    required=False,
+)
+parser.add_argument(
+    "-bbox",
+    "--boundingbox",
+    help="Bounding box for testing clipping spatial data",
     required=False,
 )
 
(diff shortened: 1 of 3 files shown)


data/catalogs/test_data_catalog.py
```diff
@@ -0,0 +1,117 @@
...
+    'python test_data_catalog.py path/to/datacatalog.yml'
+
+Separate data catalog sources can be tested by giving the dataset source name as
+an extra argument. The argument should be preceded by -ds. It is also possible to specify
+the version of the dataset source by supplying the call with a -dsv flag and version
+number. The dataset is opened and if the dataset contains geo data the dataset is clipped
+to a small bounding box surrounding Deltares Delft office.
+Example dataset source test call:
+    'python test_data_catalog.py path/to/datacatalog.yml -ds chelsa -dsv 1.2
+    -r "{'bbox':[4.333409,51.962159,4.42835,52.006873]}"'
```
savente93, line 16: if you're using multiple characters it's customary to use `--` for your flags and only `-` if you use one character. (e.g. `-d` but `--dsv`)

data/catalogs/test_data_catalog.py
```diff
@@ -0,0 +1,117 @@
...
+import json
+
+from dask.distributed import Client
+
+from hydromt import DataCatalog
+from hydromt.log import setuplog
+from hydromt.validators.data_catalog import DataCatalogValidator
+
+client = Client(processes=False)
+logger = setuplog()
```
savente93, line 31: Why proccesses = False? additionally I think this should also go in the `__main__` branch

data/catalogs/test_data_catalog.py
```diff
@@ -0,0 +1,117 @@
...
+        region_json = args.region.replace("'", '"')
+        bbox = json.loads(region_json)["bbox"]
+        print(bbox)
+    else:
+        bbox = [
+            4.333409,
+            51.962159,
+            4.42835,
+            52.006873,
+        ]  # small area surrounding Deltares Delft office
```
savente93, line 51: Should be documented in your docstring

data/catalogs/test_data_catalog.py
```diff
@@ -0,0 +1,117 @@
...
+            datacatalog.get_source(source)._resolve_paths()
+        except FileNotFoundError as e:
+            logger.error(f"File not found for dataset source {source}: {e}")
+
+            error_count += 1
+        except ValueError as e:
+            logger.error(
+                f"Something went wrong with creating path string for dataset source {source}: {e}"
+            )
+            error_count += 1
```
savente93, line 84: Nice that you can keep going and only report errors at the end!

data/catalogs/test_data_catalog.py
```diff
@@ -0,0 +1,117 @@
...
+    logger.info(f"Encountered {error_count} errors")
```
savente93, line 85: think this should be logger.error if error_count >0

data/catalogs/test_data_catalog.py
```diff
@@ -0,0 +1,117 @@
...
+    elif source.data_type == "Dataset":
+        dataset = datacatalog.get_dataset(args.dataset)
+    assert dataset is not None
+
+
+def test_data_catalog(args, datacatalog):
+    """Tests the paths of the given data catalog."""
+    error_count = 0
+    logger.info("Checking paths of data catalog sources")
+    for source in datacatalog.get_source_names():
```
savente93, line 72: I think the source manes might not be loaded at this point yet (because we lazy load sources iirc) if you checked it  and this returns all the source names, that's fine and you can just close this comment, if not please double check, it would be a shame if the script didn't actually do anything.

data/catalogs/test_data_catalog.py
```diff
@@ -0,0 +1,117 @@
...
+        try:
+            logger.info(f"Checking paths of {source}")
+            datacatalog.get_source(source)._resolve_paths()
```
savente93, line 75: Do we not slice here like in the single case for a reason?

savente93 (changes requested):
Nice work Tjalling. Couple small comments but after those ar eresolved I'll be happy to approve

data/catalogs/test_data_catalog.py
Tjalling-dejong: I'm not sure what you mean here could you elaborate?

data/catalogs/test_data_catalog.py
Tjalling-dejong: get_source_names works fine :)

data/catalogs/test_data_catalog.py
savente93: in the case of a single data source you slice on a predefined bbox, how come you don't do that when you check the whole catalog?

savente93:
@Tjalling-dejong You've missed a couple of comments it seems. Can you respond to those?

data/catalogs/test_data_catalog.py
Tjalling-dejong: Because then it will probably take a day to run this script

data/catalogs/test_data_catalog.py
Tjalling-dejong: changed it logger.error

data/catalogs/test_data_catalog.py
Tjalling-dejong: I did mention it

data/catalogs/test_data_catalog.py
Tjalling-dejong: It's for Dask. I do not really know what is does though

savente93 (approved):
LGTM

Issue #720, opened 2024-01-10
Simplify data drivers


savente93:
### Kind of request

Currently, `DataAdapter`s are responsible for both the representation of different data sources in the `DataCatalog`, reading in the data and transforming the data to a uniform data representation in memory. This makes the class responsible for a lot of functions and hard to modify or extend by the plugins.

### Enhancement Description

We propose that a `Driver` should be responsible for reading the data and creating a memory representation, while the `Adapter` should do generic transformations and filtering/slicing. A `DataSource` should represent items in the `DataCatalog`, which can check at read time whether all the required fields are present.

### Use case

This should make testing and maintenance easier, while being more flexible to customize for plugins.

Jaapel:
Posting https://github.com/Deltares/hydromt/issues/432 here for reference discussions

Jaapel:
Look at this `DataCatalog` entry:
```yaml
gtsm_codec_reanalysis_{freq}_v1:
  crs: 4326
  data_type: GeoDataset
  driver: netcdf
  kwargs:
    chunks:
      stations: 10
      time: -1
  meta:
    category: ocean
    paper_doi: 10.3389/fmars.2020.00263
    paper_ref: Muis at al (2020)
    source_license: https://cds.climate.copernicus.eu/api/v2/terms/static/licence-to-use-copernicus-products.pdf
    source_url: https://doi.org/10.24381/cds.8c59054f
    source_version: v1
  path: p:/11205028-c3s_435/01_data/01_Timeseries/timeseries2/{variable}/reanalysis_{variable}_{freq}_{year}_{month:02d}_v1.nc
  placeholders:
    freq: [10min, hourly, dailymax]
  rename:
    station_x_coordinate: lon
    station_y_coordinate: lat
    stations: index
```
There is a placeholder in the title entry, which can easily be expanded using the `placeholders` entry in the yaml document. But what about the `path` entry? There the `freq` is coming back, but there are also `year`, `month` and `variable`. For `RasterDataset` there is also `zoom_level`.
Is this some generic behavior for certain datasets or can we just use this in the more generic `DataSource` classes (e.g. `RasterDataSource`? want to split out the `_resolve_paths` between what the driver should be responsible for (remote data access -> `filesystem.glob`) and what is generic over multiple different data sources (handling name conventions). @DirkEilander @hboisgon do you think these naming conventions are truly generic? So far the logic seems to be to fill in and capture these "known" placeholders and if you do not recognize the placeholder, place `*`.
[...]

PR #750, opened 2024-01-24
Add driver for GeoDataFrame

Jaapel:
## Issue addressed
Adresses for GeoDataFrame #720

## Explanation
Since we want to work with multiple sources of data, we seperate functionality that the DataAdapter class was responsible for and split it until multiple classes. I implemented this just for the `GeoDataFrame` data type within HydroMT for a single driver (`PyogrioDriver`).
I created a `Driver` class for loading the data. This class is specific for one `DataAdapter`. This class should be extendable for plugins and data sources that we do not want to maintain (like custom APIs).
I created a `MetaDataResolver` class for resolving the URI (think naming convention, but later also a STAC catalog or a GeoNetwork). This resolver is generic for all `DataAdapters`. This class should be extendable to incorporate different ways of discovering data from a single URI.
I created a `DataSource` for loading items from the `DataCatalog` and verifying whether they are correct. These are specific for every `DataAdapter`.

## Checklist


hydromt/data_adapter_v1/geodataframe_adapter.py
```diff
@@ -0,0 +1,313 @@
...
+
+    def get_data(
+        self,
+        bbox: list[float] | None = None,
+        mask: gpd.GeoDataFrame | None = None,
+        buffer: float = 0.0,
+        variables: list[str] | None = None,
+        predicate: str = "intersects",
+        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
+        logger: Logger | None = None,
```
savente93, line 40: The `| None` syntax doesn't work for python 3.9. Until we drop that you'll have to use `Optional` from the typing library

hydromt/data_adapter_v1/geodataframe_adapter.py
```diff
@@ -0,0 +1,313 @@
...
+    ) -> gpd.GeoDataFrame:
+        """Read in data and transform them to HydroMT standards."""
+        gdf: gpd.GeoDataFrame = self.source.read_data(
+            bbox, mask, buffer, variables, predicate, handle_nodata, logger=logger
+        )
+        self.used = True  # mark used
```
savente93, line 46: should this be handled by the data catalog?

hydromt/data_adapter_v1/geodataframe_adapter.py
```diff
@@ -0,0 +1,313 @@
...
+        gdf: gpd.GeoDataFrame,
+        variables: str | list[str] | None = None,
+        geom: GEOM_TYPES | None = None,
+        bbox: GEOM_TYPES | None = None,
+        crs: CRS | None = None,
+        buffer: float = 0.0,
+        predicate: str = "intersects",  # TODO: enum available predicates
+        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,  # TODO: review NoDataStrategy + axes
+        logger: Logger = logger,
+    ) -> gpd.GeoDataFrame:
```
savente93, line 91: In the latest no data PR we landed on data returning functions can optionally turn None if the nodata strategy was set to IGNORE and triggred, since that's easier to check for

hydromt/typing.py
```diff
@@ -5,15 +5,53 @@
...
-Bbox = Tuple[float, float, float, float]
+Bbox = Annotated[Tuple[float, float, float, float], _validate_bbox]
 Crs = int
 TotalBounds = Tuple[Bbox, Crs]
-TimeRange = Tuple[datetime, datetime]
+TimeRange = Annotated[
+    Tuple[datetime, datetime],
+    BeforeValidator(_time_tuple_from_str),
+    AfterValidator(_timerange_validate),
+]
```
savente93, line 54: :D

hydromt/data_sources/data_source.py
```diff
@@ -0,0 +1,53 @@
+"""Abstract DataSource class."""
+from typing import Any
+
+from pydantic import BaseModel, Field
+
+from hydromt import DataCatalog
+from hydromt.metadata_resolvers.resolver_plugin import RESOLVERS
+
+
+class DataSource(BaseModel):
```
savente93, line 10: Not that this is wrong, but I think we should make a glossary of the terms used in the code because it's getting a bit hard for me to keep track of, let alone new folk. Here I mean the precise definition of things like source/catalog_item/adapter etc.

hydromt/data_sources/data_source.py
```diff
@@ -0,0 +1,53 @@
...
+        provider: str | None = None,
+        version: str | None = None,
+    ) -> "DataSource":
+        """Create Data source from DataCatalog."""
+        return cls.model_validate(
+            catalog.get_source(key, provider, version).update("name", key)
+        )
+
+    def _validate_metadata_resolver(cls, v: Any):
+        assert isinstance(v, str), "metadata_resolver should be string."
```
savente93, line 35: Not sure how common this is, but asserts can get turned off when someone runs python with specific settings iirc which means that under certain circumstances these checks would dissapear. If these are critical we should just use if -raise statements even though that's more laborious

hydromt/data_sources/data_source.py
```diff
@@ -0,0 +1,53 @@
...
+        assert v in RESOLVERS, f"unknown MetaDataResolver: '{v}'."
+        RESOLVERS.get(v)
+
+    name: str
+    attrs: dict[str, Any] = Field(default_factory=dict)
+    version: str | None = Field(default=None)
+    provider: str | None = Field(default=None)
+    driver: str
+    metadata_resolver: str
```
savente93, line 44: can be an item for later, but seeing the code in the resolver plugin file I think this sould be an enum or literal.

hydromt/data_sources/data_source.py
```diff
@@ -0,0 +1,53 @@
...
+    driver_kwargs: dict[str, Any] = Field(default_factory=dict)
+    unit_add: dict[str, Any] = Field(default_factory=dict)
+    unit_mult: dict[str, Any] = Field(default_factory=dict)
+    rename: dict[str, str] = Field(default_factory=dict)
+    nodata: float | int | dict[str, float | int] | None = Field(default=None)
+    extent: dict[str, Any] = Field(default_factory=dict)  # ?
```
savente93, line 50: iirc extent has a type defined in typing.py and if not, I think there should be one :sweat_smile:

hydromt/data_sources/geodataframe_data_source.py
```diff
@@ -0,0 +1,69 @@
...
+from pydantic import field_validator
+
+from hydromt.drivers.geodataframe_driver import GeoDataFrameDriver
+from hydromt.drivers.pyogrio_driver import PyogrioDriver
+from hydromt.metadata_resolvers.resolver_plugin import RESOLVERS
+from hydromt.nodata import NoDataStrategy
+
+from .data_source import DataSource
+
+_KNOWN_DRIVERS: dict[str, GeoDataFrameDriver] = {"pyogrio": PyogrioDriver}
```
savente93, line 15: I think a frozen dict would be better here so it doesn't get moddified accidentally

hydromt/data_sources/geodataframe_data_source.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        if len(uris) > 1:
+            raise ValueError("GeoDataFrames cannot have more than 1 source URI.")
+        uri = uris[0]
+        gdf_driver: GeoDataFrameDriver = driver_from_str(
+            self.driver, **self.driver_kwargs
+        )
+        return gdf_driver.read(
+            uri, bbox, mask, buffer, self.crs, predicate, logger=logger
+        )
```
savente93, line 69: I really like this flow, though I think some of the things, like constructing the driver could be done during init? since one sintance of a geodataframe_source shouldn't need to switch drivers after instantiation right?

hydromt/drivers/pyogrio_driver.py
```diff
@@ -0,0 +1,90 @@
...
+from .geodataframe_driver import GeoDataFrameDriver
+
+
+class PyogrioDriver(GeoDataFrameDriver):
+    """Driver to read GeoDataFrames using the `pyogrio` package."""
+
+    def read(
+        self,
+        uri: str,
+        bbox: list[int] | None = None,
```
savente93, line 22: Bbox definitely has a type in typing.py. If we're using validated pydantic models this is fine, if not I think we ougt to prefer tuples because for those we can assert their length in the type itself .

hydromt/metadata_resolvers/convention_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+import numpy as np
+import pandas as pd
+
+from hydromt.nodata import NoDataStrategy
+from hydromt.typing import Bbox, TimeRange
+
+from .metadata_resolver import MetaDataResolver
+
+if TYPE_CHECKING:
+    from hydromt.data_sources.data_source import DataSource
```
savente93, line 16: did not know this construct was a thing, TIL.

hydromt/metadata_resolvers/convention_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+
+
+class ConventionResolver(MetaDataResolver):
+    """MetaDataResolver using HydroMT naming conventions."""
+
+    _uri_placeholders = {"year", "month", "variable", "zoom_level"}
```
savente93, line 22: Frozen set is better if this does nto need to be chagned

hydromt/metadata_resolvers/convention_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _expand_uri_placeholders(
+        self, uri: str, time_tuple: tuple[str, str] | None, variables: set[str] | None
```
savente93, line 25: I think we can have some stricter types here, but in general I like this flow!

hydromt/metadata_resolvers/convention_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+        if "{" in uri:
+            uri_expanded = ""
+            for literal_text, key, fmt, _ in Formatter().parse(uri):
+                uri_expanded += literal_text
+                if key is None:
+                    continue
+                key_str = "{" + f"{key}:{fmt}" + "}" if fmt else "{" + key + "}"
+                # remove unused fields
+                if key in ["year", "month"] and time_tuple is None:
+                    uri_expanded += "*"
```
savente93, line 39: are these globs or more place holders for later?

hydromt/metadata_resolvers/convention_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+    def _get_variables(self, variables: list[str], rename: dict[str, str]) -> list[str]:
+        variables: list[str] = np.atleast_1d(variables).tolist()
+        mv_inv: dict[str, str] = {v: k for k, v in rename.items()}
+        vrs: dict[str] = [mv_inv.get(var, var) for var in variables]
+        return vrs
+
+    def resolve(
+        self,
+        source: "DataSource",
+        *,
```
savente93, line 74: what's this * do here? I don't think I know this syntax

hydromt/metadata_resolvers/convention_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+        # TODO: align? -> from RasterDataSetAdapter
+        geom: gpd.GeoDataFrame | None = None,
+        buffer: float = 0.0,
+        predicate: str = "intersects",
+        variables: list[str] | None = None,
+        zoom_level: int = 0,
+        handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
+        **kwargs,
+    ) -> list[str]:
+        """Resolve the placeholders in the URI."""
```
savente93, line 86: just the "hydromt placeholders" not any file like globs right?

hydromt/metadata_resolvers/convention_resolver.py
```diff
@@ -0,0 +1,109 @@
...
+        uri_expanded, keys = self._expand_uri_placeholders(
+            source.uri, timerange, variables
+        )
+        if timerange:
+            dates = self._get_dates(keys, timerange)
+        else:
+            dates = pd.PeriodIndex(["2023-01-01"], freq="d")
```
savente93, line 93: will this work for things like data frames that don't have a time dimension? I'm not sure I can quite follow where this dates go

hydromt/metadata_resolvers/resolver_plugin.py
```diff
@@ -0,0 +1,5 @@
+"""Resolvers detected."""
+from .convention_resolver import ConventionResolver
+from .metadata_resolver import MetaDataResolver
+
+RESOLVERS: dict[str, MetaDataResolver] = {"convention_resolver": ConventionResolver()}
```
savente93, line 5: Again, I think an enum of frozen dict would be better here if possible.

hydromt/typing.py
```diff
@@ -5,15 +5,53 @@
...
+def _validate_bbox(
+    bbox: tuple[float, float, float, float],
+) -> tuple[float, float, float, float]:
+    assert (
+        bbox[0] < bbox[2]
+    ), f"bbox minx: '{bbox[0]}' should be less than maxx: '{bbox[2]}'."
+    assert (
+        bbox[1] < bbox[3]
+    ), f"bbox miny: '{bbox[1]}' should be less than maxy: '{bbox[3]}'."
+
```
savente93, line 42: Do we want to validate a crs here too?

tests/data_sources/test_geodataframe_data_source.py
```diff
@@ -0,0 +1,36 @@
...
+import geopandas as gpd
+import numpy as np
+import pytest
+from pyogrio.errors import DataSourceError
+
+from hydromt.data_sources.geodataframe_data_source import GeoDataFrameDataSource
+
+
+class TestGeoDataFrame:
+    @pytest.fixture(scope="class")
```
savente93, line 12: do want a class scope here? or function? (if that's the correct term) i.e. do we want changes to this to persist for other tests? ( I'd assume no, but am happy to be corrected)

tests/drivers/test_pyogrio_driver.py
```diff
@@ -0,0 +1,34 @@
...
+        gdf: gpd.GeoDataFrame = driver.read(uri)
+        assert np.all(gdf == geodf)
+
+    def test_read_with_filters(self, uri: str):
+        driver = PyogrioDriver()
+        bbox: Bbox = (-60, -34.5600, -55, -30)
+        gdf: gpd.GeoDataFrame = driver.read(
+            uri, bbox=(-60, -34.5600, -55, -30), buffer=10000
+        )
+        assert gdf.shape == (1, 4)
```
savente93, line 30: very nice!

tests/data_sources/test_geodataframe_data_source.py
```diff
@@ -0,0 +1,36 @@
...
+        return GeoDataFrameDataSource(
+            name="geojsonfile",
+            driver="pyogrio",
+            metadata_resolver="convention_resolver",
+            uri=fn_gdf,
+        )
+
+    def test_read_data(
+        self, geodf: gpd.GeoDataFrame, example_source: GeoDataFrameDataSource
+    ):
```
savente93, line 27: not really about this code, but couldn't find a better place to put this comment. I think these tests are really nice. If possible I'd also like to see some tests for the resolver so we can more a little bit more towards having unit tests as well. This is a really nice setup thought!

savente93 (commented):
Very nice work Jaap, I think the proof of concept looks really nice. I left a lot of comments, many of them pedantic, but they're just suggestions. In general I think this is going to a really good place!

hydromt/data_adapter_v1/geodataframe_adapter.py
Jaapel: Not really sure. We can rethink this. So far the idea seems to be: DataCatalog -> DataSource( has: MetaDataResolver and Driver) -> data

hydromt/data_adapter_v1/geodataframe_adapter.py
Jaapel: It is difficult to "coerce" nodata in the driver (it just does not read anything), as wel as the metadata resolver (it does not find anything), so perhaps we always raise there and catch an error in the adapter and return None? I do not want to check for `None` so often.

hydromt/data_sources/data_source.py
Jaapel: Agreed. We can change terms as well

hydromt/data_sources/data_source.py
Jaapel: Later on people should be able to write plugins where they register new drivers or resolvers. Then we should identify them with a string, or dynamically build an enum. What do you prefer?

hydromt/data_sources/geodataframe_data_source.py
Jaapel: This is a placeholder for plugins later, I should add a comment there.

hydromt/data_sources/geodataframe_data_source.py
Jaapel: You are probably right. This also depends on which APIs we want to expose.

hydromt/metadata_resolvers/convention_resolver.py
Jaapel: You will have to then type hint using `"`, like `def f(a: "MyType"):`

hydromt/metadata_resolvers/convention_resolver.py
Jaapel: globs

hydromt/metadata_resolvers/convention_resolver.py
Jaapel: Yes, those I leave to the driver

hydromt/metadata_resolvers/convention_resolver.py
Jaapel: I build up the `**kwargs` for the `str.format(**kwargs)` function I use later on.

Jaapel: Open to suggestion on how to do this btter of course

hydromt/metadata_resolvers/resolver_plugin.py
Jaapel: I added some comments

tests/data_sources/test_geodataframe_data_source.py
Jaapel: Yeah I changed it already

hydromt/data_adapter_v1/geodataframe_adapter.py
savente93: I think I was refering to the `self.used = True` specifically, because the context in which the data was used matters I'd say. The read data is fine to stay .

hydromt/data_adapter_v1/geodataframe_adapter.py
savente93: While I agree with you, we'll have to check for errors regularly anyway, whether that's by checking for None, or using a try/except I don't really care that much. @hboisgon seemed to have a preference for None though

hydromt/data_sources/data_source.py
savente93: Forgot about that. Honestly if we build enumbs dinamically they mostly lose their usefulness to me, so strings are fine.

hydromt/metadata_resolvers/convention_resolver.py
savente93: whill this cause issues in for example API endpoints?

hydromt/metadata_resolvers/convention_resolver.py
savente93: TIL!

hydromt/metadata_resolvers/convention_resolver.py
savente93: aahhh okay. I'll have a think about this

hydromt/metadata_resolvers/convention_resolver.py
Jaapel: I wonder if we can use these placeholders in case of querying APIs. I would then point the URI to the API and have the reset of the settings as driver kwargs.

Jaapel:
@savente93 I notice that the tests are not passing due to "mamba not found" on windows test nodes. Any ideas?

savente93:
@Jaapel yeah, I still haven't had time to properly configure the windows tests, that's one me. You can ignore that until I fixed that.

hydromt/data_sources/data_source.py
```diff
@@ -0,0 +1,94 @@
...
+                return target_cls.model_validate(
+                    obj,
+                    strict=strict,
+                    from_attributes=from_attributes,
+                    context=context,
+                )
+            except StopIteration:
+                raise ValueError(f"Unknown 'data_type': '{data_type}'")
+
+        raise ValueError("DataSource needs 'data_type'")
```
savente93, line 49: I'm having a hard time following this code, can you elaborate a bit on what it's doing?

hydromt/data_sources/geodataframe.py
```diff
@@ -0,0 +1,84 @@
...
+from hydromt.drivers.geodataframe_driver import GeoDataFrameDriver
+from hydromt.drivers.pyogrio_driver import PyogrioDriver
+
+from .data_source import DataSource
+
+# placeholder for proper plugin behaviour later on.
+_KNOWN_DRIVERS: dict[str, GeoDataFrameDriver] = {"pyogrio": PyogrioDriver}
+
+
+def driver_from_str(driver_str: str, **driver_kwargs) -> GeoDataFrameDriver:
```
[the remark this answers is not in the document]
savente93: also should this be in the base class?

hydromt/data_sources/geodataframe.py
```diff
@@ -0,0 +1,84 @@
...
+    data_type = "GeoDataFrame"
+    driver: GeoDataFrameDriver
+
+    @model_validator(mode="before")
+    @classmethod
+    def _validate_data_type(cls, data: Any) -> Any:
+        if isinstance(data, dict):
+            if data.get("data_type", "") != "GeoDataFrame":
+                raise ValueError("'data_type' must be 'GeoDataFrame'.")
+        return data
```
savente93, line 45: is this for internal sanity checks or does this doe anything for the user?

hydromt/drivers/pyogrio_driver.py
```diff
@@ -0,0 +1,92 @@
...
+from pyogrio import read_dataframe, read_info
+from pyproj import CRS
+from shapely.geometry.base import BaseGeometry
+
+from hydromt._typing import Bbox, Geom
+from hydromt.gis import parse_geom_bbox_buffer
+
+from .geodataframe_driver import GeoDataFrameDriver
+
+GEOM_TYPES = Union[Geom, BaseGeometry]
```
savente93, line 15: move to typing/type_def.py please (or make a TODO for it)

hydromt/gis/utils.py
```diff
@@ -209,7 +210,12 @@ def filter_gdf(gdf, geom=None, bbox=None, crs=None, predicate="intersects"):
     return idx
 
 
-def parse_geom_bbox_buffer(geom=None, bbox=None, buffer=0):
+def parse_geom_bbox_buffer(
+    geom: GPD_TYPES | None = None,
+    bbox: Bbox | None = None,
+    buffer: float = 0.0,
+    crs: CRS | None = None,
```
savente93, line 217: 3.10 syntax, I'm surprised this got through the tests.

hydromt/metadata_resolvers/metadata_resolver.py
```diff
@@ -0,0 +1,33 @@
...
+    @abstractmethod
+    def resolve(
+        self,
+        source: "DataSource",
+        *,
+        timerange: Optional[TimeRange] = None,
+        bbox: Optional[Bbox] = None,
+        geom: Optional[gpd.GeoDataFrame] = None,
+        buffer: float = 0.0,
+        predicate: str = "intersects",
```
savente93, line 26: There's a `Predicate` literal type for this in `_typing`

hydromt/drivers/geodataframe_driver.py
```diff
@@ -0,0 +1,33 @@
...
+    def read(
+        self,
+        uri: str,
+        bbox: Optional[List[int]] = None,
+        mask: Optional[gpd.GeoDataFrame] = None,
+        buffer: float = 0.0,
+        crs: Optional[CRS] = None,
+        predicate: str = "intersects",
+        logger: Optional[Logger] = None,
+        # handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
```
savente93, line 26: should this still be commented out? if so is there a TODO for resolving this?

hydromt/metadata_resolvers/convention_resolver.py
```diff
@@ -0,0 +1,110 @@
...
+        """Expand known placeholders in the URI."""
+        keys: list[str] = []
+
+        if "{" in uri:
+            uri_expanded = ""
+            for literal_text, key, fmt, _ in Formatter().parse(uri):
+                uri_expanded += literal_text
+                if key is None:
+                    continue
+                key_str = "{" + f"{key}:{fmt}" + "}" if fmt else "{" + key + "}"
```
savente93, line 38: Totally fine to do this later, but I'm wondeirng if we can replace this whole code block with a regex.

hydromt/metadata_resolvers/convention_resolver.py
savente93: Haven't thought about this much yet, so this is fine for now.

pyproject.toml
```diff
@@ -140,6 +141,9 @@ exclude = ["docs", "examples", "envs", "tests", "binder", ".github"]
 [tool.pytest.ini_options]
 addopts = "--ff "
 testpaths = ["tests"]
+markers = [
+  "integration: marks tests as being integration tests"
+]
```
savente93, line 146: do we also want one for unit tests?

tests/conftest.py
```diff
@@ -347,6 +349,26 @@ def mesh_model(griduda):
...
+@pytest.fixture()
+def mock_resolver() -> MetaDataResolver:
+    class MockMetaDataResolver(MetaDataResolver):
+        def resolve(self, uri, *args, **kwargs):
+            return [uri]
+
+    resolver = MockMetaDataResolver.model_validate({})
+    return resolver
+
+
```
savente93, line 371: very nice!

savente93 (approved):
Very nice work Jaap. I didn't look through al lthe tests in detail, but I'm very impressed by them. I think the real proff of the pudding will be in the integration, so for now I think this is great. I left a couple of small comments, but overal I'm really liking the direction this is going in!

hydromt/data_sources/geodataframe.py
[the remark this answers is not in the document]
Jaapel: Drivers are specific per `DataType`, so I will keep them in their seperate spots.
This str is now literal, but the options will be extended when we implement plugin-behaviour.

hydromt/data_sources/data_source.py
Jaapel: I wanted to be able to construct DataSources in a generic way from it's base class, as that is what should happen when we read from the DataCatalog. This should be based on the `data_type` in the `DataCatalog` entry. In this case walk through the following steps:
1. Check whether the entry has a `data_type` field, otherwise raise an error.
2. Then try to find a subclass with a matching `data_type` field, which is a required class attribute for `DataSource`. All subclasses are loaded in the `__init__.py` of the `data_source` submodule. If this does not work, I raise a `ValueError`.
3. Then I call the Pydantic `BaseModel.model_validate` class method on the entry, so that the correct `DataSource` (e.g. `GeoDataFrameSource` is instantiated. This is also where the validation happens. The result is returned

hydromt/data_sources/geodataframe.py
Jaapel: Pydantic does not validate class variables, so I had to make a "before" validator to make sure that a `data_type` is included with the correct value. Otherwise the user can insert a wrong `data_type` and their mistake will be fixed silently.

pyproject.toml
Jaapel: We could!

hydromt/data_sources/data_source.py
savente93: that makes sense, good job!

PR #757, opened 2024-01-30
Create Model root class

### LINKED ISSUE #728: new `ModelRoot` class
### Kind of request

### Enhancement Description

Currently the `Model._root` is a `Path` instance and a separate property from the `Model._write` and `Model._read` properties and methods which are derived from the read/writing mode and used by `Model._assert_read_mode` and `Model._assert_write_mode` methods. 

These methods could be combined into a single ModelRoot class, which might be a good starting point to improve the read/write behavior together with #671

### Use case

This mainly targets better code structure. See #671 for an example

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #728 

## Explanation
It was a bit of a tough nut to crack but so far I'm actually quite happy with this implementaion. Couple of notes:
1. This should ensure that paths are handled in a cross platform(*) way
2. Dirk had mentioned that we should leave the writing to the model comonent classes but I still wanted to have at least something to test, so I've implemented it in an optional way anyway. When the comonent classes come along we can move the functionality. 
3. Added tests where I could. I experimented with the "table driven unit test" pattern from golang and I actually really like it here. I think it makes the error logs and the tests a lot cleaner to read. Let me know what you think of that. 
4. The folder creating/checking code is pretty much copy pasted from the current model implementation and cleaned up a little bit. This should be totally fine
5. About that asterisk in point 1: A lot of trouble has come from trying to mix paths from different OSes. (e.g. trying to handle windows paths on linux). In the case that you are handeling paths from your own os (e.g. windows paths on a windows machine) everything is handled as you'd expect with pretty much no intervention from us. However when mixing paths to try and get some semi workable paths I had to make some assumptions. I think this is probably the most controversial bit. I've made the best compromises I could, but please let me know what you think.

##  The conventions I've implemented:

1.  on linux a path like C:\\whatever maps to /mnt/c/whatever. for linux this is the cannonical place to mount things like network shares and I think it most likely to be easy to work with. Relative paths are simply converted to their posix/linux equivalent and used like normal
2. on Windows a linux path like /mnt/d/whatever will map to D:\\whatever mostly as an inverse as the previous point. a path like /home/user/whatever (the mostly likely) will map to C:\\Users\{login_name_of_user}\\whatever. 

The tl;dr is that I tried to match the conventions of each platform like home directories and mount points as closely as possible to give folks an easy time, but it does mean that (only) in the cross platform case you can't just place things anywhere. Given that you can still do that if you're using the paths native to your platform I think this is an acceptable compromise. Please let me know. 

## Checklist


hydromt/models/_v1/model_root.py
```diff
@@ -0,0 +1,187 @@
...
+                    # remove handler and close file
+                    self.logger.handlers.pop(i).close()
+                else:
+                    has_log_file = True
+                break
+        if not has_log_file:
+            new_path = join(self._path, "hydromt.log")
+            add_filehandler(self.logger, new_path, log_level)
+
+    def _set_path(self, path: PathLike) -> None:
```
hboisgon, line 106: I think we should have a public ``ModelRoot.set_path`` function. You could decide to still keep this function that deals with posix/windows path and return the path in correct format and setting the actual path property can be done in the ``set_path`` function.

hydromt/models/_v1/model_root.py
```diff
@@ -0,0 +1,187 @@
...
+
+            else:
+                # posix path on a posix system
+                self._path: Path = Path(abspath(ppp))
+
+    def set_mode(self, mode: ModeLike) -> None:
+        """Set the mode of the model."""
+        self.mode: ModelMode = ModelMode.from_str_or_mode(mode)
+
+    def is_writing_mode(self) -> bool:
```
hboisgon, line 155: I would add r+ aslo in the writing modes especially if you do not have here an override accessor.

hydromt/models/_v1/model_root.py
```diff
@@ -0,0 +1,187 @@
...
+        """Test whether we are in writing mode or not."""
+        return self.mode.is_writing_mode()
+
+    def is_reading_mode(self) -> bool:
+        """Test whether we are in reading mode or not."""
+        return self.mode.is_reading_mode()
+
+    def _posix_is_mounted(self, path) -> bool:
```
hboisgon, line 163: For these functions and all the ones below, could they be more added as utility functions in this script or somewhere else rather than methods of ``ModelRoot``? They do not use any propoerty of ``ModelRoot`` and may actually be useful also for eg for data catalog or other places in the code where we want to deal with cross-platform compatibility for path?

hydromt/typing.py
```diff
@@ -37,3 +37,44 @@ class ErrorHandleMethod(Enum):
...
+        elif s == "w":
+            return ModelMode.WRITE
+        elif s == "w+":
+            return ModelMode.FORCED_WRITE
+        else:
+            raise ValueError(f"Unknown mode: {s}, options are: r, r+, w, w+")
+
+    def is_writing_mode(self):
+        """Asster whether mode is writing or not."""
+        return self in [ModelMode.WRITE, ModelMode.FORCED_WRITE]
```
hboisgon, line 69: Also add APPEND r+ mode is this list as this potentially will give you write permissions

hboisgon (changes requested):
Hi @savente93 ! I think very nice work on this first implementation and also nice functions added for path and log file handling. I left some comments here and there but my main questions/points are:

- I think we need a clear ``ModelRoot.path`` property as this is a current very used property of ``Model`` (currently ``Model.root``)
- See if we need a ``ModelRoot.set_root`` or if we have two functions ``set_mode`` and ``set_path`` that are called together during init.
- Is there a reason why you added this script and test in _v1 folders rather than directly to the main folders? We have a v1 main branch so I don't think this is needed? May also create some confusion when merging later and may improve our testing of the current functionality if our test bench still works with the new functions?
- Can be done in another PR if you prefer but maybe also start to actually use this in the ``Model`` class? Might allow for extra testing as well and see if we have all we need in ``ModelRoot`` already.
- I see it's still draft so that will maybe still come but we should think of already documenting v1 changes (docs at least API and we can do the content and update of the examples later + changelog)

I'm not the best for Linux/Windows path checking but I see you did extensive tests. Maybe @Jaapel would be a better reviewer for this :)

savente93:
Thanks for the review @hboisgon. I'll adress the comments you made on the code indivdually, but to also quicly respond to your top level comments: 
> I think we need a clear ModelRoot.path property as this is a current very used property of Model (currently Model.root)

Sure, I'll see if this can be workshoped, but if this property is actually changed by the user during runtime it is going to cause issues because these manipulations also have to be handled in the cross platform way. I was thinking that we make the ModelRoot itself a property of the model, so that even `Model.root` can stay as it s, and we use some `@property` magic to make it function just like the root currently does. What do you think? 

> See if we need a ModelRoot.set_root or if we have two functions set_mode and set_path that are called together during init.

I have no strong feelings about this, whatever users prefer I'm happy to implement. 

> Is there a reason why you added this script and test in _v1 folders rather than directly to the main folders? We have a v1 main branch so I don't think this is needed? May also create some confusion when merging later and may improve our testing of the current functionality if our test bench still works with the new functions?

Yes you're right! (see next answer for more detail)

> Can be done in another PR if you prefer but maybe also start to actually use this in the Model class? Might allow for extra testing as well and see if we have all we need in ModelRoot already.

Yes, it definitely should be used in the rest of the code base. The reason that I've done it seperately is so that I can keep refering back to the original implementatoin & tests while working on the new one. Integrating it in the code get's messy fast and is likely to break all tests, so until I got the initial feedback, I wanted to keep it seperate. As soon as we've all agreed that we're happy with this approcah, I'll start integrating it in the main code base. That will definitely happen before this PR is merged. 

> I see it's still draft so that will maybe still come but we should think of already documenting v1 changes (docs at least API and we can do the content and update of the examples later + changelog)

Yes good point, I hadn't thought of that. THat should also happen before this is merged, but I will do that after we agree on the design as well so I don't duplicate a bunch of work. 


Good comments, I hope I adress everything. If I missed anything let me know!

hydromt/models/_v1/model_root.py
```diff
@@ -0,0 +1,187 @@
...
+
+class ModelRoot:
+    """A class to handle model roots in a cross platform manner."""
+
+    def __init__(
+        self,
+        path: PathLike,
+        mode: ModeLike = "w",
+        folders: Optional[List[str]] = None,
+        create_dirs: bool = False,
```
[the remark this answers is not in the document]
savente93: I'm fine with removing the `create_dirs` flag and not creating folders by default, but we should not past lists as default arguments. This is refered to as "Dangerous defaults" in python. Let me know if you want me to explain in more detail but this should not be done.

hydromt/models/_v1/model_root.py
```diff
@@ -0,0 +1,187 @@
...
+        logger: Logger = logger,
+    ):
+        self.logger: Logger = logger
+        self.folders = folders or ["."]
```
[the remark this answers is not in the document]
savente93: sure. I think I just coppied it from the current implementation. I'll remove it. Less attributes means less effort :)

hydromt/models/_v1/model_root.py
```diff
@@ -0,0 +1,187 @@
...
+        if self.is_reading_mode():
+            self._check_root_exists()
+
+    def _check_root_exists(self):
+        # check directory
+        if not isdir(self._path):
+            raise IOError(f'model root not found at "{self._path}"')
+
+    def _setup_folder_structure(self):
+        ignore_ext = [".log", ".yml"]
```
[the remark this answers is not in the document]
savente93: not sure, this was coppied from the current implementation I think. I'll have another look, but if I can't find a good reason I'll remove this.

hydromt/models/_v1/model_root.py
```diff
@@ -0,0 +1,187 @@
...
+                self._setup_log_file_hanglers()
+        if self.is_reading_mode():
+            self._check_root_exists()
+
+    def _check_root_exists(self):
+        # check directory
+        if not isdir(self._path):
+            raise IOError(f'model root not found at "{self._path}"')
+
+    def _setup_folder_structure(self):
```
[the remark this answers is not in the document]
savente93: good point, I'll put that in a seperate function

hydromt/models/_v1/model_root.py
savente93: good point.

savente93:
After a short discussion we decided that the `ModelRoot` should indeed be public to make things easier for the plugin developers

hydromt/models/_v1/model_root.py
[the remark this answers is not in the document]
savente93: Since we decided to take this out of the model root and move it to the model components, I've removed all of these arrtibutes from the Root model. I think we can pick this conversation up again when I start work on the Model Components

hydromt/models/_v1/model_root.py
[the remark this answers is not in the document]
savente93: Now that we got rid of the folders attribute here it cannot check this anymore, it will have to be done by the model components. We do however, check that the model root itself exists

hydromt/models/_v1/model_root.py
```diff
@@ -0,0 +1,115 @@
...
+    """A class to handle model roots in a cross platform manner."""
+
+    def __init__(
+        self,
+        path: PathLike,
+        mode: ModeLike = "w",
+        logger: Logger = logger,
+    ):
+        self.logger: Logger = logger
+        self.set_root(path, mode)
```
Jaapel, line 24: Options to use python-style setters and getters using `@property`. Setting in python generally works with assignments like `model_root.mode = 'w'`

tests/_v1/test_model_root.py
```diff
@@ -0,0 +1,203 @@
...
+            "abs_path": "D:\\data\\test\\model_root",
+        },
+        {
+            "name": "relative",
+            "path": "data\\test\\model_root",
+            "abs_path": join(CURRENT_PATH, "data", "test", "model_root"),
+        },
+    ],
+    ids=case_name,
+)
```
Jaapel, line 41: This is a great usecase for parametrize, as we need to write very similar tests

Jaapel (approved):
LGTM 👍 
I am in favor of using `@property` or using pydantic Models if we need some validation. Otherwise great job

savente93:
oh right, I hadn't even thought about using pydantic for this! that's a really good point!

tests/_v1/test_model_root.py
```diff
@@ -0,0 +1,203 @@
...
+        "ww",
+        "",
+        "+w",
+        "lorum ipsum",
+        1,
+        None,
+        -8,
+        3.14,
+        "⽀",
+        "ðŸ˜Š",
```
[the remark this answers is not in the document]
savente93: fun fact, I am actually a certified tester :P

savente93:
So I know the diff for this PR is a little crazy now, but when I tried to make any changes to the import structure, everything cam down because of cirular imports and the only way to fix that was basically to just do the restructure like we'd been talking about, so I just put on my naughty shoes as we say in Dutch, and just did it. Note that I didn't change any of the code except for renaming imports where necessary, so you don't have to review all the code again. If anyone is unhappy about where I put things, let me know but some choices were forced because of the circular imports. I think that if the structure I went with is not too controversial, then we could just merge and iterate from there which would be my preference, but let me know.

hboisgon:
Wow diff looks crazy indeed ^^

For the re-organization, you are saying you want to merge for now and finalize it in a different PR? I would have a preference to fix it here then all other branches can start from the right organization? Or then tackle this issue ASAP if we do it in another PR.
Anyhow shall we discuss the re-organization in the refinement session this week?

Not fun re-organization means also updating the docs (API for sure) and examples. Talking about docs, shall we also start exporting the v1 docs somewhere for checking? I think @DirkEilander you did something similar when we re-organized the docs also of the plugins (link within the PR to have a preview before merging)

hydromt/models/root.py
```diff
@@ -0,0 +1,66 @@
...
+    """A class to handle model roots in a cross platform manner."""
+
+    def __init__(
+        self,
+        path: PathLike,
+        mode: ModeLike = "w",
+        logger: Logger = logger,
+    ):
+        self.mode = mode
+        self.path = path
```
hboisgon, line 24: I'm not familiar with pydantic and @.setter so just to understand, when calling this line, it's equivalent to calling a eg set_path(path) function but in your case it link to the path setter function?

So in case you want to update the path, how does this work? ModelRoot.path = new_path? And the setter function is re-run?

Why in case of path the setter function returns a Path and for mode nothing?

savente93:
No I think it's preferable to fix it in this PR as well as well as we can, just that we can still iterate on it later if necessary, but hopefully in minor ways. E.g. where some fuctions go, but I definitely want to agree on the structure before we merge. 
I think it's a good idea to discuss this during the refinement.
Updating the examples shouldn't be that big of a problem I can do that when we agree on the structure. 
The docs might be a bit of a bigger job, but it has to happen anyway, but we should agree on a structure first.

hydromt/models/root.py
savente93: This is not anying to do with Pydantic, this is a function of the `@property` decorator. You are correct that `ModelRoot.path = new_path` should be equivalent to running `ModelRoot.set_path(new_path)`. The mode doesn't return anything because that is a mistake of mine, well caught!

hydromt/io/path.py
```diff
@@ -0,0 +1,147 @@
...
+
+    Dealing with paths that describe a file path for a platrom we're not on (e.g. handeling Windows paths on a POSIX system)
+    easily leads to trouble. This function will apply the following assumptions:
+
+    - Paths that describe paths on the same platform are handled like normally
+    - on POSIX platforms a windows path like C:\\to\\somehwere maps to /mnt/c/to/somewhere. for linux this is the cannonical place to mount things like network shares.
+    - Relative Windows paths on POSIX platforms are simply converted to their posix/linux equivalent and used like normal
+    - on Windows a POSIX path like /mnt/d/to/somewhere will map to D:\\to\\somewhere mostly as an inverse as the previous point.
+    - A path like /home/user/to/somewhere will map to C:\\Users\\{login_name_of_user}\\to\\somewhere. so keep in mind it will map to the
+        home dir of the current user, not the user specified in the path.
```
DirkEilander, line 48: Do I understand correctly that this function allows a user on linux to provide a windows-like path which is transformed to a linux path based on conventions of how folders on these platforms art typically organized? 

To be frank, I don't yet see the case for this in relation to a model root. A user should, in my opinion, simply supply a path that works on the local platform they are working on. Can you motivate the case a bit more?

For a data catalog this is different. We now have a single root per data catalog file and hence the same catalog file will not work on multiple OS platforms. But rather than automatically transforming the path I would be in favor of implementing a mechanism to support either overwriting the root that is specified in the file or setting a mapping/list of multiple roots in the file. This would provide more flexibility to where data is stored on different systems. For instance the infamous hydromt p-drive on windows is "p:\wflow_global\hydromt\", if accessed from our hpc clusters "/p/wflow_global/hydromt" and if accessed from wsl we would have to mount it manually and hence the path also depends on how the users organizes their environment (typically this would be "/mnt/p/wflow_global/hydromt" as assumed here).

hydromt/models/root.py
```diff
@@ -0,0 +1,67 @@
...
+        self.logger: Logger = logger
+
+    @property
+    def path(self) -> Path:
+        """The path to the model root in question."""
+        return self._path
+
+    @path.setter
+    def path(self, path: PathLike) -> Path:
```
DirkEilander, line 33: In the current implementation setting the model path also updates the logger filehandler. We have workflows where you read a model, then change the root to a new output path, update and write the model. In that case the logger should, after reading, print to the new output path.

DirkEilander (commented):
- Nice improvements with the `ModelRoot` class, especially with the mode property. I've left a comment on the cross platform paths and a logfile handling. We could also think of a `close()` method, but that can be done later / in a different PR.
- Regarding the repo structure. I would suggest to combine the gis methods (gis_utils, vector, raster, merge, flw) into a single submodule/folder rather than accessor. Note that the raster.py file now appears in two locations (root and accessor). Also where possible it would be really good to keep the file history by renaming rather than removing/adding files. Otherwise this looks good to me.

hydromt/io/path.py
savente93: > To be frank, I don't yet see the case for this in relation to a model root. A user should, in my opinion, simply supply a path that works on the local platform they are working on. Can you motivate the usecase a bit more? E.g. when would a user supply a POSIX path to access a model when working on Windows?

I am the case. We have a data catalog (the deltares data catalog) that provides all windows paths, but I work on linux. If I change the path in the deltares data catalog to a POSIX path then all the windows users will be unhappy. Having to change the path but not commit it every time I want to use it is too much wasted effort. Therefore cross platform teams are the motivating case here. This way you just have to provide a common mount point and the catalog will continue to work. Perhaps everyone at Deltares works on windows but that's not an assumption we should make, since things like cloud environments or docker containers almost exclusively run on POSIX systems and we don't have to have to make different catalogs for the cloud if we even watn to move in that direction. So we need _a_ mechanism for accessing paths on multiple platforms. 

I'm fine with adding the possibility of providing multiple roots or a way to override it, but I have had to spend much to long debugging weird cross platform path bugs so we have to provide SOME sollution for this, even if it's just for my sanity.

DirkEilander: So we agree there is an issue with the deltares DataCatalog root for Linux users. Can you also clarify the issue the model root? Just trying to understand it in order to be able to review.

For the DataCatalog I would be in favor of dealing with this in a more explicit and flexible manner so it does not only support wsl but also the deltares hpc (see my earlier comment) and other potential use cases. Let's make a separate issue for this.

savente93: I'm not actually sure what you're asking, but as you mentioned making a seperate issue for this, I think it would be best if we discuss this face to face and then indeed make a seperate issue. I want to continue iwth the model region as well, so I'll be merging this PR, but we can discuss any changes you like to see next week.

PR #758, opened 2024-01-30
fix COG zoom level bug

### LINKED ISSUE #754: Bug in _parse_zoom_level of rasterdataset.py
### HydroMT version checks

### Reproducible Example

For an arbitrary dataset with overviews, we would like to get the "overview" level close to the desired resolution, e.g.:

```

data_catalog = DataCatalog()
da_fn = r"p:\11203312-012-profews\coastal_flooding\Ethekwini_Flooding\01_data\2020_flt_lidar\dem_1.tif"

# check resolution of original dataset (more or less 1m)
da = data_catalog.get_rasterdataset(da_fn)
da.raster.res

# resolution overview level 0 (is already 1 step coarser, so more or less 2m)
da = data_catalog.get_rasterdataset(da_fn, zoom_level = 0)
da.raster.res

# and here comes the error: 
da = data_catalog.get_rasterdataset(da_fn, zoom_level = (200, "meter")
```

### Current behaviour

At the moment, the overview level that is returned is 1 too high, because the original resolution of the data is parsed as overview_level=0, but according to rasterio, overview_level=0 is already a factor 2 coarses.

This results in errors when you ask for the coarsest level (see example). In this case, it is looking for an overview_level which is not present and therefore rasterio throws an error.

### Desired behaviour

Overviews that are aksed for should be 1 lower, hence overview 0 is 2 times coarser than the original resolution.

### Additional context

_No response_

DirkEilander:
## Issue addressed
Fixes #754

## Explanation
When reading GOCs, the `rioxarray.open_raster(..., overview_level=0)` opens the first overview which is zoom_level 1. To read the native resolution the `overview_level` level argument should not be used, and for actual zoom levels we can use `overview_level = zoom_level - 1`

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,6 +23,7 @@ Fixed
 - Reading Vector formats that consist of more than one file via geopandas. (#691)
 - Handle NoDataStrategy consistently when reading data in adapters (#738)
 - add option to ignore empty data sets when exporting data (#743)
+- Fix bug when reading COGs at requested zoom level (#758)
 
 v0.9.2 (2024-01-09)
 ===================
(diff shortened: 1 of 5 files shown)


savente93 (approved):
LGTM

PR #761, opened 2024-01-31
fix check dims for ds with multiple vars

### LINKED ISSUE #760: raster dimension check fails for datasets of multiple variables with varying dimension size
### HydroMT version checks

### Reproducible Example

raster._check_dimensions currently expect all variables to have the same dimensions.

```
from hydromt import raster
import xarray as xr

da2d = raster.full_from_transform(
        transform=[0.5, 0.0, 3.0, 0.0, -0.5, -9.0],
        shape=(4, 6),
        nodata=-1,
        name="test",
        crs=4326,
)
da3d = da2d.expand_dims("t")
da3d.name = "test_3d"
ds = xr.merge([da2d, da3d])
ds.raster._check_dimensions()
```

### Current behaviour

The example returns a ValueError while the dataset dimension  order is valid:

> ValueError: Invalid dimension order (('y', 'x')). You can use `obj.transpose(('y', 'x')) to reorder your dimensions.

### Desired behaviour

No ValueError for valid datasets

### Additional context

_No response_

DirkEilander:
## Issue addressed
Fixes #760 

## Explanation
Relax assumption that all dataset variables must have the same dimensions.

## Checklist

### DIFF
--- a/hydromt/raster.py
+++ b/hydromt/raster.py
@@ -654,24 +654,42 @@ class XRasterBase(XGeoBase):
     def _check_dimensions(self) -> None:
         """Validate the number and order of dimensions."""
         dims = (self.y_dim, self.x_dim)
-        da = self._obj[self.vars[0]] if isinstance(self._obj, xr.Dataset) else self._obj
-        extra_dims = [dim for dim in da.dims if dim not in dims]
-        if len(extra_dims) == 1:
-            dims = tuple(extra_dims) + dims
-            self.set_attrs(dim0=extra_dims[0])
-        elif len(extra_dims) == 0:
-            self.attrs.pop("dim0", None)
-        elif len(extra_dims) > 1:
-            raise ValueError("Only 2D and 3D data arrays supported.")
-        if isinstance(self._obj, xr.Dataset):
-            check = np.all([self._obj[name].dims == dims for name in self.vars])
+
+        def _check_dims(da: xr.DataArray, dims=dims) -> bool:
+            extra_dims = [dim for dim in da.dims if dim not in dims]
+            if len(extra_dims) == 1:
+                dims = tuple(extra_dims) + dims
+            elif len(extra_dims) > 1:
+                raise ValueError("Only 2D and 3D data arrays supported.")
+            return da.dims == dims, extra_dims
+
+        extra_dims = []
+        if isinstance(self._obj, xr.DataArray):
+            check, dim0 = _check_dims(self._obj)
+            extra_dims.extend(dim0)
+            if not check:
+                raise ValueError(
+                    f"Invalid dimension order ({self._obj.dims}). "
+                    f"You can use `obj.transpose({dims}) to reorder your dimensions."
+                )
         else:
-            check = self._obj.dims == dims
-        if check == False:
-            raise ValueError(
-                f"Invalid dimension order ({da.dims}). "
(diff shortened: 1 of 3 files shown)


hydromt/raster.py
```diff
@@ -654,24 +654,42 @@ def origin(self) -> tuple[float, float]:
...
-        da = self._obj[self.vars[0]] if isinstance(self._obj, xr.Dataset) else self._obj
-        extra_dims = [dim for dim in da.dims if dim not in dims]
+
+        def _check_dims(da: xr.DataArray, dims=dims) -> bool:
+            extra_dims = [dim for dim in da.dims if dim not in dims]
+            if len(extra_dims) == 1:
+                dims = tuple(extra_dims) + dims
+            elif len(extra_dims) > 1:
+                raise ValueError("Only 2D and 3D data arrays supported.")
+            return da.dims == dims, extra_dims
```
savente93, line 664: Please don't use closures if they are not necessary, they can cause very unexpected behavour. Could you also give this function  a more descriptive name that includes what it actually checks?

savente93: Also your type hint is incorrect, it returns a bool and I think a list of dimentions?

hydromt/raster.py
```diff
@@ -654,24 +654,42 @@ def origin(self) -> tuple[float, float]:
...
+
+        extra_dims = []
+        if isinstance(self._obj, xr.DataArray):
+            check, dim0 = _check_dims(self._obj)
```
savente93, line 668: I think this name is very deceptive, since it's not just one dimention but a list of them.

hboisgon (approved):
LGTM!

hydromt/raster.py
```diff
@@ -654,24 +654,42 @@ def origin(self) -> tuple[float, float]:
...
+            extra_dims.extend(dim0)
```
savente93, line 669: Do we want to extend here even if we raise a value error on the next line?

hydromt/raster.py
```diff
@@ -654,24 +654,42 @@ def origin(self) -> tuple[float, float]:
...
+            if not check:
+                raise ValueError(
+                    f"Invalid dimension order ({self._obj.dims}). "
+                    f"You can use `obj.transpose({dims}) to reorder your dimensions."
+                )
+        else:
+            for var in self._obj.data_vars:
+                if not all([dim in self._obj[var].dims for dim in dims]):
+                    continue  # only check data variables with x and y dims
```
savente93, line 678: Does this check data vars that _have_ x and y dims or that _only have_ x and y dims?

savente93 (changes requested):
functionality seems okay from what I can tell but I'd like to see a few code changes.

hydromt/raster.py
DirkEilander: That check happens when getting the x and y dims in the raster.x_dim and raster.y_dim properties.

hydromt/raster.py
DirkEilander: Only when no error is raised we actually use extra_dims, so I don't think this matters

hydromt/raster.py
DirkEilander: will change the name

hydromt/raster.py
DirkEilander: good suggestion

PR #769, opened 2024-02-07
ModelComponent Class

### LINKED ISSUE #671: New `ModelComponent` class
### Discussed in https://github.com/Deltares/hydromt/discussions/636

<div type='discussions-op-text'>

<sup>Originally posted by **DirkEilander** November  7, 2023</sup>
## Current implementation
Currently model components, such as `maps`, `geoms`, `grid`, `mesh`, etc. are defined as properties of the `Model` class or extensions such as the `GridMixin` or `MeshMixin`. All these properties have their own associated `read_`, `write_`, and `set_` methods and in some case some additional generic `setup_` methods. This is perhaps too **restrictive** and requires quite some **duplication of code**. 

It is **restrictive** in the sense that all models must use this terminology and squeeze the model data into these objects. In practice however, models may have multiple of the same components (e.g. SFINCS has a computational grid and subgrid) or model specific components are more meaningful to users (e.g. exposure and vulnerability model component classes in FIAT rather than geoms and tables). 

The current approach also requires **duplication of code** as some components, e.g. `maps`, `forcing`, and `states` are basically the same component (all dict of xarray objects) and share a lot of the code which is now partly circumvented by calling methods defined in the workflows.

## Discussed implementation
- Create common `ModelComponent` _template_ class with abstract `read`, `write`, and `set` methods (and perhaps more) and a dynamic `data` property (not abstract) which reads the data when called for the first time (similar to e.g. current `GridModel.grid` property). Additonally, we should make subclasses for `Grid`, `Mesh`, `Geoms`, `Maps`, `Table`, etc.  The generic `setup_*` methods should also be implemented here. 
- These model components will be implemented as instance attributes of subclasses of Model (e.g. in plugins) and should be registered in a separate class attribute `model_components` that can be used in the `Model.read` and `Model.write` classes to loop over and read/write all components. This would defy the need for the GridMixin etc. 
- Where needed plugins can subclass a `ModelComponent` (sub)class to replace methods with their own methods and add more.
- Syncing data between `Model` and `modelComponent` is done using `weakref` library. 
- This also potentially solves the naming discussion in #475. Instead of `GridModel.setup_grid_data_from_raster` it would `GridModel.grid.add_data_from_raster`, etc.
- We should be able to call these model component methods from the yaml hydromt configuration files, see e.g. below. This requires a small change in the `Model._run_log_method`
- We should review the ``Model`` class and remove hidden/non clear behavior (eg `Model._FOLDERS` used in `Model.set_root`)
- Note that there is a separate issue discussing the `ModelRoot` class #728 
```yml
grid.create:
  arg: value
```

**Example code implementation**

```py
#%% test ModelComponent class 
import weakref
from pathlib import Path

class ModelComponent:
    def __init__(self, model):
        self._model_ref = weakref.ref(model)
        self._data = None

    @property
    def model(self):
        # Access the Model instance through the weak reference
        return self._model_ref()  

    @property
    def data(self):
        return self._data
    
    def set(self, value):
        self._data = value

class ModelRoot:
    def __init__(self, path, mode='r'):
        self.set(path, mode)

    def set(self, path, mode='r'):
        self.path = Path(path)
        self.mode = mode

    def __repr__(self):
        return f"ModelRoot(path={self.path}, mode={self.mode})"

class Model:
    components = {'grid': ModelComponent, 'forcing': ModelComponent}

    def __init__(self, root, mode='r', data_libs=[]):
        self.root = ModelRoot(root, mode)

        # initialize components with linked model instance
        for name, component in self.components.items():
            setattr(self, name, component(model=self))

    def __repr__(self):
        return f"Model(root={self.root})"

# access model property from component
mod = Model('test', mode='r')
mod.root.set('test2', mode='w')
assert mod.root == mod.grid.model.root
print(mod.grid.model.root)
# >>> ModelRoot(path=test2, mode=w)

## acces one component from another
mod.grid.model.forcing.set('test')
print('forcing data:', mod.forcing.data)
# >>> forcing data: test

# infinite instance.attribute.instance.attribute stack .. ?
print(mod.grid.model.forcing.model.grid.model.forcing.model)
# >>> Model(root=ModelRoot(path=test2, mode=w))
```

Tjalling-dejong:
## Issue addressed
Fixes #671 

## Explanation
Creating a model component subclass for grid model first. The idea is to make a general ModelComponent class that will replace the different model classess with subclasses of ModelComponent. 

## Checklist

### DIFF
--- a/hydromt/models/v1/grid.py
+++ b/hydromt/models/v1/grid.py
@@ -1,0 +1,458 @@
+"""Grid ModelComponent."""
+import logging
+import weakref
+from pathlib import Path
+from typing import Dict, List, Optional, Tuple, Union
+
+import geopandas as gpd
+import numpy as np
+import pandas as pd
+import xarray as xr
+from pyproj import CRS
+from shapely.geometry import box
+
+from hydromt.data_catalog import DataCatalog
+
+from .. import workflows
+
+logger = logging.getLogger(__name__)
+
+
+class GridModelComponent:
+    """GridModelComponent class."""
+
+    def __init__(self, model, logger):
+        self._model_ref = weakref.ref(model)
+        self._data = None
+        self.logger = logger
+
+    def set(
+        self,
+        data: Union[xr.DataArray, xr.Dataset, np.ndarray],
+        name: Optional[str] = None,
+        read: bool = True,
+    ):
+        """Add data to grid.
+
+        All layers of grid must have identical spatial coordinates.
+
+        Parameters
+        ----------
+        data: xarray.DataArray or xarray.Dataset
+            new map layer to add to grid
+        name: str, optional
+            Name of new map layer, this is used to overwrite the name of a DataArray
+            and ignored if data is a Dataset
+        """
+        self._initialize_grid()
+        # NOTE: variables in a dataset are not longer renamed as used to be the case in
+        # set_staticmaps
+        name_required = isinstance(data, np.ndarray) or (
+            isinstance(data, xr.DataArray) and data.name is None
+        )
+        if name is None and name_required:
+            raise ValueError(f"Unable to set {type(data).__name__} data without a name")
+        if isinstance(data, np.ndarray):
+            if data.shape != self._data.raster.shape:
+                raise ValueError("Shape of data and grid maps do not match")
+            data = xr.DataArray(dims=self._data.raster.dims, data=data, name=name)
+        if isinstance(data, xr.DataArray):
+            if name is not None:  # rename
+                data.name = name
+            data = data.to_dataset()
+        elif not isinstance(data, xr.Dataset):
+            raise ValueError(f"cannot set data of type {type(data).__name__}")
+
+        if len(self._data) == 0:  # empty grid
+            self._data = data
+        else:
+            for dvar in data.data_vars:
+                if dvar in self._data:
+                    if read:
+                        self.logger.warning(f"Replacing grid map: {dvar}")
+                self._data[dvar] = data[dvar]
+
+    def write(self):
+        """Write grid."""
+        pass
+
+    def read(
+        self, fn: str = "grid/grid.nc", read: bool = True, write: bool = False, **kwargs
+    ) -> None:
+        """Read model grid data at <root>/<fn> and add to grid property.
+
+        key-word arguments are passed to :py:meth:`~hydromt.models.Model.read_nc`
+
+        Parameters
+        ----------
+        fn : str, optional
+            filename relative to model root, by default 'grid/grid.nc'
+        **kwargs : dict
+            Additional keyword arguments to be passed to the `read_nc` method.
+        """
+        self._assert_read_mode()
+        self._initialize_grid(skip_read=True)
+
+        # Load grid data in r+ mode to allow overwritting netcdf files
+        if read and write:
+            kwargs["load"] = True
+        loaded_nc_files = self.read_nc(
+            fn, single_var_as_array=False, **kwargs
+        )  # TODO: decide where read_nc should be placed
+        for ds in loaded_nc_files.values():
+            self.set(ds)
+
+    def create(self):
+        """Create grid."""
+        pass
+
+    @property
+    def model(self):
+        """Returns model reference."""
+        # Access the Model instance through the weak reference
+        return self._model_ref()
+
+    @property
+    def data(self):
+        """Returns data."""
+        return self._data
+
+    @property
+    def res(self) -> Tuple[float, float]:
+        """Returns the resolution of the model grid."""
+        if len(self._data) > 0:
+            return self._data.raster.res
+
+    @property
+    def transform(self):
+        """Returns spatial transform of the model grid."""
+        if len(self._data) > 0:
+            return self._data.raster.transform
+
+    @property
+    def crs(self) -> Union[CRS, None]:
+        """Returns coordinate reference system embedded in the model grid."""
+        if self._data.raster.crs is not None:
+            return CRS(self._data.raster.crs)
+
+    @property
+    def bounds(self) -> List[float]:
+        """Returns the bounding box of the model grid."""
+        if len(self._data) > 0:
+            return self._data.raster.bounds
+
+    @property
+    def region(self) -> gpd.GeoDataFrame:
+        """Returns the geometry of the model area of interest."""
+        region = gpd.GeoDataFrame()
+        if len(self._data) > 0:
+            crs = self.crs
+            if crs is not None and hasattr(crs, "to_epsg"):
+                crs = crs.to_epsg()  # not all CRS have an EPSG code
+            region = gpd.GeoDataFrame(geometry=[box(*self.bounds)], crs=crs)
+        return region
+
+    @property
+    def grid(self):
+        """Model static gridded data as xarray.Dataset."""
+        if self._data is None:
+            self._initialize_grid()
+        return self._data
+
+    def _initialize_grid(self, skip_read: bool = False, read: bool = True) -> None:
+        """Initialize grid object."""
+        if self._data is None:
+            self._data = xr.Dataset()
+            if read and not skip_read:
+                self.read()
+
+    def set_crs(self, crs: CRS) -> None:
+        """Set coordinate reference system of the model grid."""
+        if len(self._data) > 0:
+            self._data.raster.set_crs(crs)
+
+    def setup_grid_from_constant(
+        self,
+        constant: Union[int, float],
+        name: str,
+        dtype: Optional[str] = "float32",
+        nodata: Optional[Union[int, float]] = None,
+        mask_name: Optional[str] = "mask",
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Adds a grid based on a constant value.
+
+        Parameters
+        ----------
+        constant: int, float
+            Constant value to fill grid with.
+        name: str
+            Name of grid.
+        dtype: str, optional
+            Data type of grid. By default 'float32'.
+        nodata: int, float, optional
+            Nodata value. By default infered from dtype.
+        mask_name: str, optional
+            Name of mask in self.grid to use for masking raster_fn. By default 'mask'.
+            Use None to disable masking.
+
+        Returns
+        -------
+        list
+            Names of added model grid layer.
+        """
+        da = workflows.grid.grid_from_constant(
+            grid_like=self._data,
+            constant=constant,
+            name=name,
+            dtype=dtype,
+            nodata=nodata,
+            mask_name=mask_name,
+        )
+        # Add to grid
+        self.set(da)
+
+        return [name]
+
+    def setup_grid_from_rasterdataset(
+        self,
+        raster_fn: Union[str, Path, xr.DataArray, xr.Dataset],
+        data_catalog: DataCatalog,
+        variables: Optional[List] = None,
+        fill_method: Optional[str] = None,
+        reproject_method: Optional[Union[List, str]] = "nearest",
+        mask_name: Optional[str] = "mask",
+        rename: Optional[Dict] = None,
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Add data variable(s) from ``raster_fn`` to grid object.
+
+        If raster is a dataset, all variables will be added unless ``variables`` list
+        is specified.
+
+        Adds model layers:
+
+        * **raster.name** grid: data from raster_fn
+
+        Parameters
+        ----------
+        raster_fn: str, Path, xr.DataArray, xr.Dataset
+            Data catalog key, path to raster file or raster xarray data object.
+            If a path to a raster file is provided it will be added
+            to the data_catalog with its name based on the file basename without
+            extension.
+        variables: list, optional
+            List of variables to add to grid from raster_fn. By default all.
+        fill_method : str, optional
+            If specified, fills nodata values using fill_nodata method.
+            Available methods are {'linear', 'nearest', 'cubic', 'rio_idw'}.
+        reproject_method: list, str, optional
+            See rasterio.warp.reproject for existing methods, by default 'nearest'.
+            Can provide a list corresponding to ``variables``.
+        mask_name: str, optional
+            Name of mask in self.grid to use for masking raster_fn. By default 'mask'.
+            Use None to disable masking.
+        rename: dict, optional
+            Dictionary to rename variable names in raster_fn before adding to grid
+            {'name_in_raster_fn': 'name_in_grid'}. By default empty.
+
+        Returns
+        -------
+        list
+            Names of added model map layers
+        """
+        rename = rename or {}
+        self.logger.info(f"Preparing grid data from raster source {raster_fn}")
+        # Read raster data and select variables
+        ds = data_catalog.get_rasterdataset(
+            raster_fn,
+            geom=self.region,
+            buffer=2,
+            variables=variables,
+            single_var_as_array=False,
+        )
+        # Data resampling
+        ds_out = workflows.grid.grid_from_rasterdataset(
+            grid_like=self._data,
+            ds=ds,
+            variables=variables,
+            fill_method=fill_method,
+            reproject_method=reproject_method,
+            mask_name=mask_name,
+            rename=rename,
+        )
+        # Add to grid
+        self.set(ds_out)
+
+        return list(ds_out.data_vars.keys())
+
+    def setup_grid_from_raster_reclass(
+        self,
+        data_catalog: DataCatalog,
+        raster_fn: Union[str, Path, xr.DataArray],
+        reclass_table_fn: Union[str, Path, pd.DataFrame],
+        reclass_variables: List,
+        variable: Optional[str] = None,
+        fill_method: Optional[str] = None,
+        reproject_method: Optional[Union[List, str]] = "nearest",
+        mask_name: Optional[str] = "mask",
+        rename: Optional[Dict] = None,
+        **kwargs,
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Add data variable(s) to grid object by reclassifying the data in ``raster_fn`` based on ``reclass_table_fn``.
+
+        Adds model layers:
+
+        * **reclass_variables** grid: reclassified raster data
+
+        Parameters
+        ----------
+        raster_fn: str, Path, xr.DataArray
+            Data catalog key, path to raster file or raster xarray data object.
+            Should be a DataArray. Else use `variable` argument for selection.
+        reclass_table_fn: str, Path, pd.DataFrame
+            Data catalog key, path to tabular data file or tabular pandas dataframe
+            object for the reclassification table of `raster_fn`.
+        reclass_variables: list
+            List of reclass_variables from reclass_table_fn table to add to maps.
+            Index column should match values in `raster_fn`.
+        variable: str, optional
+            Name of raster_fn dataset variable to use. This is only required when
+            reading datasets with multiple variables.
+            By default None.
+        fill_method : str, optional
+            If specified, fills nodata values in `raster_fn` using fill_nodata method
+            before reclassifying. Available methods are
+            {'linear', 'nearest', 'cubic', 'rio_idw'}.
+        reproject_method: str, optional
+            See rasterio.warp.reproject for existing methods, by default "nearest".
+            Can provide a list corresponding to ``reclass_variables``.
+        mask_name: str, optional
+            Name of mask in self.grid to use for masking raster_fn. By default 'mask'.
+            Use None to disable masking.
+        rename: dict, optional
+            Dictionary to rename variable names in reclass_variables before adding to
+            grid {'name_in_reclass_table': 'name_in_grid'}. By default empty.
+
+        **kwargs : dict
+            Additional keyword arguments to be passed to `get_rasterdataset`
+
+        Returns
+        -------
+        list
+            Names of added model grid layers
+        """  # noqa: E501
+        rename = rename or dict()
+        self.logger.info(
+            f"Preparing grid data by reclassifying the data in {raster_fn} based "
+            f"on {reclass_table_fn}"
+        )
+        # Read raster data and remapping table
+        da = data_catalog.get_rasterdataset(
+            raster_fn, geom=self.region, buffer=2, variables=variable, **kwargs
+        )
+        if not isinstance(da, xr.DataArray):
+            raise ValueError(
+                f"raster_fn {raster_fn} should be a single variable. "
+                "Please select one using the 'variable' argument"
+            )
+        df_vars = data_catalog.get_dataframe(
+            reclass_table_fn, variables=reclass_variables
+        )
+        # Data resampling
+        ds_vars = workflows.grid.grid_from_raster_reclass(
+            grid_like=self._data,
+            da=da,
+            reclass_table=df_vars,
+            reclass_variables=reclass_variables,
+            fill_method=fill_method,
+            reproject_method=reproject_method,
+            mask_name=mask_name,
+            rename=rename,
+        )
+        # Add to maps
+        self.set(ds_vars)
+
+        return list(ds_vars.data_vars.keys())
+
+    def setup_grid_from_geodataframe(
+        self,
+        data_catalog: DataCatalog,
+        vector_fn: Union[str, Path, gpd.GeoDataFrame],
+        variables: Optional[Union[List, str]] = None,
+        nodata: Optional[Union[List, int, float]] = -1,
+        rasterize_method: Optional[str] = "value",
+        mask_name: Optional[str] = "mask",
+        rename: Optional[Dict] = None,
+        all_touched: Optional[bool] = True,
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Add data variable(s) to grid object by rasterizing the data from ``vector_fn``.
+
+        Several type of rasterization are possible:
+            * "fraction": the fraction of the grid cell covered by the vector
+                shape is returned.
+            * "area": the area of the grid cell covered by the vector shape is returned.
+            * "value": the value from the variables columns of vector_fn are used.
+                If this is used, variables must be specified.
+
+        Parameters
+        ----------
+        vector_fn : str, Path, gpd.GeoDataFrame
+            Data catalog key, path to vector file or a vector geopandas object.
+        variables : List, str, optional
+            List of variables to add to grid from vector_fn. Required if
+            rasterize_method is "value", by default None.
+        nodata : List, int, float, optional
+            No data value to use for rasterization, by default -1. If a list is
+            provided, it should have the same length has variables.
+        rasterize_method : str, optional
+            Method to rasterize the vector data. Either {"value", "fraction", "area"}.
+            If "value", the value from the variables columns in vector_fn are used
+            directly in the raster. If "fraction", the fraction of the grid
+            cell covered by the vector file is returned. If "area", the area of the
+            grid cell covered by the vector file is returned.
+        mask_name: str, optional
+            Name of mask in self.grid to use for masking raster_fn. By default 'mask'.
+            Use None to disable masking.
+        rename: dict, optional
+            Dictionary to rename variable names in variables before adding to grid
+            {'name_in_variables': 'name_in_grid'}. To rename with method fraction or
+            area use {'vector_fn': 'name_in_grid'}. By default empty.
+        all_touched : bool, optional
+            If True (default), all pixels touched by geometries will be burned in.
+            If false, only pixels whose center is within the polygon or that are
+            selected by Bresenham's line algorithm will be burned in.
+
+        Returns
+        -------
+        list
+            Names of added model grid layers
+        """  # noqa: E501
+        rename = rename or dict()
+        self.logger.info(f"Preparing grid data from vector '{vector_fn}'.")
+        gdf = data_catalog.get_geodataframe(
+            vector_fn, geom=self.region, dst_crs=self.crs
+        )
+        if gdf.empty:
+            self.logger.warning(
+                f"No shapes of {vector_fn} found within region,"
+                " skipping setup_grid_from_vector."
+            )
+            return
+        # Data resampling
+        if vector_fn in rename.keys():
+            # In case of choosing a new name with area or fraction method pass
+            # the name directly
+            rename = rename[vector_fn]
+        ds = workflows.grid.grid_from_geodataframe(
+            grid_like=self._data,
+            gdf=gdf,
+            variables=variables,
+            nodata=nodata,
+            rasterize_method=rasterize_method,
+            mask_name=mask_name,
+            rename=rename,
+            all_touched=all_touched,
+        )
+        # Add to grid
+        self.set(ds)
+
+        return list(ds.data_vars.keys())


hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+            by default False
+        rename_dims: bool, optional
+            If True and gdal_compliant, rename x_dim and y_dim to standard names
+            depending on the CRS (x/y for projected and lat/lon for geographic).
+        force_sn: bool, optional
+            If True and gdal_compliant, forces the dataset to have
+            South -> North orientation.
+        """
+        if len(self._data) == 0:
+            self.logger.debug("No grid data found, skip writing.")
```
savente93, line 105: THis is probably a good place for the `NoDataStrategy`

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+        )
+        if name is None and name_required:
+            raise ValueError(f"Unable to set {type(data).__name__} data without a name")
+        if isinstance(data, np.ndarray):
+            if data.shape != self._data.raster.shape:
+                raise ValueError("Shape of data and grid maps do not match")
+            data = xr.DataArray(dims=self._data.raster.dims, data=data, name=name)
+        if isinstance(data, xr.DataArray):
+            if name is not None:  # rename
+                data.name = name
```
savente93, line 62: can't you do this when you check `name_required`?

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+            data = data.to_dataset()
+        elif not isinstance(data, xr.Dataset):
+            raise ValueError(f"cannot set data of type {type(data).__name__}")
+
+        if len(self._data) == 0:  # empty grid
+            self._data = data
```
savente93, line 68: maybe use `NoDataStrategy` here?

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+        else:
+            for dvar in data.data_vars:
+                if dvar in self._data:
+                    if read:
+                        self.logger.warning(f"Replacing grid map: {dvar}")
+                self._data[dvar] = data[dvar]
+
+    def write(
+        self,
+        fn: str = "grid/grid.nc",
```
savente93, line 78: is this sting supposed to be a `Path` or a data catalog entry?

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+            depending on the CRS (x/y for projected and lat/lon for geographic).
+        force_sn: bool, optional
+            If True and gdal_compliant, forces the dataset to have
+            South -> North orientation.
+        """
+        if len(self._data) == 0:
+            self.logger.debug("No grid data found, skip writing.")
+        else:
+            self._assert_write_mode()  # Model class method
+            # write_nc requires dict - use dummy 'grid' key
```
savente93, line 108: we should look into if this is still necessary

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+            self.write_nc(
+                {"grid": self._data},
```
savente93, line 110: perhaps name is a better key to use? because if we have multiple grid components this is going to overwrite previous ones. (I presume)

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+                fn,
+                gdal_compliant=gdal_compliant,
+                rename_dims=rename_dims,
+                force_sn=force_sn,
+                **kwargs,
+            )
+
+    def read(
+        self, fn: str = "grid/grid.nc", read: bool = True, write: bool = False, **kwargs
```
savente93, line 119: is this `fn` supposed to be a data catalog entry or a Path?

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+    def crs(self) -> Union[CRS, None]:
+        """Returns coordinate reference system embedded in the model grid."""
+        if self._data.raster.crs is not None:
+            return CRS(self._data.raster.crs)
+
+    @property
+    def bounds(self) -> List[float]:
+        """Returns the bounding box of the model grid."""
+        if len(self._data) > 0:
+            return self._data.raster.bounds
```
savente93, line 181: We need to do something here when there is no data to return. Also your type arguments don't reflect that the functions can return None

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+    @property
+    def model(self):
+        """Returns model reference."""
+        # Access the Model instance through the weak reference TODO: discuss if this is needed
+        return self._model_ref()
+
+    @property
+    def data(self):
+        """Returns data."""
+        return self._data
```
savente93, line 157: we're not doing anything special like lazy loading to these ones so I wonder if these need to be properties.

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,722 @@
...
+    def set_crs(self, crs: CRS) -> None:
+        """Set coordinate reference system of the model grid."""
+        if len(self._data) > 0:
+            self._data.raster.set_crs(crs)
+
+    def setup_grid_from_constant(
+        self,
+        constant: Union[int, float],
+        name: str,
+        dtype: Optional[str] = "float32",
```
savente93, line 217: If you use `np.dtype` here we can get some better help from the compiler.

savente93 (commented):
Looking good Tjalling. I think this is nice for a first try and we'll see about polishing some things when we start integrating it with the rest and moving over tests.

hydromt/models/v1/grid.py
Tjalling-dejong: I did not change this part of the original grid model code. I'm not sure if we should change since I assume it was working before.

hydromt/models/v1/grid.py
Tjalling-dejong: We could use the `NoDataStrategy`

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,749 @@
...
+            _exec_nodata_strat(
+                msg="No grid data found, skip writing.",
+                strategy=NoDataStrategy.IGNORE,
+                logger=logger,
+            )
+        else:
+            _assert_write_mode(write)
+            # write_nc requires dict - use dummy 'grid' key
+            return write_nc(
+                {"grid": self._data},  # TODO: use a name argument or fn as key
```
hboisgon, line 125: Not needed to use a name argument or fn as key. The "grid" name is only temporary used for the write_nc function, is not appearing anywhere in the output file name or properties and disappears as soon as write_nc is done running :)

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,749 @@
...
+            self._data = xr.Dataset()
+            if read and not skip_read:
+                self.read()
+
+    def set_crs(self, crs: CRS) -> None:
+        """Set coordinate reference system of the model grid."""
+        if len(self._data) > 0:
+            self._data.raster.set_crs(crs)
+
+    def setup_grid_from_constant(
```
hboisgon, line 240: We discussed in a new issue we wanted to change these names. Here we could rename to add_data_from_constant

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,749 @@
...
+            dtype=dtype,
+            nodata=nodata,
+            mask_name=mask_name,
+        )
+        # Add to grid
+        self.set(da)
+
+        return [name]
+
+    def setup_grid_from_rasterdataset(
```
hboisgon, line 282: Rename to add_data_from_rasterdataset

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,749 @@
...
+            reproject_method=reproject_method,
+            mask_name=mask_name,
+            rename=rename,
+        )
+        # Add to grid
+        self.set(ds_out)
+
+        return list(ds_out.data_vars.keys())
+
+    def setup_grid_from_raster_reclass(
```
hboisgon, line 353: Rename to add_data_from_raster_reclass

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,749 @@
...
+            reproject_method=reproject_method,
+            mask_name=mask_name,
+            rename=rename,
+        )
+        # Add to maps
+        self.set(ds_vars)
+
+        return list(ds_vars.data_vars.keys())
+
+    def setup_grid_from_geodataframe(
```
hboisgon, line 442: rename to add_data_from_geodataframe

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,749 @@
...
+            mask_name=mask_name,
+            rename=rename,
+            all_touched=all_touched,
+        )
+        # Add to grid
+        self.set(ds)
+
+        return list(ds.data_vars.keys())
+
+    def setup_grid(
```
hboisgon, line 526: This is actually your create function above which for now you have RaiseNotImplemented :)

tests/model/model_components/test_grid_model_componet.py
hboisgon, line 1: Typo in the title :)
In a way all tests also for the current setup methods already exist for the GridModel so let's check those out when we start integrating the ModelComponent to Model in a next phase

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,749 @@
...
+    _assert_read_mode,
+    _assert_write_mode,
+    read_nc,
+    write_nc,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class GridModelComponent(ModelComponent):
```
hboisgon, line 32: GridModelComponent or GridComponent actually? It's up for discussion but I would maybe prefer the second option?

hboisgon (commented):
Hi @Tjalling-dejong , very nice first implementation of the Grid ModelComponent!
I think it's almost ready, I just added a few answers to Sam and asked to rename the setup methods but for the rest almost ready to merge I think :)
Just one thing, but before you merge, I think we don't need the v1 folder and maybe rename grid.py script to grid_component.py

hydromt/models/v1/grid.py
```diff
@@ -0,0 +1,749 @@
...
+    """GridModelComponent class."""
+
+    def __init__(self, root, logger=logger):
+        # self._model_ref = weakref.ref(model) TODO: discuss if this is necessary
```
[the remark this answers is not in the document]
DirkEilander: We need to pass model rather than root as we also want to be able to access other properties such as region, datacatalog, etc. See also discussion on model components  https://github.com/Deltares/hydromt/discussions/636

DirkEilander (commented):
Some quick suggestions:
- we need to initialize a component with the `model` rather than just `root` to be able to acces the model region, data catalog, etc.. Using a model property that calls the weak_ref to model we can avoid circular dependencies, see also discussion https://github.com/Deltares/hydromt/discussions/636
- it would be good to keep the 'just in time reading' we have currently. This is achieved by creating a `data` property which calls `_initialize_data` if `_data is None`. In most other methods we can than simply use `data` rather than `_data`, see also current implementation for grid, maps, geoms, etc.
- Both the `model` and `data` properties should land in the generic `ModelComponent` class

hydromt/models/v1/grid.py
savente93: Okay, if you think it's not necessary to use here then we don't need it

hydromt/models/v1/grid.py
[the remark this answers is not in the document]
Tjalling-dejong: I think it would reduce complexity if we initialize a model component with model,root, model.region, model.datacatalog etc., instead of passing the model object. This way the model object and model component object are kept separate.

hydromt/models/v1/grid.py
Tjalling-dejong: I agree, GridComponent is a better option.

hydromt/models/v1/grid.py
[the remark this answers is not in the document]
DirkEilander: In the discussion that's also where we started. However, we found that other properties of model might also be needed by some components and in the future new properties. Passing model with a weak_ref works for all and can therefore be implemented in a general way for all components and avoid potential breaking changes in future versions if new properties are required.

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,165 @@
...
+            data = xr.DataArray(dims=self._data.raster.dims, data=data, name=name)
+        elif isinstance(data, xr.DataArray):
+            if name is not None:
+                data.name = name
+            data = data.to_dataset()
+        elif not isinstance(data, xr.Dataset):
+            raise ValueError(f"cannot set data of type {type(data).__name__}")
 
-        return list(ds_out.data_vars.keys())
+        if len(self._data) == 0:  # empty grid
```
savente93, line 109: could `self._data` be none at this point?

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,165 @@
...
+        fn: str = "grid/grid.nc",
+        gdal_compliant: bool = False,
+        rename_dims: bool = False,
+        force_sn: bool = False,
         **kwargs,
-    ) -> List[str]:
-        """HYDROMT CORE METHOD: Add data variable(s) to grid object by reclassifying the data in ``raster_fn`` based on ``reclass_table_fn``.
-
-        Adds model layers:
+    ) -> DeferedFileClose | None:
```
savente93, line 125: python 3.10 syntax. please use 3.9

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,165 @@
...
-        vector_fn: Union[str, Path, gpd.GeoDataFrame],
-        variables: Optional[Union[List, str]] = None,
-        nodata: Optional[Union[List, int, float]] = -1,
-        rasterize_method: Optional[str] = "value",
-        mask_name: Optional[str] = "mask",
-        rename: Optional[Dict] = None,
-        all_touched: Optional[bool] = True,
-    ) -> List[str]:
-        """HYDROMT CORE METHOD: Add data variable(s) to grid object by rasterizing the data from ``vector_fn``.
+        fn: str = "grid/grid.nc",
```
savente93, line 167: definitely something that we should have a larger discussion about but I'm really not fond of this abbreviation, not least of which because it's not a filename and it can also be many things besides a path, such as the actual data or a catalog entry. (at least it could be in the original implementation) Can we at least pick something more descriptive like "rel_path" in the case of string/path or "data_like" or anything that hints at all the things it could be?

hydromt/models/model_utils.py
```diff
@@ -0,0 +1,144 @@
...
+
+def read_nc(
+    fn: StrPath,
+    root,
+    logger: Logger,
+    mask_and_scale: bool = False,
+    single_var_as_array: bool = True,
+    load: bool = False,
+    **kwargs,
+) -> Dict[str, xr.Dataset]:
```
savente93, line 23: This should more to the IO submodule

hydromt/models/model_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    -------
+    Dict[str, xr.Dataset]
+        dict of xarray.Dataset
+    """
+    ncs = dict()
+    fns = glob.glob(join(root, fn))
+    if "chunks" not in kwargs:  # read lazy by default
+        kwargs.update(chunks="auto")
+    for fn in fns:
+        name = basename(fn).split(".")[0]
```
savente93, line 59: please use `splitext` instead. if there are multiple periods in the file name this will break I think.

hydromt/models/model_utils.py
```diff
@@ -0,0 +1,144 @@
...
+            ds = xr.open_dataset(fn, mask_and_scale=mask_and_scale, **kwargs).load()
+            ds.close()
+        else:
+            ds = xr.open_dataset(fn, mask_and_scale=mask_and_scale, **kwargs)
+        # set geo coord if present as coordinate of dataset
+        if GEO_MAP_COORD in ds.data_vars:
+            ds = ds.set_coords(GEO_MAP_COORD)
+        # single-variable Dataset to DataArray
+        if single_var_as_array and len(ds.data_vars) == 1:
+            (ds,) = ds.data_vars.values()
```
savente93, line 72: why the tuple here?

hydromt/models/model_utils.py
```diff
@@ -0,0 +1,144 @@
...
+        ncs.update({name: ds})
+    return ncs
+
+
+def write_nc(
```
savente93, line 77: should also move to IO submodule.

hydromt/models/model_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    nc_dict: XArrayDict,
+    fn: str,
+    root,
+    logger: Logger,
+    temp_data_dir=None,
+    gdal_compliant: bool = False,
+    rename_dims: bool = False,
+    force_sn: bool = False,
+    **kwargs,
+) -> DeferedFileClose | None:
```
savente93, line 87: 3.10 syntax

hydromt/models/model_utils.py
```diff
@@ -0,0 +1,144 @@
...
+    force_sn: bool, optional
+        If True, forces the dataset to have South -> North orientation. Only used
+        if ``gdal_compliant`` is set to True. By default, False.
+    **kwargs:
+        Additional keyword arguments that are passed to the `to_netcdf`
+        function.
+    """
+    for name, ds in nc_dict.items():
+        if not isinstance(ds, (xr.Dataset, xr.DataArray)) or len(ds) == 0:
+            logger.error(f"{name} object of type {type(ds).__name__} not recognized")
```
savente93, line 121: should this be `self.logger`? currently you're always using the logger defined at the top instead of one the user provides

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,165 @@
...
+        """
+        if len(self.data) == 0:
+            _exec_nodata_strat(
+                msg="No grid data found, skip writing.",
+                strategy=NoDataStrategy.IGNORE,
+                logger=logger,
+            )
+        else:
+            if not self.root.is_writing_mode():
+                raise IOError("Model opened in read-only mode")
```
savente93, line 154: not exactly sure if this is the right place to check this. We can definitely keep this but we'll also have to check at some point whether files exist and whether we're in overwrite mode. If you did this else where my bad, but I didn't think I saw that.

hydromt/models/components/grid.py
```diff
@@ -386,85 +178,20 @@ def read_grid(self, fn: str = "grid/grid.nc", **kwargs) -> None:
         **kwargs : dict
             Additional keyword arguments to be passed to the `read_nc` method.
         """
-        self._assert_read_mode()
+        if not self.root.is_reading_mode():
+            raise IOError("Model opened in write-only mode")
         self._initialize_grid(skip_read=True)
 
         # Load grid data in r+ mode to allow overwritting netcdf files
```
savente93, line 185: `r+` does not overwrite, but append. is this what we want?

hydromt/models/components/grid.py
```diff
@@ -684,98 +411,362 @@ def setup_grid(
...
-        super().write(components=components)
-
-    # Properties for subclass GridModel
     @property
-    def res(self) -> Tuple[float, float]:
+    def res(self) -> Tuple[float, float] | None:
         """Returns the resolution of the model grid."""
-        if len(self.grid) > 0:
-            return self.grid.raster.res
+        if len(self.data) > 0:
```
savente93, line 421: cant `self.data` be None here?

tests/model/model_components/test_grid_component.py
```diff
@@ -0,0 +1,334 @@
...
+    grid_component = GridComponent(
+        root=model_root, data_catalog=data_catalog, model_region=None
+    )
+    # Test setting xr.Dataset
+    grid_component.set(data=hydds)
+    assert len(grid_component.data) > 0
+    assert isinstance(grid_component.data, xr.Dataset)
+    # Test setting xr.DataArray
+    data_array = hydds.to_array()
+    grid_component.set(data=data_array, name="data_array")
```
savente93, line 31: Could you make these into seperate tests please?

tests/model/model_components/test_grid_component.py
```diff
@@ -0,0 +1,334 @@
...
+    with pytest.raises(IOError, match="Model opened in write-only mode"):
+        grid_component.read()
+    model_root = ModelRoot(path=tmp_dir, mode="r+")
+    data_catalog = DataCatalog()
+    grid_component = GridComponent(
+        root=model_root, data_catalog=data_catalog, model_region=None, logger=logger
+    )
+    with patch("hydromt.models.components.grid.read_nc", return_value={"grid": hydds}):
+        grid_component.read()
+        assert grid_component.data == hydds
```
savente93, line 83: nice!

savente93 (commented):
Very nice work Tjalling, I especially like the amount of testing. I made a couple of small comments, not all of which are for you to resolve (namely the `fn` discussion) but overall I think this is good

hydromt/models/components/grid.py
deltamarnix: Perhaps we should make the default pixi environment 3.9

savente93: I'd be in favor of that if it doesn't upset people

hydromt/models/components/grid.py
Tjalling-dejong: I agree. I changed it data_like for now, but maybe good to discuss this and decide if we need to have a standardized name across the code base for these arguments.

hydromt/models/components/grid.py
Tjalling-dejong: Not sure, I did not change this code. Maybe @DirkEilander knows?

hydromt/models/components/grid.py
Tjalling-dejong: No it will be an empty xarray dataset if there is no data

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,163 @@
...
-        constant: Union[int, float],
-        name: str,
-        dtype: Optional[str] = "float32",
-        nodata: Optional[Union[int, float]] = None,
-        mask_name: Optional[str] = "mask",
-    ) -> List[str]:
-        """HYDROMT CORE METHOD: Adds a grid based on a constant value.
+        root: ModelRoot,
+        data_catalog: DataCatalog,
+        model_region=None,
```
deltamarnix, line 47: Please add typing

hydromt/models/components/model_component.py
```diff
@@ -0,0 +1,22 @@
...
+
+    @abstractmethod
+    def write(self):
+        """Write data."""
+        pass
+
+    @abstractmethod
+    def read(self):
+        """Read data."""
+        pass
```
deltamarnix, line 22: All these functions don't seem to be implemented by GridComponent at all, because they have different parameters. We should consider how we want to implement them.

We either need to go for a dynamic solution with `**kwargs`, or make an `initialize(self, foo, bar, baz)` that is not in the base class ModelComponent, or we make all setters per property: `GridComponent.model_root = X`, `GridComponent.region = Y`.

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,163 @@
...
-
-    def setup_grid_from_rasterdataset(
+        # self._model_ref = weakref.ref(model) TODO: discuss if this is necessary
+        self.logger = logger
+        self.root = root
+        self.data_catalog = data_catalog
+        self.model_region = model_region
+        self._data = None
+
+    def set(
```
deltamarnix, line 70: I am wondering if this function will be called from outside the class. Perhaps it's better to make this function private and add the `create` function to the base class.

hydromt/models/components/grid.py
```diff
@@ -683,99 +408,363 @@ def setup_grid(
             grid = grid.drop_vars("mask")
 
         # Add region and grid to model
-        self.set_geoms(geom, "region")
-        self.set_grid(grid)
+        # self.set_geoms(geom, "region") TODO: needs implemenation of ModelRegion
```
hboisgon, line 411: Which is where you could need the weakref to model: self.model.region.set(geom)
We discussed with @savente93 that we will see how to keep region updated based on model main_component once we merge this PR and the ModelRegion PR together.

hydromt/models/components/model_component.py
hboisgon: I think we need to add more functions here especially the __init__ with a weakref to model. Suggestion from @DirkEilander  in PR #810 

@deltamarnix: is there an issue with using abstractmethod for methods that the subclasses MUST have (here set, read, write are musts).
We can check afterwards when merging together with ModelRoot if we need extra properties as we have yet some unresolved issues there before we can check.

``` python
class ModelComponent:

    def __init__(self, model):
        self._model_ref = weakref.ref(model)

    @property
    def model(self) -> hydromt.Model:
        # Access the Model instance through the weak reference
        return self._model_ref()
    
    @property
    def data_catalog(self) -> hydromt.DataCatalog:
        return self.model.data_catalog
    
    @property
    def logger(self) -> logging.Logger:
        return self.model.logger

    @property
    def model_root(self) -> str:
        return self.model.root
```

tests/cli/test_cli.py
hboisgon, line 1: Instead of skipping all these simple tests can you already start updating ``Model`` in api.py and make it possible to create a ``GridModel`` by initializing a ``GridComponent``?
Or you prefer to leave that in the next PR when we merge ModelRegion and ModelComponent better together?

hboisgon (changes requested):
Nice @Tjalling-dejong ! Almost there :)
Again I really like the new tests you added!
To finalize, I really think we should initialize ModelComponent with a weakref to model.

Also we still need to connect ModelComponent to Model. If you prefer we can leave that to a next PR where we merge also ModelRegion in?
After that then we are ready to start with other Model Components.

hboisgon:
> GridComponent still looks like a huge class. I am wondering if we could make this a bit smarter. For example, by changing the way we have all these `add_data_from...` functions. Would it be easier to have different child classes that inherit from GridComponent? Like GeoDataFrameGridComponent? Or should we perhaps make a factory class with a builder pattern?
> 
> ```python
> class GridComponentFactory:
>     @staticmethod
>     def create_geodataframe_grid() -> GridComponent:
>         c = GeoDataFrameComponent()
>         c.create(all_the_config_stuff)
>         return c
> ```

Maybe I don't understand but not sure I get the use for this?

In your example ``GridComponent.create_from_geodataframe`` means we will read GeoDataFrame data using the ``GeoDataFrameAdapter`` and then transform it so that we create the grid definition of ``GridComponent.data`` from it. 
Or the ``add_`` methods are here to read data with the right Adapter and transform it so that we add a new xarray.Variable to ``GridComponent.data`` (xr.Dataset).

hydromt/io/readers.py
```diff
@@ -836,3 +839,79 @@ def parse_values(
                 value = None
             cfdict[section].update({key: value})
     return cfdict
+
+
+def _process_config_in(d):
```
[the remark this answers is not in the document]
savente93: It's not, it was removed in #791

tests/cli/test_cli.py
savente93: I would actually prefer to add that in a separate PR, since that's going to add quite a bit of logic of it's own, I think it's best to keep them separate. It will be the next thing we pick up after this PR, so the tests shouldn't stay skipped for long

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,163 @@
...
+    grid_from_geodataframe,
+    grid_from_raster_reclass,
+    grid_from_rasterdataset,
+    rotated_grid,
+)
+
+logger = getLogger(__name__)
+
+
+class GridComponent(ModelComponent):
```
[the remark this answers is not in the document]
savente93: I think it's good that the import names actually reflect the hirarchy. Users can just import them from the model sub module because we'll export them through the `__init__.py` so for them it would still be 
`from hydromt.models import GridComponent` or even `from hydromt.model.components import GridComponent` if they want to be more specific

hydromt/models/components/grid.py
hboisgon: So we should try to avoid it indeed at the ``ModelComponent`` level but at the level of ``Model`` (where my setup_reservoirs function is) I think it is fine and unavoidable that several components can get updated

hydromt/models/components/grid.py
Tjalling-dejong: I need the ModelRegion class for that

hydromt/models/components/base.py
```diff
@@ -12,7 +12,7 @@
 
 
 class ModelComponent:
-    """Base Model component class."""
+    """Abstract base class for ModelComponent."""
 
     def __init__(self, model: "Model"):
         self._model_ref: ReferenceType["Model"] = ref(model)
```
hboisgon, line 18: Before merging, you had added a couple of abstractmethod for set / read /write.
I think maybe we still want that in case users want to implement their own component then it's clear what methods they need to add.
What we can do is open an issue to finalise ModelComponent together with region in a different PR.
@savente93 already created a couple of issues for region that should be solved to get everything working properly.
Maybe @savente93 could you open a more general issue that recaps all the steps that are needed to finalize ModelComponent implementation before we are then ready to add new ones? Then you can mention there the abstractmethod of ModelComponent.

Then @Tjalling-dejong you can leave it as is for this PR.

hydromt/models/components/grid.py
```diff
@@ -13,370 +10,152 @@
...
-    _API = {"grid": xr.Dataset}
+class GridComponent(ModelComponent):
+    """ModelComponent class for grid components.
 
-    def __init__(self, *args, **kwargs) -> None:
-        super().__init__(*args, **kwargs)
-        self._grid = None  # xr.Dataset()
+    This class is used for setting, creating, writing, and reading grids for a
+    HydroMT model. The grid data in the data property of this class is of the
+    hydromt.gis.raster.RasterDataset type.
```
hboisgon, line 39: Nice additionnal explanation! We do need that :) a few suggestions to improve:
```suggestion
    This class is used for setting, creating, writing, and reading regular grid data for a
    HydroMT model. The grid component data stored in the ``data`` property of this class is of the
    hydromt.gis.raster.RasterDataset type which is an extension of xarray.Dataset for regular grid.
```

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,163 @@
...
-    ) -> List[str]:
-        """HYDROMT CORE METHOD: Add data variable(s) from ``raster_fn`` to grid object.
-
-        If raster is a dataset, all variables will be added unless ``variables`` list
-        is specified.
-
-        Adds model layers:
+        data: Union[xr.DataArray, xr.Dataset, np.ndarray],
+        name: Optional[str] = None,
+        read: bool = True,
```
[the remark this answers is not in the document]
hboisgon: Please remove the read argument here and replace later by ``self.model_root.is_reading_mode()``

hydromt/models/components/grid.py
hboisgon: For this PR leave set as a public method. @deltamarnix if you still think this method should be private, you can open an issue and we can discuss further then.

hydromt/models/components/grid.py
```diff
@@ -13,370 +10,152 @@
...
+        elif not isinstance(data, xr.Dataset):
+            raise ValueError(f"cannot set data of type {type(data).__name__}")
 
-        return list(ds_out.data_vars.keys())
+        if len(self._data) == 0:  # empty grid
+            self._data = data
+        else:
+            for dvar in data.data_vars:
+                if dvar in self._data:
+                    if read:
```
hboisgon, line 100:
```suggestion
                    if self.model_root.is_reading_mode():
```

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,163 @@
...
+            _exec_nodata_strat(
+                msg="No grid data found, skip writing.",
+                strategy=NoDataStrategy.IGNORE,
+                logger=logger,
+            )
+        else:
+            if not self.root.is_writing_mode():
+                raise IOError("Model opened in read-only mode")
+            # write_nc requires dict - use dummy 'grid' key
+            return write_nc(  # Can return DeferedFileClose object
```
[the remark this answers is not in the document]
hboisgon: Please remove the return 
```suggestion
            write_nc(  # Can return DeferedFileClose object
```

hydromt/models/components/grid.py
```diff
@@ -13,371 +11,163 @@
...
-        vector_fn: Union[str, Path, gpd.GeoDataFrame],
-        variables: Optional[Union[List, str]] = None,
-        nodata: Optional[Union[List, int, float]] = -1,
-        rasterize_method: Optional[str] = "value",
-        mask_name: Optional[str] = "mask",
-        rename: Optional[Dict] = None,
-        all_touched: Optional[bool] = True,
-    ) -> List[str]:
-        """HYDROMT CORE METHOD: Add data variable(s) to grid object by rasterizing the data from ``vector_fn``.
+        data_like: str = "grid/grid.nc",
```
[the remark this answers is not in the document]
hboisgon: Please use ``fn`` as in write. We will refine later the name we should use. And in that case it cannot be xarray just str or path so data_like is confusing.
```suggestion
        fn: str = "grid/grid.nc",
```

hydromt/models/components/grid.py
```diff
@@ -683,86 +401,363 @@ def setup_grid(
             grid = grid.drop_vars("mask")
 
         # Add region and grid to model
-        self.set_geoms(geom, "region")
-        self.set_grid(grid)
+        self.model.set_geoms(geom, "region")
```
hboisgon, line 404: Let's see in the next PR how to deal with this together with @savente93

hydromt/models/components/grid.py
```diff
@@ -683,86 +401,363 @@ def setup_grid(
...
+
+    def add_data_from_constant(
+        self,
+        constant: Union[int, float],
+        name: str,
+        dtype: Optional[str] = "float32",  # TODO: change dtype to np.dtype
+        nodata: Optional[Union[int, float]] = None,
+        mask_name: Optional[str] = "mask",
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Adds a grid based on a constant value.
```
hboisgon, line 490:
```suggestion
        """HYDROMT CORE METHOD: Adds data to grid component based on a constant value.
```

hydromt/models/components/grid.py
```diff
@@ -683,86 +401,363 @@ def setup_grid(
...
+
+        Parametersfr
```
hboisgon, line 492:
```suggestion
        Parameters
```

hydromt/models/components/grid.py
```diff
@@ -683,86 +401,363 @@ def setup_grid(
...
+    def add_data_from_rasterdataset(
+        self,
+        raster_fn: Union[str, Path, xr.DataArray, xr.Dataset],
+        variables: Optional[List] = None,
+        fill_method: Optional[str] = None,
+        reproject_method: Optional[Union[List, str]] = "nearest",
+        mask_name: Optional[str] = "mask",
+        rename: Optional[Dict] = None,
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Add data variable(s) from ``raster_fn`` to grid object.
```
hboisgon, line 533:
```suggestion
        """HYDROMT CORE METHOD: Add data variable(s) from ``raster_fn`` to grid component.
```

hydromt/models/components/grid.py
```diff
@@ -683,86 +401,363 @@ def setup_grid(
...
+        reclass_table_fn: Union[str, Path, pd.DataFrame],
+        reclass_variables: List,
+        variable: Optional[str] = None,
+        fill_method: Optional[str] = None,
+        reproject_method: Optional[Union[List, str]] = "nearest",
+        mask_name: Optional[str] = "mask",
+        rename: Optional[Dict] = None,
+        **kwargs,
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Add data variable(s) to grid object by reclassifying the data in ``raster_fn`` based on ``reclass_table_fn``.
```
hboisgon, line 606:
```suggestion
        """HYDROMT CORE METHOD: Add data variable(s) to grid component by reclassifying the data in ``raster_fn`` based on ``reclass_table_fn``.
```

hydromt/models/components/grid.py
```diff
@@ -683,86 +401,363 @@ def setup_grid(
...
+        self,
+        vector_fn: Union[str, Path, gpd.GeoDataFrame],
+        variables: Optional[Union[List, str]] = None,
+        nodata: Optional[Union[List, int, float]] = -1,
+        rasterize_method: Optional[str] = "value",
+        mask_name: Optional[str] = "mask",
+        rename: Optional[Dict] = None,
+        all_touched: Optional[bool] = True,
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Add data variable(s) to grid object by rasterizing the data from ``vector_fn``.
```
hboisgon, line 692:
```suggestion
        """HYDROMT CORE METHOD: Add data variable(s) to grid component by rasterizing the data from ``vector_fn``.
```

hboisgon (changes requested):
Almost there @Tjalling-dejong ! Just a few minor suggestions before this is finalized and can be merged :)

After this PR it would be good to harmonise and finalise the link of ``GridComponent`` to ``ModelRegionComponent`` and to update and link ``GridComponent`` to ``Model`` so that we can create gridded models again.

hydromt/models/components/base.py
savente93: I think this should get solved when Marnix and I work on integrating the components with the model itself since that is when that API really has to become concrete. I'll make a ticket if you like, but I think this is okay as is in this PR

savente93 (commented):
I broadly agree with Hélène's comments so once you resolve those I'm happy for it to be merged. Nearly there!

hydromt/models/__init__.py
```diff
@@ -2,8 +2,7 @@
 """HydroMT models API."""
 from hydromt._compat import HAS_XUGRID
 from hydromt.models.api import Model
-from hydromt.models.components import ModelComponent, ModelRegionComponent
```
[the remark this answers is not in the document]
Tjalling-dejong: I added ModelRegionComponent back to the init. I dont think ModelComponent needs to be added.

deltamarnix (approved):
Good busy! Looks like a solid PR. I assume we can get those tests up and running as soon as the Model class is also ready.

hydromt/models/__init__.py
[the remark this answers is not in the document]
hboisgon: I think ModelComponent will be part of the public API for example if users want to create a completely new component. We can add this in the next PR though.

hydromt/models/components/grid.py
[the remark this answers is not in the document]
hboisgon: Let's discuss this in the next harmonisation PR

hydromt/models/components/grid.py
```diff
@@ -13,461 +10,194 @@
...
-            Use None to disable masking.
-
-        Returns
-        -------
-        list
-            Names of added model grid layer.
+        model: Model
+            HydroMT model instance
+        logger : Logger, optional
+            logger object, by default logger
```
hboisgon, line 53:
```suggestion
```
You forgot to remove it from the docstrings

hydromt/models/components/grid.py
```diff
@@ -13,461 +10,194 @@
...
-        raster_fn: Union[str, Path, xr.DataArray],
-        reclass_table_fn: Union[str, Path, pd.DataFrame],
-        reclass_variables: List,
-        variable: Optional[str] = None,
-        fill_method: Optional[str] = None,
-        reproject_method: Optional[Union[List, str]] = "nearest",
-        mask_name: Optional[str] = "mask",
-        rename: Optional[Dict] = None,
+        fn: str = "grid/grid.nc",
+        temp_data_dir: Optional[StrPath] = None,
```
hboisgon, line 106: Could this still be a ``Model`` property like before instead of a new argument? 
It's really to avoid PermissionError when writing the model. The argument should be part of ``write_nc`` as it is not linked to ``Model`` anymore but ``GridComponent`` still is.

```suggestion
```

hydromt/models/components/grid.py
```diff
@@ -13,461 +10,194 @@
...
-            Use None to disable masking.
-        rename: dict, optional
-            Dictionary to rename variable names in reclass_variables before adding to
-            grid {'name_in_reclass_table': 'name_in_grid'}. By default empty.
-
+        fn : str, optional
+            filename relative to model root, by default 'grid/grid.nc'
+        temp_data_dir: StrPath, optional
+            Temporary directory to write grid to. If not given a TemporaryDirectory
+            is generated.
```
hboisgon, line 122:
```suggestion
```

hydromt/models/components/grid.py
```diff
@@ -13,461 +10,194 @@
...
-
-        return list(ds_vars.data_vars.keys())
+        else:
+            if not self.model_root.is_writing_mode():
+                raise IOError("Model opened in read-only mode")
+            # write_nc requires dict - use dummy 'grid' key
+            write_nc(  # Can return DeferedFileClose object
+                {"grid": self.data},
+                fn,
+                temp_data_dir=temp_data_dir,
```
hboisgon, line 148:
```suggestion
                temp_data_dir=self.model._TMP_DATA_DIR,
```

hboisgon (approved):
Great work @Tjalling-dejong ! Looking good now :)
I still have a couple of very little comments but at the level of for example an argument that you forgot to remove from the docstrings. So would be great if you could still update before merging. But I approve already as they are quite minor.
Great job!

PR #771, opened 2024-02-08
Release v0.9.3

savente93:
This release fixes several bugs. Most notably the `NoDataSrategy` is available in much more data reading methods so plugins can use it more directly. Additionally there are some bug fixes relating to reading shapefiles and reading COGs.

### DIFF
--- a/docs/_static/switcher.json
+++ b/docs/_static/switcher.json
@@ -1,57 +1,52 @@
 [
-
   {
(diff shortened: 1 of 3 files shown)


savente93:
The failing windows tests are because of a misconfigured CI and do not impact the release, so I'll release regardless of them given that all the other checks pass.

PR #780, opened 2024-02-14
Add Model region class

savente93:
# Functional requirements

## Solved
- [ ] deprecate the res argument

## Postponed
- [ ] Implementing of basin, sub basin and inter basin 
- [ ] Model based region

## Needs decision
- [x] Currently append mode is not treated as an override during writing. is this okay? (instead it's okay to overwrite if we're in any write mode, so this is not necessary)


DirkEilander:
# Functional requirements
- Defines geographical region of a Model instance.
- User provides region dictionary (bbox, geom, subbasin, basin, etc)
- This gets parsed to a *single* geopandas geodataframe and saved in a ModelRegion property (currently this is done in `Model.parse_region`)
- This property should be accessible from components to retrieve data.

savente93:
@DirkEilander @hboisgon quick question about the read/write. I assume that this just refers to the geodataframe that the parsed dictionary will get turned into right? in which case, the write method is quite straight forward, but I still have a question about the read method. How is doing `region.read` different from `region({'geom':'path/to/geom/file'})`? i.e. is the `region.read` the same as defining a region based on a file or is there more to it?

savente93:
oh and one after thought: do these `read/write` need to obey the model mode? (e.g. read, write, append, overwrite)

hboisgon:
Hey @savente93, good questions!
The most important part is indeed to read/write the geopandas geodataframe of the region and maybe let the user path indeed the file path (relative to model.root) and maybe also format and use GeoJSON by default (same as geoms).
Region is normally really more a HydroMT thing rather than a Model thing but it's something that a model could decide to overwrite if they do need the region in a more specific format.
For saving the original region dict, that might be nice for reproducibility as for now we only save it in the hydromt.log. Not sure if there is a nice format for it though? 
And yes ``read/write`` should obey the model mode.

savente93:
I know there are a lot of region types to still be implemented, but I first wanted to get an okay on the basic flow. The idea is now:

1. you construct a `Region` class with the same dictionary formats you are used to
2. this will give you a Region class with a `spec` which is an abstract description of the region. this spec does the validation of syntactic validity
3. when you want the `GeoDataFrame` you call `construct()` on the `Region` object at which point all the I/O or necessary calculations will happen.
4. Internally in the rest of HydroMT we just use GeoDataFrame objects to refer to regions as we agreed on after the standup yesterday. 

This is not exactly the same way as the rest of the code but is well in line with the lazy operations which I like. It also simplifies a lot of the other code since they have to deal with a lot less options, so that's also very good.

Finally I added two questions that I still need some help on, but other than that I think this is a good proof of concept. after we agree on this, we'll start work on implementing more region types again.

hydromt/region/region.py
```diff
@@ -0,0 +1,85 @@
...
+class Region:
+    """A class to handle all operations to do with the model region."""
+
+    def __init__(
+        self,
+        region_dict: Dict[str, Any],
+        logger: Logger = logger,
+    ):
+        self._dict: Dict[str, Any] = region_dict
+        self._data: Optional[GeoDataFrame] = None
```
hboisgon, line 27: The geopandas object is the main property of region for the model so I would make it easier to access it.
For example, a ``data`` property that returns ``_data`` or try and calls ``construct`` if it is None?

Might be nice to have shortcut properties to ``crs`` and ``bounds`` (total_bounds) as well to make it easier to use in model but this is really small shortcuts so maybe not worth it (model.region.crs rather than model.region.data.crs)?

hydromt/workflows/basin_mask.py
hboisgon, line 1: Are we not running at all the "old/main" tests for a merge in v1? Maybe for some parts the tests need to be adapted a lot but for model I would say they are quite useful to run still to check if we are not forgetting things when we start replacing to v1 architecture.

hydromt/workflows/mesh.py
```diff
@@ -13,7 +13,6 @@
 
 from hydromt.gis import utils
 from hydromt.gis.raster import GEO_MAP_COORD
-from hydromt.workflows.basin_mask import parse_region
```
hboisgon, line 16: Keep or update the ``create_mesh2d`` method already in this PR

tests/cli/test_cli.py
```diff
@@ -53,6 +53,7 @@ def test_cli_clip_help(tmpdir):
     )
 
 
+@pytest.mark.skip(reason="Needs implementation of RasterDataSet.")
```
hboisgon, line 56: I don't see this skip for MeshModel, is xugrid properly imported in the test environment?

tests/data_adapter/test_geodataframe.py
hboisgon, line 1: Same here, revert using region to use instead geom/bbox/buffer

tests/data_sources/test_geodataframe_data_source.py
hboisgon, line 1: Same here, revert using region to use instead geom/bbox/buffer

tests/drivers/test_pyogrio_driver.py
hboisgon, line 1: Same here, revert using region to use instead geom/bbox/buffer

tests/models/test_model.py
```diff
@@ -39,8 +40,6 @@ def test_api_attrs():
     assert hasattr(dm, "_API")
     assert "asdf" in dm.api
     assert dm.api["asdf"] == "yeah"
-    assert "region" in dm.api
```
hboisgon, line 42: Can update to check for Region class. I see in the current implementation we are not initializing with an empty gdf anymore, good to reference in the v1 changelog

tests/workflows/test_basin_mask.py
```diff
@@ -1,123 +1,18 @@
...
-    _parse_region_value,
     get_basin_geometry,
-    parse_region,
 )
 
-logger = logging.getLogger("tets_basin")
-
-
-@pytest.mark.skip(reason="Needs refactor to Pydantic BaseModel.")
-def test_region(tmpdir, world, geodf, rioda):
```
hboisgon, line 25: But you did move them all to test_model_region.py even if they are now skipped right?

hydromt/gis/raster.py
```diff
@@ -2286,9 +2286,9 @@ def interpolate_na(
         kwargs.update(dict(method=method, extrapolate=extrapolate))
         if dim0:
             interp_data = np.empty(self._obj.shape, dtype=self._obj.dtype)
-            for i, (_, sub_xds) in enumerate(self._obj.groupby(dim0)):
+            for i, (_, sub_xds) in enumerate(self._obj.groupby(dim0, squeeze=False)):
```
hboisgon, line 2289: Why did you change this?

hboisgon (commented):
Hi @savente93 ! Some nice progress still :)
I left comments here and there, let's also discuss somethings in our meeting together and hopefully you can go on with this PR

DirkEilander (commented):
A few suggestions / question:
- I noticed the basins/ subbasins and interbasins are still TODO. Note that these are quite different to the others (geom /bbox) as these need access to hydrography data via the data catalog to be derived. Keep this in mind in the design.
- Would it make sense to implement ModelRegion as a special case of the ModelComponent class and initialize it with a link to Model? That provides link to root, the data catalog and the logger.
- I'm happy with moving xugrid to the required dependencies for v1 as more plugins require it, and it has become more stable and lightweight. This would remove the necessity for the HAS_XUGRID statements.

hydromt/models/root.py
```diff
@@ -1,4 +1,4 @@
-"""the new model root class."""
```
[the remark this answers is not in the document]
savente93: good catch

hydromt/region/region.py
savente93: sure, I've added them.

hydromt/gis/raster.py
savente93: because it produces the following warning:

```
tests/gis/test_raster.py::test_interpolate_na
  /home/sam/Documents/work/hydromt/hydromt/gis/raster.py:2289: UserWarning: The `squeeze` kwarg to GroupBy is being removed.Pass .groupby(..., squeeze=False) to disable squeezing, which is the new default, and to silence this warning.
    for i, (_, sub_xds) in enumerate(self._obj.groupby(dim0)):
```

tests/workflows/test_basin_mask.py
savente93: The relevant ones yes. I think the one that test `parse_region_value` I omitted because these are now done by pydantic directly and therefore don't have to be tested there, but the coverage should be equivalent.

hydromt/data_adapter/geodataframe.py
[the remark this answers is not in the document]
savente93: I've thought about it a bit, and while we can still have a discussion about this, this is not really part of this PR. So I will revert these changes here so we can continue working on this, and make an issue for this that we can refine during the next refinement. Are you okay with this?

hydromt/models/api.py
```diff
@@ -1260,10 +1186,10 @@ def set_geoms(self, geom: Union[gpd.GeoDataFrame, gpd.GeoSeries], name: str):
             )
         if name in self._geoms:
             self.logger.warning(f"Replacing geom: {name}")
-        if hasattr(self, "crs"):
```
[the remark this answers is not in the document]
savente93: I don't remember why and restoring it fails no tests so I'll just restore it.

tests/cli/test_cli.py
savente93: Sorry I don't understand the question. This test is skipped because we haven't implemented Rasterdataset yet. What does that have to do with the MeshModel?

tests/models/test_model.py
savente93: I'll add the check that we have a region here, but I don't understand the rest of your comment. What would you like to be referenced in the chagnelog?

hydromt/workflows/basin_mask.py
savente93: The basin code will be reenabled once I work on implementing the basins which I've postponed for now. The other code that was removed in this file was adapted into the new region checking code, so no tests are lost.

hydromt/models/api.py
```diff
@@ -342,106 +338,36 @@ def update(
     def setup_region(
         self,
         region: dict,
-        hydrography_fn: str = "merit_hydro",
```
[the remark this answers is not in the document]
savente93: I will add the hydrography and basin index back in when I implement the basin functionalities. I just temproarily removed them for my own sanity.

hydromt/region/region.py
```diff
@@ -0,0 +1,85 @@
...
+        self._spec = self._parse_region(region_dict, logger=logger)
+        self.logger = logger
+
+    def write(self, path: StrPath, mode: ModelMode = ModelMode.WRITE):
```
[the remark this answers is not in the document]
savente93: can do but that is going to make the testing much more convoluted since then we can't test region and model separately anymore. (again moving to having only integration tests) is this really necessary? I'll make sure that it's always set correctly when used from the model.

DirkEilander:
> I know there are a lot of region types to still be implemented, but I first wanted to get an okay on the basic flow. The idea is now:
> 
> 1. you construct a `Region` class with the same dictionary formats you are used to
> 2. this will give you a Region class with a `spec` which is an abstract description of the region. this spec does the validation of syntactic validity
> 3. when you want the `GeoDataFrame` you call `construct()` on the `Region` object at which point all the I/O or necessary calculations will happen.
> 4. Internally in the rest of HydroMT we just use GeoDataFrame objects to refer to regions as we agreed on after the standup yesterday.
> 
> This is not exactly the same way as the rest of the code but is well in line with the lazy operations which I like. It also simplifies a lot of the other code since they have to deal with a lot less options, so that's also very good.
> 
> Finally I added two questions that I still need some help on, but other than that I think this is a good proof of concept. after we agree on this, we'll start work on implementing more region types again.

@savente93 I had missed this comment earlier. Some more info to clarify the user requirements: 
- When building a model, a user will call `ModelRegion.create` with the region_dict (in v0.9 this was `Model.setup_region`). And in reading mode, a region.geojson file should be read from model folder triggered by a call to `ModelRegion.data`. In the current setup the region_dict is required at initialization of `ModelRegion` which won't work for these requirements I think. Instead I suggest to initialize `ModelRegion` _empty_ at initialization of `Model` with a only weak ref to model. The `ModelRegion.create` method would than receive the region_dict and parse this to a geodataframe to store in `ModelRegion._data`. And in read mode, `ModelRegion.data` would trigger reading the region.geojson file if `ModelRegion._data` is None. This workflow is similar for all ModelComponents, e.g.  `GridModelComponent.create` creates and stores a grid a `GridModelComponent._data` and `GridModelComponent.data` in reading mode triggers the reading method if _data is still None. I therefore suggest to make the ModelRegion also a child of ModelComponent where we can deal with the lazy reading behavior.  
- Constructing the region (i.e. translating dict input to a geometry) is always the first step in the model building process and the geometry is immediately required by the next steps (e.g. to request a geospatial slice of some data). Although it's fine to separate the validation and the construct, there is no need to make this a lazy operation, and the `ModelRegion.create` should do both.

savente93:
After merging #810 it will be a lot more work to merge this back to a working state than I thought. The code will still be available after the PR is closed, so I'll be closing this now, and come back to it when I'm ready to work on this again. I think there are plenty other things that have higher priority right now.

PR #782, opened 2024-02-15
relax warning in rasterdataset

### LINKED ISSUE #770: relax warnings in NoDataStrategy
### HydroMT version checks

### Reproducible Example

Yo can find the example here: https://deltares.github.io/hydromt_wflow/latest/_examples/prepare_ldd.html

### Current behaviour

It's not really a bug I think but seems the nodata strategy is a bit excessive in its reaction:
I think so far this a warning but not an error but when I open data for a specific bbox, I see a warning that not all of my bbox is covered by the data due to rounding errors:

image
image

### Desired behaviour

In case in the future this leads to unexpected behavior, I think we should be a bit more relaxed here in our warning if we can.

### Additional context

_No response_

Jaapel:
## Issue addressed
Fixes #770
## Explanation

## Checklist

### DIFF
--- a/hydromt/data_adapter/rasterdataset.py
+++ b/hydromt/data_adapter/rasterdataset.py
@@ -597,10 +597,17 @@ class RasterDatasetAdapter(DataAdapter):
             bbox_str = ", ".join([f"{c:.3f}" for c in bbox])
             logger.debug(f"Clip to [{bbox_str}] (epsg:{epsg}))")
             ds = ds.raster.clip_bbox(bbox, buffer=buffer, align=align)
+
             # if np.any(np.array(ds.raster.shape) < 2):
             # check if bbox is fully covered
+            def _lt_or_close(a: float, b: float) -> bool:
+                return np.isclose(a, b) or a < 0
+
             w, s, e, n = ds.raster.bounds
-            if not (w <= bbox[0] and s <= bbox[1] and e >= bbox[2] and n >= bbox[3]):
+            if not any(
+                map(_lt_or_close, (w, s, bbox[2], bbox[3]), (bbox[0], bbox[1], e, n))
+            ):
+                # if not (w <= bbox[0] and s <= bbox[1] and e >= bbox[2] and n >= bbox[3]):
                 logger.warning(
                     f"Dataset does [{w}, {s}, {e}, {n}] does not fully cover bbox [{bbox_str}]"
                 )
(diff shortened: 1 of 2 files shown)


hydromt/data_adapter/rasterdataset.py
```diff
@@ -597,10 +597,17 @@ def _slice_spatial_dimensions(
             bbox_str = ", ".join([f"{c:.3f}" for c in bbox])
             logger.debug(f"Clip to [{bbox_str}] (epsg:{epsg}))")
             ds = ds.raster.clip_bbox(bbox, buffer=buffer, align=align)
+
             # if np.any(np.array(ds.raster.shape) < 2):
             # check if bbox is fully covered
+            def _lt_or_close(a: float, b: float) -> bool:
+                return np.isclose(a, b) or a < 0
```
DirkEilander, line 604: Shouldn't this be `a < b` instead of `a < 0`

```suggestion
                return np.isclose(a, b) or a < b
```

Alternatively this check could also be done before the clip_bbox (L599) to circumvent most rounding errors. But probably not all for some edge cases, so the new check is still better.

hydromt/data_adapter/rasterdataset.py
```diff
@@ -597,10 +597,17 @@ def _slice_spatial_dimensions(
...
+
             w, s, e, n = ds.raster.bounds
-            if not (w <= bbox[0] and s <= bbox[1] and e >= bbox[2] and n >= bbox[3]):
+            if not any(
+                map(_lt_or_close, (w, s, bbox[2], bbox[3]), (bbox[0], bbox[1], e, n))
+            ):
+                # if not (w <= bbox[0] and s <= bbox[1] and e >= bbox[2] and n >= bbox[3]):
                 logger.warning(
                     f"Dataset does [{w}, {s}, {e}, {n}] does not fully cover bbox [{bbox_str}]"
```
hboisgon, line 612:
```suggestion
                    f"Dataset [{w}, {s}, {e}, {n}] does not fully cover bbox [{bbox_str}]"
```

hboisgon (commented):
Thanks for picking this up @Jaapel. I agree with the comments from @DirkEilander

DirkEilander (approved):
Thanks Jaap! The code looks good to me! I haven't actually verified it fixes the warning in the particular example of Hélène, but I'm confident it does.

PR #783, opened 2024-02-16
Pixi multi envs

deltamarnix:
## Explanation
mamba comes with its quirks and complications. Pixi seems to have some nice solutions for multi-environment dependency setup. 

mamba/conda is still supported, but life is way easier with pixi as a task runner.

## Checklist


savente93 (commented):
I like the direction this is going in, since it will save us a lot of maintenance heavy Ci stuff. What would be necessary to make this PR a not-draft?

Dockerfile
```diff
@@ -1,70 +1,23 @@
...
- && micromamba clean -ayf \
- && find /opt/conda/ -follow -type f -name '*.a' -delete \
- && find /opt/conda/ -follow -type f -name '*.pyc' -delete \
- && find /opt/conda/ -follow -type f -name '*.js.map' -delete
- ENTRYPOINT [ "micromamba", "run", "-n", "hydromt" ]
- CMD ["hydromt","--models"]
+FROM debian:bookworm-slim as base
+ARG PIXIENV
+WORKDIR /home/hydromt
+RUN apt-get update && apt-get install -y curl
```
deltamarnix, line 4: It's interesting how micromamba made a base image and then added an extra stage to prevent the install programs to remain within the docker image: https://github.com/mamba-org/micromamba-docker/blob/main/Dockerfile

Maybe we can do something similar.

Dockerfile
```diff
@@ -1,70 +1,23 @@
...
- && micromamba clean -ayf \
- && find /opt/conda/ -follow -type f -name '*.a' -delete \
- && find /opt/conda/ -follow -type f -name '*.pyc' -delete \
- && find /opt/conda/ -follow -type f -name '*.js.map' -delete
- ENTRYPOINT [ "micromamba", "run", "-n", "hydromt" ]
- CMD ["hydromt","--models"]
+FROM base as full
+COPY tests/ ./tests
+
+FROM base as slim
```
deltamarnix, line 23: For me it's not clear why tests should be in full and minimal, but not in slim. Should the examples be in min?

Dockerfile
savente93: if we can do that all the better!

Dockerfile
savente93: I think that was a mistake of mine. examples and tests should only be in full I think.

deltamarnix: done

deltamarnix:
I have removed the pixi cache to save a couple of 100 MBs. Other cache folders are not so interesting. site-packages is the largest for sure. Docker desktop tells me 3.26GB right now. I could try to remove all pyc files.

deltamarnix:
I've come back to 105MB!

savente93 (approved):
Very nice Marnix. This saves us a whole lot of code maintenance. I trust that you've tested the CIs et al. where appropriate and the files look good to me so I'm happy to approve. we'll have to keep the `pixi.toml` and `pyproject.toml` in sync manually until one starts supporting the other, but I'm okay with this small amount of duplication if it saves us a bunch of maintenance. Cloud you also make an announcement about this before merging?

savente93:
saw there was a small docs error causing the build to fail. it's fixed on main now so if you update your branch again, the tests should pass

Jaapel (approved):
LGTM 👍

DirkEilander (approved):
nice work @deltamarnix! 

I still have a couple of Qs regarding the pixi.lock file and related install behavior, which is unclear to me from the pixi docs.

- I understand the pixi.lock get's updated when you run `pixi install` after you have updated the pixi.toml, correct? 
- If so, how do we keep the pixi.lock up to date? And when should we commit an updated version? Perhaps all steps to take when updating/adding a dependency would be good to add to the dev guide The way I understand it these are: update pyproj.toml; update pixi.toml; run pixi install; commit all changes incl pixi.lock.
- And what happens if new versions of dependencies are released, these are not updated in the pixi.lock right? Hence we always test against the frozen solve of our environment from the time the lock file was last updated and we may miss deprecation warning or even bugs caused by updated dependencies right. Can we develop a strategy to regularly update the pixi.lock file?

deltamarnix:
> nice work @deltamarnix!
> 
> I still have a couple of Qs regarding the pixi.lock file and related install behavior, which is unclear to me from the pixi docs.
> 
>     * I understand the pixi.lock get's updated when you run `pixi install` after you have updated the pixi.toml, correct?
> 
>     * If so, how do we keep the pixi.lock up to date? And when should we commit an updated version? Perhaps all steps to take when updating/adding a dependency would be good to add to the dev guide The way I understand it these are: update pyproj.toml; update pixi.toml; run pixi install; commit all changes incl pixi.lock.
> 
>     * And what happens if new versions of dependencies are released, these are not updated in the pixi.lock right? Hence we always test against the frozen solve of our environment from the time the lock file was last updated and we may miss deprecation warning or even bugs caused by updated dependencies right. Can we develop a strategy to regularly update the pixi.lock file?

@Hofer-Julian Could you give an advise on how you perform regular updates on package dependencies? I have committed the pixi.lock file, because it makes the build process faster, but I understand Dirk's concerns.

savente93:
having the `pixi.lock` file check in also aid in reproducibility so that's also a plus. if it helps, I could quite easily make a workflow that updates the `pixi.lock` very sunday or so

Hofer-Julian:
> having the `pixi.lock` file check in also aid in reproducibility so that's also a plus. if it helps, I could quite easily make a workflow that updates the `pixi.lock` very sunday or so

Yeah, that's exactly what we did: https://github.com/Deltares/Ribasim/blob/main/.github/workflows/pixi_auto_update.yml

deltamarnix:
I have copied the Ribasim script and added more documentation on dependencies.

Issue #721, opened 2024-01-10
Improve cross platform support for the DataCatalog root


savente93:
### HydroMT version checks

- [X] I have checked that this issue has not already been reported.
- [X] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

Currently, for example the root of the deltares data catalog, is formatted as a windows path, which can cause issues when Hydromt is running in a linux environment. 

### Current behaviour

Mixing of paths can cause iether crashes or undefined behaviour. 

### Desired behaviour

hydromt can detect which platform it is running on and adjust the paths accordingly.

DirkEilander:
suggested implementation:
List with roots ordered by priority. We loop over this list and check which root exists and build all paths from there.

Issue #696, opened 2023-12-20
Update artifact_data catalog for EPSG 54009


hboisgon:
### HydroMT version checks

- [X] I have checked that this issue has not already been reported.
- [X] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

We have updated deltares_data catalog for data with crs 54009 but not artifact_data.

### Current behaviour

artifact_data with crs 54009 cannot be read

### Desired behaviour

Update artifact_data for data with crs: 54009 to crs: "ESRI:54009"

savente93:
Fixed in v1 in #786

PR #786, opened 2024-02-19
Data Catalog roots

savente93:
# Functional Requirements

## Solved
- [x] Data catalogs can not specify multiple roots to make the catalogs more portable and give better cross platform support as these can be either Windows or POSIX paths
- [x] Roots provided will be checked in order, first one to be found will be used. 
- [x] When none of the provided roots are found an error is raised 
- [x] if no roots are specified anywhere, the current working directory will be used. 

## Postponed
- [ ] Removing the other cross platform path transformations (Fixed in #788)
- [ ] Re enable Model Root logger file path handling  (Fixed in #788)

## To be agreed on
None so far that I am aware of

### DIFF
--- a/data/catalogs/artifact_data.yml
+++ b/data/catalogs/artifact_data.yml
@@ -1,7 +1,8 @@
 meta:
   version: v0.0.8
-  hydromt_version: '>=0.9, <1.0'
-  root: https://github.com/DirkEilander/hydromt-artifacts/releases/download/0.0.8/data.tar.gz
+  hydromt_version: '>=1.0, <2.0'
+  roots:
+    - https://github.com/DirkEilander/hydromt-artifacts/releases/download/0.0.8/data.tar.gz
 
 chelsa:
   crs: 4326
--- a/data/catalogs/artifact_data.yml
+++ b/data/catalogs/artifact_data.yml
@@ -266,7 +266,7 @@ gebco:
     unit: m+MSL
   path: gebco.tif
 ghs-smod_2015_v2:
-  crs: 54009
+  crs: ESRI:54009
   data_type: RasterDataset
   driver: raster
   meta:
@@ -292,7 +292,7 @@ ghs_pop_2015:
     source_version: R2019A_v1.0
   path: ghs_pop_2015.tif
 ghs_pop_2015_54009:
-  crs: 54009
+  crs: ESRI:54009
   data_type: RasterDataset
   driver: raster
   meta:
@@ -303,7 +303,7 @@ ghs_pop_2015_54009:
     source_url: https://ghsl.jrc.ec.europa.eu/download.php?ds=pop
   path: ghs_pop_2015_54009.tif
 ghs_smod_2015:
-  crs: 54009
+  crs: ESRI:54009
   data_type: RasterDataset
   driver: raster
   meta:
--- a/data/catalogs/artifact_data.yml
+++ b/data/catalogs/artifact_data.yml
@@ -1,6 +1,6 @@
 meta:
   version: v0.0.8
-  hydromt_version: '>=1.0, <2.0'
+  hydromt_version: '>=1.0a, <2.0'
   roots:
     - https://github.com/DirkEilander/hydromt-artifacts/releases/download/0.0.8/data.tar.gz
 
--- a/data/catalogs/aws_data.yml
+++ b/data/catalogs/aws_data.yml
@@ -1,7 +1,7 @@
 ---
 meta:
   version: 2024.1.30
-  hydromt_version: '>=0.9, <1.0'
+  hydromt_version: '>=1.0, <2.0'
 
 
 esa_worldcover_2020_v100:
--- a/data/catalogs/aws_data.yml
+++ b/data/catalogs/aws_data.yml
@@ -1,7 +1,7 @@
 ---
 meta:
   version: 2024.1.30
-  hydromt_version: '>=1.0a, <2.0'
+  hydromt_version: '>=0.9, <2.0'
 
 
 esa_worldcover_2020_v100:
--- a/data/catalogs/aws_data.yml
+++ b/data/catalogs/aws_data.yml
@@ -1,7 +1,7 @@
 ---
 meta:
   version: 2024.1.30
-  hydromt_version: '>=1.0, <2.0'
+  hydromt_version: '>=1.0a, <2.0'
 
 
 esa_worldcover_2020_v100:
--- a/data/catalogs/deltares_data.yml
+++ b/data/catalogs/deltares_data.yml
@@ -4,7 +4,7 @@ meta:
     - p:/wflow_global/hydromt
     - /mnt/p/wflow_global/hydromt
   version: 2024.1.30
-  hydromt_version: '>=1.0, <2.0'
+  hydromt_version: '>=1.0a, <2.0'
 
 
 basin_atlas_level12_v10:
--- a/data/catalogs/deltares_data.yml
+++ b/data/catalogs/deltares_data.yml
@@ -3,6 +3,7 @@ meta:
   roots:
     - p:/wflow_global/hydromt
     - /mnt/p/wflow_global/hydromt
+    - /p/wflow_global/hydromt
   version: 2024.1.30
   hydromt_version: '>=1.0a, <2.0'
 
--- a/data/catalogs/deltares_data.yml
+++ b/data/catalogs/deltares_data.yml
@@ -2,6 +2,7 @@
 meta:
   roots:
     - p:/wflow_global/hydromt
+    - /mnt/p/wflow_global/hydromt
   version: 2024.1.30
   hydromt_version: '>=0.9, <1.0'
 
--- a/data/catalogs/deltares_data.yml
+++ b/data/catalogs/deltares_data.yml
@@ -1,6 +1,7 @@
 ---
 meta:
-  root: p:/wflow_global/hydromt
+  roots:
+    - p:/wflow_global/hydromt
   version: 2024.1.30
   hydromt_version: '>=0.9, <1.0'
 
--- a/data/catalogs/deltares_data.yml
+++ b/data/catalogs/deltares_data.yml
@@ -4,7 +4,7 @@ meta:
     - p:/wflow_global/hydromt
     - /mnt/p/wflow_global/hydromt
   version: 2024.1.30
-  hydromt_version: '>=0.9, <1.0'
+  hydromt_version: '>=1.0, <2.0'
 
 
 basin_atlas_level12_v10:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,6 +14,12 @@ Added
 - Added Driver class for customizable io
 - Added MetaDataResolver class for customizable metadata discovery
 - Added DataSource class to represent and validate DataCatalog entries.
+- Data catalogs can now list multiple roots which will be searched for files. (#786)
+
+Changed
+-------
+- The `root` meta key of data catalogs has become `roots` (#786)
+- support for `artifact_keys` has now been removed. (#786)
 
 
 Unreleased
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -19,8 +19,11 @@ Added
 Changed
 -------
 - The `root` meta key of data catalogs has become `roots` (#786)
-- support for `artifact_keys` has now been removed. (#786)
 
+
+Removed
+-------
+- support for `artifact_keys` has now been removed. (#786)
 
 Unreleased
 ==========
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,16 +14,16 @@ Added
 - Added Driver class for customizable io
 - Added MetaDataResolver class for customizable metadata discovery
 - Added DataSource class to represent and validate DataCatalog entries.
-- Data catalogs can now list multiple version of the roots which will be searched for files depending on the system used (linux, windows etc). (#786)
+- Data catalogs can now list multiple roots depending on the system used (linux, windows etc). where the first existing root will be used. (#786)
 
 Changed
 -------
-- The `root` meta key of data catalogs has become `roots` (#786)
+- The `root` meta key of data catalogs yaml files has become `roots` (#786)
 
 
 Removed
 -------
-- support for `artifact_keys` has now been removed. (#786)
+- support for `**artifact_keys` when initializing the DataCatalog has been removed. (#786)
 
 Unreleased
 ==========
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,7 +14,7 @@ Added
 - Added Driver class for customizable io
 - Added MetaDataResolver class for customizable metadata discovery
 - Added DataSource class to represent and validate DataCatalog entries.
-- Data catalogs can now list multiple roots which will be searched for files. (#786)
+- Data catalogs can now list multiple version of the roots which will be searched for files depending on the system used (linux, windows etc). (#786)
 
 Changed
 -------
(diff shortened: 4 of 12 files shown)


savente93:
Fixes #721

data/catalogs/deltares_data.yml
```diff
@@ -1,6 +1,8 @@
 ---
 meta:
-  root: p:/wflow_global/hydromt
+  roots:
+    - p:/wflow_global/hydromt
+    - /mnt/p/wflow_global/hydromt
   version: 2024.1.30
   hydromt_version: '>=0.9, <1.0'
```
DirkEilander, line 7: With these changes we should update the hydromt_version to >1.0 right? We should then also update the ref in the predefined catalogs.

DirkEilander (changes requested):
Nice work @savente93. I have one major suggestion, see comment on the prioritisation of the `root` method argument vs the `yml['meta']['root']` value. Otherwise this looks good to me. If you can respond to that one I'll do another more thorough check. And perhaps good to ask a second reviewer.

hydromt/data_catalog.py
```diff
@@ -961,10 +898,40 @@ def from_dict(
...
+
         if "category" in meta and category is None:
             category = meta.pop("category")
+        version = meta.get("version", None)
+
+        if "roots" not in meta:
+            if root is not None:
+                meta["roots"] = [root]
+            else:
+                meta["roots"] = [realpath(".")]
```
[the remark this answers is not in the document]
savente93: Fair enough. I think I added this more as a testing as sometimes you call `from_dict` that isn't originating from a file (e.g. in testing) so this was meant as a fallback but I agree it has more potential for harm than good so I'll specify that in the tests if necessary .

data/catalogs/deltares_data.yml
savente93: sort of. You're right that the refs need to be updated, but only once this lands on main right? this one is going to v1 not main, and when we merge v1 to main, we'll have to update the refs anyway. Am I correct?

hydromt/data_catalog.py
```diff
@@ -909,11 +827,30 @@ def _is_compatible(
         else:
             return version in requested
 
+    def _determine_catalog_root(
+        self, meta: Dict[str, Any], urlpath: Optional[StrPath] = None
+    ) -> Path:
+        root = None
+        if self.root is not None:
+            return Path(self.root)
```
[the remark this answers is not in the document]
savente93: Good catch. This was from an earlier revision when `from_dict` and `from_yml` both needed to call `_determine_roots` and I think I forgot to remove it when that was no longer necessary. I'll remove it now

savente93:
@DirkEilander should be good now, could you take another look?

hydromt/_validators/data_catalog.py
```diff
@@ -46,21 +49,31 @@ class Extent(BaseModel):
 class DataCatalogMetaData(BaseModel):
     """The metadata section of a Hydromt data catalog."""
 
-    root: Optional[Path] = None
+    roots: List[Path]
     version: Optional[Union[str, Number]] = None
+    hydromt_version: str
```
hboisgon, line 54: I see in from_yml and from_dict that version can still be None. So should the hydromt_version here also be optional?

hydromt/data_catalog.py
```diff
@@ -843,57 +803,18 @@ def from_yml(
...
-            category=meta.get("category", None),
-            mark_used=mark_used,
-        )
+            if "name" in meta:
+                catalog_name = cast(str, meta["name"])
+            else:
+                catalog_name = cast(str, "".join(splitext(basename(urlpath))[:-1]))
+
+            if "roots" not in meta:
+                meta["roots"] = [urlpath]
```
hboisgon, line 814: Not dirname(urlpath)?

tests/validators/test_data_catalog_validation.py
hboisgon, line 1: Maybe add a test with from_yml and no root?

hboisgon (changes requested):
Nice changes @savente93 and good we can finally support cross platform for the data catalogs :)
I just wonder if the case of from_yml with no root works but for the rest technically looks good to go!

I put a request changes because this time I won't approve if the changelog is not updated to reflect that multi platforms is supported, root became roots in yml and that now artifact_data kwargs is really removed. If we keep postponing we will forget things in the v1 changelog :)
Also I would like to already update the data catalog part of the documentation to update the change from root to roots and maybe a simple example to explain how this works because it sure is nice functionnality!

hydromt/_validators/data_catalog.py
savente93: That version refers to the version of the catalogs, which it may or may not have, but `hydromt_version` is the version of hydromt that the catalog is compatible with. I think it is good to force people to specify what versions they are compatible with, since if they don't they won't get warnings if we make incompatible changes and use their old catalogs. What do you think?

hydromt/data_catalog.py
savente93: Good catch.

tests/validators/test_data_catalog_validation.py
savente93: sure, I'll add one.

data/catalogs/deltares_data.yml
savente93: You're right, it's probably best to update them now. I've updated all the catalog versions, but I do wonder if we still need to update the `predefined_catalogs.yml` since this hasn't landed on main yet, and I'm not sure how it will handle things that aren't on main yet. What do you two think? if so I'll make an issue to update those refs when we finsih v1

savente93:
@hboisgon you make good points, thanks for that! I posed two questions in response to your review, when we can agree on those we should be good. in the mean time I'll add some documentation but if you both could take another look that'd be great (@DirkEilander cc)

savente93:
@hboisgon I've updated the documentation. Do you think this is enough or will more be required?

data/catalogs/deltares_data.yml
hboisgon: True, forgot that the latest version points to main and only the others to a commit number. Is there a reason why actually we point to main and not a commit number for the latest version?
But yes it's not so easy to update so let's open the issue to update for later.

savente93: Why we point to main and not a specific commit I'm not sure. @DirkEilander knows most about that I think.

hydromt/_validators/data_catalog.py
hboisgon: It's good to encourage it but not sure we want to force it. Especially since after v1 we hope the syntax and keywords of the catalog should be more stable so this might be a lot of trouble for the user to find out this information and the syntax and keep it updated with new hydromt versions. 
Maybe if None just print a warning that if the hydromt_version was not found, errors can happen later due to that. And do add the current used version? Then if the catalog is exported later it will print and can be re-used?
@DirkEilander what do you think?

docs/changelog.rst
```diff
@@ -14,6 +14,12 @@ Added
 - Added Driver class for customizable io
 - Added MetaDataResolver class for customizable metadata discovery
 - Added DataSource class to represent and validate DataCatalog entries.
+- Data catalogs can now list multiple roots which will be searched for files. (#786)
+
+Changed
+-------
+- The `root` meta key of data catalogs has become `roots` (#786)
+- support for `artifact_keys` has now been removed. (#786)
```
hboisgon, line 22: Suggestion: maybe create a removed deprecation list in the v1 changelog rather than include them in the changed part? Also here we will list the --dd flag, staticmaps/staticgeoms etc.

docs/changelog.rst
```diff
@@ -14,6 +14,12 @@ Added
 - Added Driver class for customizable io
 - Added MetaDataResolver class for customizable metadata discovery
 - Added DataSource class to represent and validate DataCatalog entries.
+- Data catalogs can now list multiple roots which will be searched for files. (#786)
```
hboisgon, line 17:
```suggestion
- Data catalogs can now list multiple version of the roots which will be searched for files depending on the system used (linux, windows etc). (#786)
```

docs/user_guide/data_prepare_cat.rst
```diff
@@ -29,7 +29,9 @@ The ``rename``, ``nodata``, ``unit_add`` and ``unit_mult`` options are set per v
 .. code-block:: yaml
 
     meta:
-      root: /path/to/data_root/
+      roots:
+        - /linxu/path/to/data_root/
```
hboisgon, line 33:
```suggestion
        - /linux/path/to/data_root/
```

docs/user_guide/data_prepare_cat.rst
```diff
@@ -64,11 +66,13 @@ The ``rename``, ``nodata``, ``unit_add`` and ``unit_mult`` options are set per v
 
 The yaml file has an *optional* global **meta** data section:
 
-- **root** (optional): root folder for all the data sources in the yaml file.
+- **roots**: root folders for all the data sources in the yaml file.
```
hboisgon, line 69: Still optional in the case of cloud data for example. 
I would also explain that they actually link to the same folder and are for cross platform support only. Not like a data source per data source check will be done to be able to store data in different folders.

docs/user_guide/data_prepare_cat.rst
```diff
@@ -64,11 +66,13 @@ The ``rename``, ``nodata``, ``unit_add`` and ``unit_mult`` options are set per v
...
   If not provided the folder of where the yaml file is located will be used as root.
   This is used in combination with each data source **path** argument to avoid repetition.
+  The roots listed will be checked in the order they are provided. The first one to be found to exist will be used as the actual root.
+  This can be used for corss platform and corss machiene compatability as can be seen above.
```
hboisgon, line 73:
```suggestion
  This can be used for cross platform and cross machine compatibility as can be seen above.
```

docs/user_guide/data_prepare_cat.rst
```diff
@@ -64,11 +66,13 @@ The ``rename``, ``nodata``, ``unit_add`` and ``unit_mult`` options are set per v
...
 - **version** (recommended): data catalog version; we recommend `calendar versioning <https://calver.org/>`_
-- **hydromt_version** (recommended): range of hydromt version that can read this catalog. Format should be acording to `PEP 440 <https://peps.python.org/pep-0440/#version-specifiers>`_.
+- **hydromt_version**: range of hydromt version that can read this catalog. Format should be acording to `PEP 440 <https://peps.python.org/pep-0440/#version-specifiers>`_.
```
hboisgon, line 75: See the advice of Dirk but I have a preference to keep this as recommended

hboisgon (commented):
Thanks @savente93 !
For the optional/mandatory hydromt_version, I personally prefer optional but let's see what @DirkEilander thinks.
Docs addition looks good too!
One thing to check that popped to mind is if for the cloud catalogs (no root), things still work and the data can be accessed and read. If so I think the rest is good :)

data/catalogs/aws_data.yml
```diff
@@ -1,7 +1,7 @@
 ---
 meta:
   version: 2024.1.30
-  hydromt_version: '>=0.9, <1.0'
+  hydromt_version: '>=1.0a, <2.0'
```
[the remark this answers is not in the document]
savente93: good point, I'll double check that I wrote tests for that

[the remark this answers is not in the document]
savente93: it turns out I didn't and there were some troubles there. I fixed them now but good catch!

docs/changelog.rst
savente93: sure, good idea!

hydromt/_validators/data_catalog.py
savente93: I'm okay with this solution, I'll go with that

data/catalogs/artifact_data.yml
```diff
@@ -266,7 +266,7 @@ gebco:
     unit: m+MSL
   path: gebco.tif
 ghs-smod_2015_v2:
-  crs: 54009
+  crs: ESRI:54009
```
hboisgon, line 269: This is actually issue #696 and I think we should still fix it for main as well. If you need this for tests to pass then we can keep and solve the merge conflicts with main when they come :)

hydromt/_validators/data_catalog.py
```diff
@@ -60,6 +61,11 @@ class DataCatalogMetaData(BaseModel):
 
     @model_validator(mode="after")
     def _check_version_compatible(self) -> "DataCatalogMetaData":
+        if self.hydromt_version is None:
+            warning(
+                f"No hydromt version was specified for the data catalog, thus compatability between used hydromt version ({HYDROMT_VERSION}) and the catalog could not be determined."
```
hboisgon, line 66:
```suggestion
                f"No hydromt version was specified for the data catalog, thus compatibility between used hydromt version ({HYDROMT_VERSION}) and the catalog could not be determined."
```

tests/validators/test_data_catalog_validation.py
hboisgon, line 1: Are these two tests still run in the current test bench? Just to be extra sure the data can be read as well:

https://github.com/Deltares/hydromt/blob/d634534fd1c3c5e470d6835a1ee1c518705576ad/tests/test_data_adapter.py#L98-L126

hboisgon (approved):
Great! Looks good now :)
Just a final couple of comments/typos but else looks good to go.

data/catalogs/artifact_data.yml
savente93: They are indeed necessary for these tests. I'll open a separate PR to fix these in main if you want.

tests/validators/test_data_catalog_validation.py
savente93: I'm not quite sure I understand what the question is, but yes, the tests are run against the files on this branch, though they only test the syntax of the file, not the actually data as the test you linked seems to do

data/catalogs/artifact_data.yml
hboisgon: I think @Tjalling-dejong will or has opened a larger one to update artifact_data catalog (also driver_kwargs should be updated I think)

tests/validators/test_data_catalog_validation.py
hboisgon: If in the current v1 tests, these are included and pass or if they are skipped for some reason. If they are not skipped and pass, then we are sure the data can actually still successfully be read on top of the extra test you implemented in this PR

savente93: they should be included yes

tests/validators/test_data_catalog_validation.py
```diff
@@ -116,22 +139,33 @@ def test_valid_catalog_variants():
...
     d = {
+        "meta": {"hydromt_version": ">=1.0a,<2", "roots": [""]},
         "chelsa": {"alias": "chelsa_v1.2"},
     }
 
     with pytest.raises(AssertionError):
         _ = DataCatalogValidator.from_dict(d)
 
 
+def test_no_hydrmt_version_loggs_warning(caplog):
```
DirkEilander, line 157: nice one! didn't know it was possible to test the logs like this

DirkEilander (approved):
Looks great @savente93!
I've just added a few comments for clarification in the changelog, removing an error message and adding a root to deltares_data.yml. With these changes the PR can be merged I think.

tests/validators/test_data_catalog_validation.py
```diff
@@ -22,6 +22,27 @@ def test_deltares_data_catalog():
     _ = DataCatalogValidator.from_dict(yml_dict)
 
 
+def test_artifact_data_catalog():
```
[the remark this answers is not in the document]
savente93: good one, forgot about that

Issue #778, opened 2024-02-13
Model.close() and/or Model as context manager


DirkEilander:
### Kind of request

None

### Enhancement Description

Currently there is no method to close filehandles to a logfile, input and output data which may prevents moving / deleting files when the python kernel is still open.

### Use case

For instance, when updating models

LuukBlom:
Adding some more context from https://github.com/Deltares/hydromt_wflow/issues/271#issuecomment-2785624853

1. Would be nice to have the DataCatalog be explicit in what it returns (with a kwarg?): 
  - a DataSet/-Array that is still open but has not loaded all from disk.
  - a DataSet/-Array that is closed and loaded from disk.

2. Would be nice: implement file handling functions using a context manager.

3. Would be nice if there would be a function in the model / DataCatalog class that would close all (internally) opened file handles. (how to achieve this is not so clear, maybe with garbage collector or by keeping a list of weak references of all files opened in hydromt? )

Note: xarray is not 100% perfect either. See this [comment](https://github.com/Deltares/hydromt_wflow/issues/271#issuecomment-2783100455) as well:

deltamarnix:
@DirkEilander In #1254 I was able to make a close function for the model that closes all files that are read as part of the model, but I'm not closing the datasets that are opened by the data catalog, since there is low risk that they need to be moved while the python interpreter is open.

If I have to figure out how these datasets are being opened, and need to keep track of them, we have a long way to go and would need a different way to open datasets. Do you think this is sufficient?

PR #788, opened 2024-02-20
Fix logger config in set root

### LINKED ISSUE #779: Add filehandle to logger when setting `ModelRoot.path` in v1
### HydroMT version checks

### Reproducible Example

Setting the model root should create/update the logger file handle

### Current behaviour

We don't do anything with the logger in ModelRoot.path
https://github.com/Deltares/hydromt/blob/8556208b34a72f5a38db4a953864e0998070bf2f/hydromt/models/root.py#L33

### Desired behaviour

Closing the old logger file handle (if exits); copying the old logger file to the new path (new functionality); setting this as filehandler for the logger (see old code). 
When setting a new logger filehandle in overwrite mode, make sure any old hydromt.log file is deleted first (new functionality)
https://github.com/Deltares/hydromt/blob/3edef05bf25231f8ca9a888a5df37dec9ce10100/hydromt/models/model_api.py#L500

### Additional context

_No response_

savente93:
## Issue addressed
Fixes #779 

### Solved
- [x] When a new root is set, the logging to the old root is closed
- [x] old logs are copied over so the logs at the new root remain complete
- [x] Adds a small utility to close all loggers which will pave the way for #778 

### Posponed
None that I am aware of

### To be agreed upon
None that I am aware of


hydromt/models/root.py
```diff
@@ -58,6 +86,30 @@ def is_reading_mode(self) -> bool:
         """Test whether we are in reading mode or not."""
         return self._mode.is_reading_mode()
 
+    # old_root is so we can copy over the old log file if needed.
+    def _create_loggers(self, old_path: Optional[Path] = None, overwrite: bool = False):
+        if not exists(self._path):
+            mkdir(self._path)
```
DirkEilander, line 92: I always use os.makedirs which recursively creates folders, rather than just a subfolder.
It also has an argument `exist_ok` which avoids the if statement

```suggestion
        os.makedirs(self._path, exist_ok=True)
```

hydromt/models/root.py
```diff
@@ -58,6 +86,30 @@ def is_reading_mode(self) -> bool:
...
+
+        log_level = 20  # default, but overwritten by the level of active loggers
+        for i, h in enumerate(self.logger.handlers):
+            log_level = h.level
+            if isinstance(h, FileHandler):
+                if dirname(h.baseFilename) != self._path:
```
DirkEilander, line 98: In case we are already logging to `self._path` there is no need to create a new filehandler right? Otherwise we would be logging the same message twice to the file. If the logfile already exists in the handlers, I think we can exit the method?

DirkEilander (changes requested):
Nice work @savente93!
I've left two small comments.

Something to discuss (and perhaps for another issue): In the current code we can also initialize model without a root to build your model in memory. Would that still be possible in the new code? In that case should we be initialize the ModelRoot without a path, or does the Model class not initialize the ModelRoot.

hydromt/models/root.py
savente93: We agreed that the root should only create the root itself not any of the subfolders, so it is the same in this case

hydromt/models/root.py
savente93: We could exit the loop instead, but if we exit the method we won't overwrite the logs if you just change the mode to overwrite which was part of the functional requirements.

savente93: I added an early exit if the old and new paths are the same since then we don't have to do anything, (and from your comment I infer that in that case we don't have to overwrite the old log file)

hydromt/models/root.py
DirkEilander: We do want to be able to create a new root `"/c/models/sfincs/delft"` in a folder `"/c/models/"`.  With `os.mkdir` you get a FileNotFoundError because `"/c/models/sfincs"` does not yet exist. `os.makedirs` solves this as it recursively creates the requested root folder.

With subfolders I meant that in v0.9 the `set_root` creates several subfolderes within the root based on the `Model._FOLDERS` class property, e.g. `"<root>/geoms"` in WflowModel to save all geoms. The responsibility to create the geoms subfolder should be moved to the geoms write method.

hydromt/models/root.py
```diff
@@ -58,6 +86,33 @@ def is_reading_mode(self) -> bool:
         """Test whether we are in reading mode or not."""
         return self._mode.is_reading_mode()
 
+    # old_root is so we can copy over the old log file if needed.
+    def _create_loggers(self, old_path: Optional[Path] = None, overwrite: bool = False):
```
DirkEilander, line 90: For me it would be more clear to change the name into something like `_update_logger_filehandler` as it does not create a new logger instance.

Also, could you add a one line docstring to clarify what the method does. Something like
"""Update the file handler to save logs in self.path/hydromt.log. If a log file at old_path exists copy it to self.path and append."""

hydromt/models/root.py
savente93: fair enough.

DirkEilander (approved):

PR #790, opened 2024-02-20
remove dict-like features from data catalog

### LINKED ISSUE #550: Using dictionary like features on the data catalog should be removed in 1.0
### Kind of request

Removing existing functionality

### Enhancement Description

Using dictionary like features on the datacatalog such as `if foo in data_catalog:` or `data_catalog[foo]` has been marked as deprecated because it does not offer access to the full variants functionality. To keep a consistent API support for these features should be removed in 1.0 

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #550 

## Explanation
removed what was marked as deprecated

## Checklist


DirkEilander (commented):
Can you remind me why we want to remove these again? I actually think it's quite useful and intuitive to be able to check with `source in data_catalog` and be able to get the default version and variant of a source with `data_catalog[source]`. 

If we do go ahead with this it would be really good to also start creating a migration page in the docs where we explain how to replace old syntax by new syntax. A one line changelog will be too cryptic for many plugin developers I think.

savente93:
First of all because they don't work with the variant and source paradigm, if people use them they will only be able to use the default variant and it might be confusing for users if they then want to access other variants or sources and can't with the same syntax. 

Secondly because we agreed to. It's on the backlog, and the code is marked as deprecated.

DirkEilander:
If everyone agrees it's fine to remove them. 

What do you think about the migration guide? This will save us a lot time later I think.

savente93:
I think a migration guide is a good idea. I'll add it.

savente93:
I've added one on my local repo, but because of the new deltares rules I can't push from my work laptop right now, so I'll push it when I get home

savente93:
@DirkEilander I've added the migration guide, could you take another look?

deltamarnix (approved):
Nice with the changelog and everything. More lines were added than released ;)

PR #791, opened 2024-02-21
Remove support for `.ini` and `.toml` format for configurations

### LINKED ISSUE #548: Support for .ini and .toml configuration files should be removed in v1.0
### Kind of request

None

### Enhancement Description

Support for `ini` files has been marked as deprecated for quite a while now. Therefore the functionality for using it should be removed in the next breaking release (i.e. v1.0)

### Use case

Making sure out communication is consistent with the users. 

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #548 

## Explanation
As planned support for using `.ini` and `.toml` formats should be removed. this code removes the associated code, and adds a migration guide. 

## Checklist


hydromt/io/writers.py
```diff
@@ -34,41 +30,6 @@ def write_xy(fn, gdf, fmt="%.4f"):
...
-        cf = cf()
-    cf.optionxform = str  # preserve capital letter
-    if noheader:  # add dummy header
-        cfdict = {"dummy": cfdict}
-    cf.read_dict(cfdict)
-    with codecs.open(config_fn, "w", encoding=encoding) as fp:
-        cf.write(fp)
-
-
 def configwrite(config_fn: Union[str, Path], cfdict: dict, **kwargs) -> None:
```
DirkEilander, line 33: This method is used to write model simulation configuration files, e.g. wflow.toml, rather than HydroMT config file. 
For HydroMT build/update config files we want to have only yaml, but we can't decide what file format a model kernel requires to configure a simulation. Since this format is toml in case of most recent model kernels, I suggest to keep the write to toml option.

hydromt/io/writers.py
```diff
@@ -92,32 +53,15 @@ def configwrite(config_fn: Union[str, Path], cfdict: dict, **kwargs) -> None:
...
             yaml.dump(_cfdict, f, sort_keys=False)
-    elif ext == ".toml":  # user defined
-        _cfdict = _process_config_out(_cfdict)
-        with open(config_fn, "wb") as f:
-            dump_toml(_cfdict, f)
     else:
-        write_ini_config(config_fn, _cfdict, **kwargs)
-
-
-def _process_config_out(d):
```
DirkEilander, line 109: Why did you remove this method?

hydromt/io/readers.py
```diff
@@ -892,17 +836,3 @@ def parse_values(
                 value = None
             cfdict[section].update({key: value})
     return cfdict
-
-
-def _process_config_in(d):
```
DirkEilander, line 897: Why did you remove this method?

DirkEilander (changes requested):

dalmijn:
Why would we want to remove this functionality? 
It looks like the maintainability of this would be fairly easy, as it's simply parsing the formats. 
Wflow users are mostly used to the toml format as its used as the settings file format. 
For them, using toml could be more intuitive.
I'd say having multiple formats is a plus.

hydromt/io/writers.py
savente93: Okay, it is currently really unclear for me which file types should be supported where. Could you please make a table for me with the files we should be able to read and write and which file types should be supported? (excluding data files, that is the adapter part) because I currently can't tell up from down.

hydromt/io/writers.py
savente93: because this was only necessary for toml support (since toml doesn't support `null` values, or a variant thereof) and I thought that was supposed to be removed. if `toml` stays then the function will stay too

hydromt/io/writers.py
DirkEilander: These are the files:
- HydroMT config: **yaml**
- Model simulation config files:  **toml** by default, but plugins my implement their own format. 
- HydroMT data catalog **yaml**, but this file has its own read and write methods in `DataCatalog` 

So far, we have used the `configread` and `configwrite` functions for both type of config files as these used to be in same format (_ini_) and require the same processing of relative to absolute paths and vice versa. Perhaps its best to move the read and write methods for model simulation configuration **toml** files to the future `ConfigModelComponent` class. 

In that case  `configread` and `configwrite` only need to support HydroMT config **yaml** files.
So basically we need to make a new issue for a `ConfigModelComponent` class to support **toml** and this  PR is fine. Do you agree?

hydromt/io/writers.py
DirkEilander: This method was also called for yaml, not only toml, that's why I asked. I also don't know (anymore) why the method is required in that case though. But I would suggest to keep it until we know its redundant.

hydromt/io/writers.py
savente93: yes I think that would be good. I'll make the issue. Could you approve then?

hydromt/io/writers.py
savente93: The reason it's there is that since TOML does not support some kind of `null` value, (the way you denote this in toml is simply by omitting the key). To make sure that we didn't lose any keys a special "NONE" value was inserted including in the YAML files, so when reading the YAML files back out agian these had to be removed as they were not necessary. 

git blame reveals that this was added in #444 (adding support for TOML files), so before it ran fine, and the tests pass without it so I'd say that's enough evidence for it  being redundant.

hydromt/io/writers.py
savente93: see #799

hydromt/io/writers.py
```diff
@@ -92,32 +53,15 @@ def configwrite(config_fn: Union[str, Path], cfdict: dict, **kwargs) -> None:
...
-            if v is None:
-                ret[k] = "NONE"
-            else:
-                ret[k] = _process_config_out(v)
-    else:
-        ret = d
-
-    return ret
+        raise ValueError(
+            f"Could not write to unknown extention: {ext} hydromt only supports yaml"
```
deltamarnix, line 63:
```suggestion
            f"Could not write to unknown extension: {ext} hydromt only supports yaml"
```

deltamarnix: I would opt to add spell checker in your visual studio code extensions.

savente93: I don't use visual studio code, but I'll see if I can find one for my editor

hydromt/cli/_utils.py
```diff
@@ -85,19 +83,15 @@ def parse_json(ctx, param, value: str) -> Dict[str, Any]:
...
     opt = {}
     if path is not None and isfile(path):
         if str(path).endswith(".ini"):
-            warn(
-                "Support for .ini configuration files will be deprecated",
-                PendingDeprecationWarning,
-                stacklevel=2,
+            raise ValueError(
+                "Support for ini config files has been removed. Please convert to yaml."
             )
```
deltamarnix, line 95: Consider removing this. It's the same as a deprecation warning. Or consider throwing an error on everything that's not yaml.

hydromt/io/path.py
```diff
@@ -21,13 +21,13 @@ def _relpath(value, root):
...
     for key, val in cfdict.items():
         if isinstance(val, dict):
-            cfdict[key] = parse_relpath(val, root)
+            cfdict[key] = make_config_paths_relative(val, root)
         else:
             cfdict[key] = _relpath(val, root)
     return cfdict
 
 
-def parse_abspath(
```
deltamarnix, line 30: Is there a test for this function?

hydromt/cli/_utils.py
savente93: good point

hydromt/workflows/forcing.py
```diff
@@ -86,7 +86,7 @@ def precip(
     return p_out
 
 
-# use dem_model (from staticmaps) and dem_forcing (meteo) in ini file
+# use dem_model (from staticmaps) and dem_forcing (meteo) in yml file
```
deltamarnix, line 89:
```suggestion
# use dem_model (from staticmaps) and dem_forcing (meteo) in config file
```

hydromt/models/api.py
```diff
@@ -323,7 +323,7 @@ def update(
             opt.pop(method)  # remove from opt
             self.logger.warning(f'"{method}" can only be called when building a model.')
 
-        # loop over other methods from ini file
+        # loop over other methods from yaml file
```
deltamarnix, line 326:
```suggestion
        # loop over other methods from config file
```

hydromt/cli/_utils.py
```diff
@@ -85,19 +83,15 @@ def parse_json(ctx, param, value: str) -> Dict[str, Any]:
 
 
 ### general parsing methods ##
-
-
 def parse_config(
     path: Optional[Union[Path, str]] = None, opt_cli: Optional[Dict] = None
 ) -> Dict:
-    """Parse config from ini `path` and combine with command line options `opt_cli`."""
+    """Parse config from yaml `path` and combine with command line options `opt_cli`."""
```
deltamarnix, line 89:
```suggestion
    """Parse config from `path` and combine with command line options `opt_cli`."""
```

hydromt/io/writers.py
```diff
@@ -92,32 +53,15 @@ def configwrite(config_fn: Union[str, Path], cfdict: dict, **kwargs) -> None:
...
-            if v is None:
-                ret[k] = "NONE"
-            else:
-                ret[k] = _process_config_out(v)
-    else:
-        ret = d
-
-    return ret
+        raise ValueError(
+            f"Could not write to unknown extension: {ext} hydromt only supports yaml"
```
deltamarnix, line 63: Did you add a test for this case?

hydromt/models/api.py
```diff
@@ -199,10 +199,10 @@ def build(
         r"""Single method to build a model from scratch based on settings in `opt`.
 
         Methods will be run one by one based on the order of appearance in `opt`
-        (.ini configuration file). All model methods are supported including
+        (.yaml configuration file). All model methods are supported including
```
deltamarnix, line 202:
```suggestion
        (configuration file). All model methods are supported including
```

hydromt/io/path.py
savente93: not directly but it's included in https://github.com/Deltares/hydromt/blob/84aafa49bc5f73359d5893da610e33e8931956d7/tests/models/test_config.py#L10 
I can make a separate one if you like

hydromt/models/api.py
```diff
@@ -199,10 +199,10 @@ def build(
...
         setup\_\*, read\_\* and write\_\* methods.
 
-        If a write\_\* option is listed in `opt` (ini file) the full writing of the
+        If a write\_\* option is listed in `opt` (yaml file) the full writing of the
```
deltamarnix, line 205:
```suggestion
        If a write\_\* option is listed in `opt` (configuration file) the full writing of the
```

hydromt/models/api.py
```diff
@@ -214,7 +214,7 @@ def build(
             Write complete model after executing all methods in opt, by default True.
         opt: dict, optional
             Model build configuration. The configuration can be parsed from a
-            .ini file using :py:meth:`~hydromt.io.readers.configread`.
+            .yaml file using :py:meth:`~hydromt.io.readers.configread`.
```
deltamarnix, line 217:
```suggestion
            configuration file using :py:meth:`~hydromt.io.readers.configread`.
```

hydromt/models/api.py
```diff
@@ -262,10 +262,10 @@ def update(
         r"""Single method to update a model based the settings in `opt`.
 
         Methods will be run one by one based on the order of appearance in `opt`
-        (ini configuration file).
+        (yaml configuration file).
```
deltamarnix, line 265:
```suggestion
        (configuration file).
```

hydromt/models/api.py
```diff
@@ -262,10 +262,10 @@ def update(
...
 
         All model methods are supported including setup\_\*, read\_\* and write\_\* methods.
-        If a write\_\* option is listed in `opt` (ini file) the full writing of the model
+        If a write\_\* option is listed in `opt` (yaml file) the full writing of the model
```
deltamarnix, line 268:
```suggestion
        If a write\_\* option is listed in `opt` (configuration file) the full writing of the model
```

hydromt/models/api.py
```diff
@@ -278,7 +278,7 @@ def update(
             Write the updated model schematization to disk. By default True.
         opt: dict, optional
             Model build configuration. The configuration can be parsed from a
-            .ini file using :py:meth:`~hydromt.io.readers.configread`.
+            .yaml file using :py:meth:`~hydromt.io.readers.configread`.
```
deltamarnix, line 281:
```suggestion
            configuration file using :py:meth:`~hydromt.io.readers.configread`.
```

hydromt/cli/_utils.py
```diff
@@ -85,20 +83,12 @@ def parse_json(ctx, param, value: str) -> Dict[str, Any]:
 
 
 ### general parsing methods ##
-
-
-def parse_config(
+def _parse_config(
```
deltamarnix, line 86: Can you explain how this function is a private function but is still exported via `__all__`? What is the purpose of making this a private function?

deltamarnix (commented):
I have a question about private functions and modules :)

hydromt/cli/_utils.py
savente93: The exported via `__all__` is my mistake. whether this is a private or a public functions is a bit debatable since it's in a private sub module. This function used to be public, but I think it's a good idea to start clamping down the internal stuff like cli and config parsing a bit more. I just didn't remember to apply it consistently. What do you think? Should this be applied to all functions here or should I revert it to public?

deltamarnix: I think that the `_utils` already says enough about the privacy of the module. Meaning that the functions themselves can just be public. That makes them easier to find in alphabetical order.

DirkEilander (approved):
LGTM!

PR #792, opened 2024-02-21
Mesh component

Tjalling-dejong:
## Issue addressed
Adresses issue #671 
Implementation of the ModelComponent class for the mesh component.

## Explanation

## Checklist


Tjalling-dejong:
@hboisgon I am encountering an issue with the get_mesh() method, I saw through git blame that you made the last commit to this code. I have not changed the code. Now that we are using XUgrid version 0.0.9 the code breaks. I did not manage to figure it out if the bug is from our side or in XUgrid. Could you maybe have a look?

hboisgon:
> @hboisgon I am encountering an issue with the get_mesh() method, I saw through git blame that you made the last commit to this code. I have not changed the code. Now that we are using XUgrid version 0.0.9 the code breaks. I did not manage to figure it out if the bug is from our side or in XUgrid. Could you maybe have a look?

I see we get the same error in the hydromt_delft3dfm test. I'll have a look but in main and not v1

hydromt/models/__init__.py
```diff
@@ -8,9 +8,8 @@
 from hydromt.models.components.vector import VectorModel
 from hydromt.models.plugins import ModelCatalog
 
-if HAS_XUGRID:
-    from hydromt.models.components.mesh import MeshModel
-
+# if HAS_XUGRID:
+#     from hydromt.models.components.mesh import MeshModel
```
deltamarnix, line 12: Remove?

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
 
-    ## general setup methods
-    def setup_mesh2d_from_rasterdataset(
+    def set(
+        self,
+        data: Union[xu.UgridDataArray, xu.UgridDataset],
+        name: Optional[str] = None,
+        grid_name: Optional[str] = None,
+        overwrite_grid: bool = False,
+    ) -> None:
```
deltamarnix, line 42:
```suggestion
    def set(
        self,
        data: Union[xu.UgridDataArray, xu.UgridDataset],
        *,
        name: Optional[str] = None,
        grid_name: Optional[str] = None,
        overwrite_grid: bool = False,
    ) -> None:
```

I would make all naming mandatory

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+        if crs:  # Restore crs
+            for grid in self.data.ugrid.grids:
+                grid.set_crs(crs)
+
+        # # update related geoms if necessary: region TODO: Check if is still needed
+        # if overwrite_grid or new_grid:
+        #     # add / updates region
+        #     if "region" in self.geoms:
+        #         self._geoms.pop("region", None)
+        #     _ = self.region
```
deltamarnix, line 89: Is it still needed? ;)

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+        """
+        # Checks on data
+        data = _check_UGrid(data, name)
+
+        # Checks on grid topology
+        # TODO: check if we support setting multiple grids at once. For now just one
+        if len(data.ugrid.grids) > 1:
+            raise ValueError(
+                "set_mesh methods only supports adding data to one grid at a time."
+            )
```
deltamarnix, line 69: What would that entail? I can imagine that we want one grid per component.

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+        if grid_name is None:
+            grid_name = data.ugrid.grid.name
+        elif grid_name != data.ugrid.grid.name:
+            data = rename_mesh(data, name=grid_name)
+
+        # new_grid = grid_name not in self.mesh_names
```
deltamarnix, line 75: Remove?

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+        name: str, optional
+            Name of new object layer, this is used to overwrite the name of
+            a UgridDataArray.
+        grid_name: str, optional
+            Name of the mesh grid to add data to. If None, inferred from data.
+            Can be used for renaming the grid.
+        overwrite_grid: bool, optional
+            If True, overwrite the grid with the same name as the grid in self.mesh.
+        """
+        # Checks on data
```
deltamarnix, line 61: I expected the first call to be `_initialize_mesh()`, just like other components.

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+            True.
+        **kwargs : dict
+            Additional keyword arguments to be passed to the
+            `xarray.Dataset.to_netcdf` method.
+        """
+        if self.data is None:
+            self.logger.debug("No mesh data found, skip writing.")
+            return
+        if not self.model_root.is_writing_mode():
+            raise IOError("Model opened in read-only mode")
```
deltamarnix, line 116:
```suggestion
    self._model_root._assert_write_mode()
```

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+            filename relative to model root, by default 'mesh/mesh.nc'
+        crs : CRS or int, optional
+            Coordinate Reference System (CRS) object or EPSG code representing the
+            spatial reference system of the mesh file. Only used if the CRS is not
+            found when reading the mesh file.
+        **kwargs : dict
+            Additional keyword arguments to be passed to the `read_nc` method.
+        """
+        if not self.model_root.is_reading_mode():
+            raise IOError("Model not opend in read mode")
```
deltamarnix, line 149:
```suggestion
    self._model_root._assert_read_mode()
```

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+        ds = xr.merge(
+            read_nc(
+                fn, root=self.model_root.path, logger=self.logger, **kwargs
```
deltamarnix, line 152:
```suggestion
                fn, root=self._model_root.path, logger=self._logger, **kwargs
```

Do you have tests for these, or should you update from the refactor-models branch?

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+                    "no crs is found in the file nor passed to the reader."
+                )
+            else:
+                uds.ugrid.set_crs(crs)
+                self.logger.info(
+                    "no crs is found in the file, assigning from user input."
+                )
+        self._data = uds
+
+    def create(
```
deltamarnix, line 171: In my branch `region-component-2`, I am trying to get rid of the `region: dict`. I think this will influence this piece of code as well.

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+
+    @property
+    def region(self) -> gpd.GeoDataFrame:
+        """Returns geometry of region of the model area of interest based on mesh total bounds."""  # noqa: E501
+        region = gpd.GeoDataFrame()
+        if self.data is not None:
+            region = gpd.GeoDataFrame(
+                geometry=[box(*self.data.ugrid.total_bounds)], crs=self.crs
+            )
+        return region
```
deltamarnix, line 286: Shouldn't this be handled by the `region` component instead?

hydromt/models/components/mesh.py
```diff
@@ -10,34 +10,389 @@
...
+
+        else:
+            return self.mesh_grids[grid_name]
+
+    def add_data_from_rasterdataset(
         self,
         raster_fn: Union[str, Path, xr.DataArray, xr.Dataset],
         grid_name: Optional[str] = "mesh2d",
         variables: Optional[list] = None,
         fill_method: Optional[str] = None,
```
deltamarnix, line 394:
```suggestion
    def add_data_from_rasterdataset(
        self,
        *,
        raster_fn: Union[str, Path, xr.DataArray, xr.Dataset],
        grid_name: Optional[str] = "mesh2d",
        variables: Optional[list] = None,
        fill_method: Optional[str] = None,
```

tests/conftest.py
```diff
@@ -363,3 +367,15 @@ def artifact_data():
     datacatalog = DataCatalog()
     datacatalog.from_predefined_catalogs("artifact_data")
     return datacatalog
+
+
+@pytest.fixture()
+def mock_model(tmpdir):
```
deltamarnix, line 373: There are already some other mocks for a Model. Not sure if you want to put this in conftest, as it will be different per test case.

tests/conftest.py
```diff
@@ -363,3 +367,15 @@ def artifact_data():
...
+    logger = logging.getLogger(__name__)
+    logger.propagate = True
+    model = create_autospec(Model)
+    model.root = ModelRoot(path=tmpdir)
+    model.data_catalog = DataCatalog()
+    model.region = ModelRegionComponent(model=model)
```
deltamarnix, line 379: Let's mock the other parts as well

tests/conftest.py
deltamarnix, line 373: Try using `def mock_model(tmpdir, mocker: MockerFixture):`, so that you can use the `pytest-mock` wrapper for making your mocks.

deltamarnix (commented):
This is my initial review. I will probably come with more in a second round, but that also depends on my own experiences with creating components.

Let's have another look at how we mock in our repository. I prefer using `pytest-mock`, where we make use of `MockerFixture`. In that case we don't need to write things like: `with patch(...):`.

hydromt/models/plugins.py
```diff
@@ -13,7 +13,6 @@
 # local generic models
 LOCAL_EPS = {
     "vector_model": "hydromt.models.components.vector:VectorModel",
-    "mesh_model": "hydromt.models.components.mesh:MeshModel",
     "network_model": "hydromt.models.components.network:NetworkModel",
 }
```
savente93, line 17: I can see this code is not up to date with `model-refactor`. if we're going to wait to get that in before you merge this you might want to merge it because I can already see it's going to cause a few merge issues here.

hydromt/models/components/mesh.py
```diff
@@ -199,113 +554,41 @@ def setup_mesh2d_from_raster_reclass(
...
             if grid_name in self.mesh_names:
                 # check if the two grids are the same
-                if (
-                    not self.mesh_grids[grid_name]
-                    .to_dataset()
-                    .equals(data.ugrid.grid.to_dataset())
-                ):
-                    if not overwrite_grid:
+                if not self._grid_is_equal(grid_name, data):
+                    if not overwrite_grid:  # TODO: add is_overwrite_mode
```
savente93, line 591: Fix todo, or make issue?

savente93 (commented):
Nice tests. I have a few small comments but once it's up to date it should be very close.

tests/conftest.py
Tjalling-dejong: I'm using this fixture in test_grid_component as well. Should I use I put this fixture in both test_mesh_component.py and test_grid_component.py?

hydromt/models/components/mesh.py
Tjalling-dejong: I don't know. The TODO was already there in the original code and left it in for now.

hydromt/models/components/mesh.py
Tjalling-dejong: I think it is done this way because you cannot create an empty UgridDataset

hydromt/components/mesh.py
```diff
@@ -11,33 +11,370 @@
...
+from hydromt.io.readers import read_nc
+from hydromt.models import Model
+from hydromt.workflows.mesh import (
+    create_mesh2d,
+    mesh2d_from_raster_reclass,
+    mesh2d_from_rasterdataset,
+    rename_mesh,
+)
 
-__all__ = ["MeshModel"]
```
hboisgon, line 18: I think you still need the __all__ but his time with MeshModelComponent.
@savente93 previously xugrid was an optional package. How does this work now with components discovery? Can we still make this one discovered only if xugrid is installed? Or should in the end xugrid become a mandatory dependency of hydromt ?

hydromt/components/mesh.py
```diff
@@ -11,33 +11,370 @@
...
+        elif grid_name != data.ugrid.grid.name:
+            data = rename_mesh(data, name=grid_name)
+
+        crs = self._add_mesh(
+            data=data, grid_name=grid_name, overwrite_grid=overwrite_grid
+        )
+
+        if crs:  # Restore crs
+            for grid in self.data.ugrid.grids:
+                grid.set_crs(crs)
```
hboisgon, line 81: Nice that you did a new function, I would however also include updating crs in that function as well instead of here.

hydromt/components/mesh.py
```diff
@@ -11,33 +11,370 @@
...
+
+    def write(
```
hboisgon, line 83: You'll need to add ``@hydromt_step`` decorator to all methods available for build and update ie read/write/create/add_data

hydromt/components/mesh.py
```diff
@@ -11,33 +11,370 @@
...
+                    "no crs is found in the file nor passed to the reader."
+                )
+            else:
+                uds.ugrid.set_crs(crs)
+                self._logger.info(
+                    "no crs is found in the file, assigning from user input."
+                )
+        self._data = uds
+
+    def create(
```
hboisgon, line 162: This is to create a 2D mesh so I would keep it in the function name. the current MeshComponent handles multiple grids which can be 1D or 2D or 3D.

```suggestion
    def create2d(
```

hydromt/components/mesh.py
```diff
@@ -11,33 +11,370 @@
...
+                uds.ugrid.grid.set_crs(grid.crs)
+            else:
+                uds = self.data.copy()
+
+            return uds
+
+        else:
+            return self.mesh_grids[grid_name]
+
+    def add_data_from_rasterdataset(
```
hboisgon, line 370:
```suggestion
    @hydromt_step
    def add_2ddata_from_rasterdataset(
```

hydromt/components/mesh.py
```diff
@@ -11,33 +11,370 @@
...
+        self,
+        *,
```
hboisgon, line 372: Maybe the * can go after raster_fn? We're already not happy with the name :)

hydromt/components/mesh.py
```diff
@@ -11,33 +11,370 @@
...
         raster_fn: Union[str, Path, xr.DataArray, xr.Dataset],
         grid_name: Optional[str] = "mesh2d",
         variables: Optional[list] = None,
         fill_method: Optional[str] = None,
-        resampling_method: Optional[Union[str, List]] = "centroid",
+        resampling_method: Optional[str] = "centroid",
```
hboisgon, line 377: It can still be a list right? In case method is different per variable in ``variables``

hydromt/components/mesh.py
```diff
@@ -78,35 +415,35 @@ def setup_mesh2d_from_rasterdataset(
...
+            logger=self._logger,
         )
 
-        self.set_mesh(uds_sample, grid_name=grid_name, overwrite_grid=False)
+        self.set(uds_sample, grid_name=grid_name, overwrite_grid=False)
 
         return list(uds_sample.data_vars.keys())
 
-    def setup_mesh2d_from_raster_reclass(
+    def add_data_from_raster_reclass(
```
hboisgon, line 446:
```suggestion
    @hydromt_step
    def add_2ddata_from_raster_reclass(
```

hydromt/components/mesh.py
```diff
@@ -195,117 +532,45 @@ def setup_mesh2d_from_raster_reclass(
...
             if not data.ugrid.grid.crs:
                 raise ValueError("Data should have CRS.")
-            self._mesh = data
+            self._data = data
+            return None
         else:
             # Check on crs
             if not data.ugrid.grid.crs == self.crs:
-                raise ValueError("Data and self.mesh should have the same CRS.")
+                raise ValueError("Data and self.data should have the same CRS.")
```
hboisgon, line 566:
```suggestion
                raise ValueError("Data and Mesh should have the same CRS.")
```

hydromt/components/mesh.py
```diff
@@ -11,33 +11,370 @@
...
+        else:
+            return None
+
+    @property
+    def bounds(self) -> Optional[Dict]:
+        """Returns model mesh bounds."""
+        if self.data is not None:
+            return self.data.ugrid.bounds
+        return None
+
```
hboisgon, line 268: Since you removed ``region`` maybe you could create an equivalent ``mesh_region`` property? 
```suggestion
    @property 
    def mesh_region(self) -> Optional[gpd.GeoDataFrame]:
        """Return mesh total_bounds as a geodataframe."""
        if self.data is not None:
            region = gpd.GeoDataFrame(
                geometry=[box(*self.data.ugrid.total_bounds)], crs=self.crs
            )
            return region
        return None
```

hboisgon (changes requested):
Nice work @Tjalling-dejong especially with the extra testing!
A few comments to address before merging but the refactor PR changed a little again so I think you'll have to wait for this one to be merged first before you can address everything.

hydromt/models/components/mesh.py
hboisgon: No but then I suggested to add a new ``mesh_region`` property to replace the old ``region`` one. It will be quite useful to update ``MeshComponentRegion`` as we discussed.

hydromt/models/components/mesh.py
hboisgon: For now ``MeshComponent`` is still multi-grid. We can see later if we want to add single grid Component like ``Mesh1dComponent``, ``Mesh2dComponent`` and ``Mesh3dComponent``. The TODO can be removed and open as an issue if anyone complains. It's just that for the set methods I only allow to update data or grid schematization one grid at a time.

hydromt/models/components/mesh.py
hboisgon: Indeed! So the choice would either be xr.Dataset or None but I think None is easier and clearer to handle.

hydromt/models/components/mesh.py
hboisgon: Why was the is_overwrite_mode added? Calling the whole ``set`` method means you are overwriting anyway if not the grid then the data...

hydromt/components/mesh.py
Tjalling-dejong: XUgrid will become a mandatory dependency in v1

hydromt/components/mesh.py
Tjalling-dejong: I based the typing of the typing in mesh2d_from_rasterdataset. There it is Optional[str], but I see in the docstring now that a list is also possible. I'll correct it in mesh2d_from_rasterdataset as well

hydromt/components/mesh.py
```diff
@@ -12,33 +12,376 @@
...
+from hydromt.workflows.mesh import (
+    create_mesh2d,
+    mesh2d_from_raster_reclass,
+    mesh2d_from_rasterdataset,
+    rename_mesh,
+)
 
-__all__ = ["MeshModel"]
-logger = logging.getLogger(__name__)
+logger = getLogger(__name__)
```
deltamarnix, line 26: I believe that this `logger` can go

hydromt/models/components/mesh.py
Tjalling-dejong: I removed is_overwrite_mode. Not sure why I added that

hydromt/components/mesh.py
```diff
@@ -12,33 +12,376 @@
...
+            data=data, grid_name=grid_name, overwrite_grid=overwrite_grid
+        )
+
+        if crs:  # Restore crs
+            for grid in self.data.ugrid.grids:
+                grid.set_crs(crs)
+
+    def write(
+        self,
+        fn: str = "mesh/mesh.nc",
```
deltamarnix, line 86: I would opt for a `DEFAULT_FN = "mesh/mesh.nc` at the top of this file and use it in `fn` for both the `read()` and `write()` functions.

hydromt/components/mesh.py
```diff
@@ -12,33 +12,376 @@
...
+        """
+        Model static mesh data. It returns a xugrid.UgridDataset.
+
+        Mesh can contain several grids (1D, 2D, 3D) defined according
+        to UGRID conventions. To extract a specific grid, use get_mesh
+        method.
+        """
+        # XU grid data type Xarray dataset with xu sampling.
+        if self._data is None and self._root.is_reading_mode():
+            self.read()
```
deltamarnix, line 238: I miss this `_initialize_mesh` function that we also see in the grid component. Would be nice to have it consistent.

hydromt/components/mesh.py
```diff
@@ -12,33 +12,376 @@
...
+            crs = None
+            for _k, v in grid_crs.items():
+                if crs is None:
+                    crs = v
+                if v == crs:
+                    continue
+                else:
+                    raise ValueError(
+                        f"Mesh crs is not uniform, please check {grid_crs}"
+                    )
```
deltamarnix, line 256: I would expect this in a setter, not in a getter.

hydromt/components/mesh.py
```diff
@@ -12,33 +12,376 @@
...
+            )
+            return region
+        return None
+    @property
+    def mesh_names(self) -> List[str]:
+        """List of grid names in mesh."""
+        if self.data is not None:
+            return [grid.name for grid in self.data.ugrid.grids]
+        else:
+            return []
```
deltamarnix, line 282: Would be nice if your editor did autoformatting. We have enabled that in vs code settings. Perhaps you have some overwritten settings in your vs code personal settings. I expect some new lines in between

hydromt/components/mesh.py
deltamarnix: It almost looks like these are two different functions. One for a file, one for an actual dataset

hydromt/components/mesh.py
Tjalling-dejong: Weird that it did not autoformat this time

hydromt/components/mesh.py
Tjalling-dejong: I think this feature can go. There is a check in _add_mesh that checks if the crs is uniform.

Tjalling-dejong:
@deltamarnix @hboisgon @savente93  Thanks for the feedback!
I added your suggestions and updated code + tests. 
There are still multiple breaking tests but these are related to plugins and circular imports. I will look into that tomorrow.

docs/dev/migrating-to-v1.rst
```diff
@@ -240,6 +240,7 @@ In the core of HydroMT, the available components are:
 | model.tables          | TablesComponent  | Component for managing non-geospatial data in pandas DataFrames  |
 | model.grid            | GridComponent   | Component for managing regular gridded data in single hydromt RasterDataset    |
 | model.geoms('region') | RegionComponent | Component for managing the area of interest for the model in a geopandas GeoDataFrame.    |
+| model.mesh			| MeshComponent	  | Component for managing unstructured grids as a hydromt RasterDataset
```
hboisgon, line 243: Table format looks off. Can you check? All columns length should match ie "|" should be aligned. If it was because of my suggestion then it's because I cant easily check alignment in this github view sorry..

docs/dev/migrating-to-v1.rst
```diff
@@ -355,22 +356,36 @@ has not been changed compared to the GridModel.
...
+| model.read_grid(...)         | model.grid.read(...)            |
 +------------------------------+-------------------------------------------+
-| model.write_grid(...)        | model.grid_component.write(...)           |
+| model.write_grid(...)        | model.grid.write(...)           |
 +------------------------------+-------------------------------------------+
-| model.setup_grid(...)        | model.grid_component.create(...)          |
+| model.setup_grid(...)        | model.grid.create(...)          |
 +------------------------------+-------------------------------------------+
-| model.setup_grid_from_*(...) | model.grid_component.add_data_from_*(...) |
+| model.setup_grid_from_*(...) | model.grid.add_data_from_*(...) |
```
hboisgon, line 367: Same here for table alignement

hydromt/components/mesh.py
```diff
@@ -1,40 +1,384 @@
...
+__all__ = ["MeshComponent"]
 
-    ## general setup methods
-    def setup_mesh2d_from_rasterdataset(
+
+class MeshComponent(ModelComponent):
+    """ModelComponent class for mesh components."""
+
+    def __init__(self, model: "Model"):
+        super().__init__(model=model)
```
hboisgon, line 39: We just discussed yesterday that we want to pass a default filename in the component __init__ and then use None in the argument for read and write. Can you update?
You can check out the TablesComponent PR

hydromt/components/mesh.py
```diff
@@ -1,40 +1,384 @@
...
+        """
+        Model static mesh data. It returns a xugrid.UgridDataset.
+
+        Mesh can contain several grids (1D, 2D, 3D) defined according
+        to UGRID conventions. To extract a specific grid, use get_mesh
+        method.
+        """
+        # XU grid data type Xarray dataset with xu sampling.
+        if self._data is None:
+            self._initialize_mesh()
```
hboisgon, line 241:
```suggestion
            self._initialize()
```

hydromt/components/mesh.py
```diff
@@ -1,40 +1,384 @@
...
+        return self._data
+
+    def _initialize_mesh(self, skip_read: bool = False) -> None:
+        """Initialize mesh object."""
+        if self._data is None:
+            self._data = xu.UgridDataset(xr.Dataset())
```
hboisgon, line 247: Does this work now? Then nice we can initialize even with an empty grid :)
You'll need to add an initialize step in the read function. You can check maybe TablesComponent to see how initialize is used

hydromt/components/mesh.py
```diff
@@ -1,40 +1,384 @@
...
+            if self._root.is_reading_mode() and not skip_read:
+                self.read()
+
+    @property
+    def crs(self) -> Optional[CRS]:
+        """Returns model mesh crs."""
+        if self.data is not None:
```
hboisgon, line 254: Maybe you need to update all the properties now no? Because with initialize you will get an empty xugrid object and not None anymore for data. Check this also works for empty xugrid object.
Please update all properties like bounds etc accordingly

hydromt/components/mesh.py
```diff
@@ -1,40 +1,384 @@
...
+
+    @property
+    def bounds(self) -> Optional[Dict]:
+        """Returns model mesh bounds."""
+        if self.data is not None:
+            return self.data.ugrid.bounds
+        return None
+
+    @property
+    def mesh_region(self) -> Optional[gpd.GeoDataFrame]:
```
hboisgon, line 267: I think this will be named region again in @deltamarnix PR on regions so maybe you could change the name already or check with him how to best update?

docs/api.rst
```diff
@@ -553,6 +553,59 @@ General methods
...
+General methods
+---------------
+
+.. autosummary::
+   :toctree: _generated
+
+   MeshComponent.set
+   MeshComponent.write
+   MeshComponent.read
+
```
hboisgon, line 595:
```suggestion
   MeshComponent.read
   MeshComponent.get_mesh
```

hydromt/components/mesh.py
```diff
@@ -1,40 +1,384 @@
...
+            else:
+                uds = self.data.copy()
+
+            return uds
+
+        else:
+            return self.mesh_grids[grid_name]
+
+    @hydromt_step
+    def add_data_from_rasterdataset(
```
hboisgon, line 378: I still miss in the name that this method works for 2D grids and not 1D and 3D. I previously suggested ``add_2ddata_from_rasterdataset``. I'm never good with names but I still think we should highlight this fact somehow in the name.

hboisgon (changes requested):
Almost there @Tjalling-dejong ! Nice that you could add an initialize function for mesh! But I think you still need to update a couple of things to fully take it into account.
I wonder if we are not missing model update tests in our tests benches? Ie model opened in r+ mode and check if the existing mesh/grid/other components are indeed read properly before being updated. If these tests are in place please ignore.

hydromt/components/mesh.py
Tjalling-dejong: I renamed it to `add_2d_data_from_rasterdataset`

tests/components/test_mesh_component.py
```diff
@@ -166,6 +180,33 @@ def test_read(mock_model, caplog, tmpdir, griduda):
...
+    mesh._data = None
+    # Test with sample data
+    data = xu.data.elevation_nl().to_dataset()
+    data.grid.crs = 28992
+    component.set(data=data, grid_name="elevation_mesh")
+    assert "elevation_mesh" in component.mesh_names
+    assert component.data == data
+    component.write()
+    component._data = None
+    component.read()
```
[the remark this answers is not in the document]
hboisgon:
```suggestion
    component._data = None
    component.read()
    assert component.data == data
    
    m2 = Model(root=str(tmpdir), mode="r+")
    # Sample data
    data = xu.data.elevation_nl().to_dataset()
    data.grid.crs = 28992
    data.name = "elevation_v2"
    m.mesh.set(data=data, grid_name="elevation_mesh")
    
    # check both data variables are still there ie mesh was read before being updated
    assert "elevation_v2" in m2.mesh.data
    assert "elevation" in m2.mesh.data # I think this is the correct name?
    
```

docs/dev/migrating-to-v1.rst
```diff
@@ -362,19 +367,20 @@ has not been changed compared to the GridModel.
...
 +------------------------------+-------------------------------------------+
-| model.write_grid(...)        | model.grid.write(...)           |
+| model.write_grid(...)        | model.grid.write(...)            		   |
 +------------------------------+-------------------------------------------+
-| model.setup_grid(...)        | model.grid.create(...)          |
+| model.setup_grid(...)        | model.grid.create(...)          		   |
 +------------------------------+-------------------------------------------+
-| model.setup_grid_from_*(...) | model.grid.add_data_from_*(...) |
+| model.setup_grid_from_*(...) | model.grid.add_data_from_*(...) 		   |
 +------------------------------+-------------------------------------------+
```
hboisgon, line 379: Alignment still looks off? Or is it just this view?

hboisgon (changes requested):
Thanks for the changes @Tjalling-dejong. Almost there but still a couple of steps to add to fully take advantage of the new ``_initialize`` function.
I also explained the test I wanted in a little more details, sorry it was not clear enough.

If you update the initialize part then we are good to merge :)

docs/dev/migrating-to-v1.rst
Tjalling-dejong: It is just the view, if you check this file in the branch the table looks fine

hboisgon (approved):
Looking all good now, thanks @Tjalling-dejong :)

PR #829, opened 2024-03-06
Fix warnings

savente93:
## Issue addressed
Fixes NA

## Explanation
I noticed that our current test suite had some warnings so I decided to fix them in a separate PR. I actually think that warnings are important to pay attention to since they do often indicate problems. I've fixed them now for the current test suite and if it gets merged I would like to argue for trying to keep the test suite warning free going forward. Of course, sometimes there are issues we don't have control over, and in that case we can add them to the filters in the `pyproject.toml` but then at least we're making concussions decisions.

## Checklist

### DIFF
--- a/hydromt/gis/raster.py
+++ b/hydromt/gis/raster.py
@@ -2290,9 +2290,9 @@ class RasterDataArray(XRasterBase):
         kwargs.update(dict(method=method, extrapolate=extrapolate))
         if dim0:
             interp_data = np.empty(self._obj.shape, dtype=self._obj.dtype)
-            for i, (_, sub_xds) in enumerate(self._obj.groupby(dim0)):
+            for i, (_, sub_xds) in enumerate(self._obj.groupby(dim0, squeeze=False)):
                 interp_data[i, ...] = self._interpolate_na(
-                    sub_xds.load().data, **kwargs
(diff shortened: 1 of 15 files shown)


hydromt/gis/raster.py
```diff
@@ -2290,9 +2290,9 @@ def interpolate_na(
         kwargs.update(dict(method=method, extrapolate=extrapolate))
         if dim0:
             interp_data = np.empty(self._obj.shape, dtype=self._obj.dtype)
-            for i, (_, sub_xds) in enumerate(self._obj.groupby(dim0)):
+            for i, (_, sub_xds) in enumerate(self._obj.groupby(dim0, squeeze=False)):
```
Jaapel, line 2293: What is this doing?

savente93: Squeeze removes empty dimensions. Currently the default is that it sill squeeze, but according to the error, this is being removed, so instead I set squeeze to False (which is the only way to silence the warning) and call squeeze on the result to maintain the same behavior.

Jaapel:
Does the migration guide need to be updated?

savente93:
Not for this one I think since there are no changes necessary on the part of the user. I have updated the changelog, but I don't think the migration guide needs anything

PR #843, opened 2024-03-12
fix mesh_2d_from_rasterdataset for rasterdatasets with multidimension…

### LINKED ISSUE #842: Bug in mesh2d_from_rasterdataset for rotated rasters
### HydroMT version checks

### Reproducible Example

We created a SFINCS quadtree model, resulting in an xugrid. We have a rotated dataset with bed-levels, hence its coordinates are multi-dimensional. The following code should add the data to the mesh.

```
import xarray as xr
import xugrid as xu
from hydromt import workflows
from pyproj import CRS

# open quadtree and set CRS
mesh = xu.open_dataset(r"n:\Deltabox\Postbox\Goede,de Roel\vanTim\quadtree.nc")
mesh.grid.set_crs(CRS.from_wkt(mesh["crs"].crs_wkt))     

# open rotated geotiff
da = xr.open_dataarray(r"n:\Deltabox\Postbox\Goede,de Roel\vanTim\Input_bathy_sfincs_TvE_epsg32611.tif")

# reproject single-dataset to mesh
mesh = workflows.mesh.mesh2d_from_rasterdataset(
    ds = da,
    mesh2d = mesh.grid,
    resampling_method = "barycentric",
)

```

### Current behaviour

The raw data has coordinates xc(y,c) and yc(y,x). By default, the "from_structured" method of xugrid uses "x" and "y" as coordinates. As a result, the reprojection doens't go as expected (x and y start at 0), no overlap is found.

image

### Desired behaviour

Since https://github.com/Deltares/xugrid/pull/202 , xugrid can deal with structured grids that have multi-dimensional coordinates. 

Instead of using the dim_names (they are still y and x) as is done in: https://github.com/Deltares/hydromt/blob/f59f7a59fbd28d3e54567d8e12b061dc36e0f3e0/hydromt/workflows/mesh.py#L264 we should use the coordinate names (in this case "xc" and "yc").

### Additional context

_No response_

roeldegoede:
…al coordinates

## Issue addressed
Fixes #842 

## Explanation
This very simple solution solves the problems on my end, but require the latest version of xugrid.

## Checklist

### DIFF
--- a/hydromt/workflows/mesh.py
+++ b/hydromt/workflows/mesh.py
@@ -1,4 +1,5 @@
 """Implementation for mesh based workflows."""
+
 import logging
 from typing import Dict, List, Optional, Union
 
@@ -261,7 +262,7 @@ def mesh2d_from_rasterdataset(
     # Get one variable name in ds to simplify to da
     var = [v for v in ds.data_vars][0]
     uda = xu.UgridDataArray.from_structured(
-        ds[var].rename({ds.raster.x_dim: "x", ds.raster.y_dim: "y"})
+        ds[var], x=ds.raster.xcoords.name, y=ds.raster.ycoords.name
     )
     uda.ugrid.set_crs(ds.raster.crs)
 


DirkEilander:
Do you know what the minimal required xugrid version is @roeldegoede? Does it only work with the latest or already from earlier versions? 

I'm fine with this solution and pinning a minimal version. Do you agree at @hboisgon?

TODO:
- [ ] update changelog
- [ ] set minimal xugrid version in pyproject.toml and pixi.toml

roeldegoede:
Yes, this support for multi-dimensional coordinates only works for versions of xugrid >= 0.9.0.

Added it to changelog and updates the minimum version.

PR #850, opened 2024-03-25
Table component

### LINKED ISSUE #826: ``ModelTablesComponent`` class
### Kind of request

None

### Enhancement Description

Move ``Model.tables`` to ``ModelTablesComponent``

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #826 

## Explanation
Very straightforward component implementation. I did update the API docs however I didn't update the migration guide (really) or the changelog because I think it's not very instructive to repeat ourselves many times for the very generic components. I made a note in the migration guide so that we won't forget when we restructure it for the alpha release, but I think it's better to do a move general section about components overall than repeat ourselves N times. 

## Checklist

## Additional context
I've targeted `refactor-model` for this PR to make it clearer what the changes will be when that is merged into v1, however I do not intend to merge it into that branch, I'll wait until it is merged into v1 and merge this one to v1 as well.


docs/api.rst
```diff
@@ -515,6 +515,37 @@ Setup methods
    GridComponent.add_data_from_raster_reclass
    GridComponent.add_data_from_geodataframe
 
+TableComponent
+=========
```
hboisgon, line 519:
```suggestion
TableComponent
==============
```

docs/dev/migrating-to-v1.rst
```diff
@@ -333,6 +333,10 @@ has not been changed compared to the GridModel.
 | model.setup_grid_from_*(...) | model.grid_component.add_data_from_*(...) |
 +------------------------------+-------------------------------------------+
 
+TableComponent
+^^^^^^^^^^^^^^
+
+Same as Grid, but it's called Table component and has less methods.
```
hboisgon, line 339: Maybe add that this used to be self.tables.
I thought I made you start a list of added components as well and their meaning (is it here still a dict of pandas.DataFrame for non-geospatial tabular data?)

hydromt/_typing/type_def.py
```diff
@@ -95,3 +98,4 @@ def _validate_bbox(
 GeomBuffer = int
 
 ModeLike = Union[ModelMode, str]
+PandasLike = Union[pd.DataFrame, pd.Series]
```
hboisgon, line 101: I know it avoids repetition but not sure it makes the api definition of the function and accepted argument types clearer to users in the end

hydromt/components/table.py
```diff
@@ -0,0 +1,120 @@
...
+        self._data: Optional[Dict[str, PandasLike]] = None
+        super().__init__(model=model)
+
+    @property
+    def data(self) -> Dict[str, PandasLike]:
+        """Model tables."""
+        if self._data is None:
+            self._initialize_tables()
+        if self._data is None:
+            raise RuntimeError("Could not load data for table component")
```
hboisgon, line 48: Is this check really necessary?

hydromt/components/table.py
```diff
@@ -0,0 +1,120 @@
...
+            return self._data
+
+    def _initialize_tables(self, skip_read=False) -> None:
+        """Initialize the model tables."""
+        if self._data is None:
+            self._data = dict()
+            if self._root.is_reading_mode() and not skip_read:
+                self.read()
+
+    def write(self, fn: str = "tables/{name}.csv", **kwargs) -> None:
```
hboisgon, line 59: add @hydromt_step to all necessary functions

hydromt/components/table.py
```diff
@@ -0,0 +1,120 @@
...
+        self._initialize_tables(skip_read=True)
+        self._model.logger.info("Reading model table files.")
+        fns = glob.glob(join(self._root.path, fn.format(name="*")))
+        if len(fns) > 0:
+            for fn in fns:
+                name = basename(fn).split(".")[0]
+                tbl = pd.read_csv(fn, **kwargs)
+                self.set(tbl, name=name)
+
+    def set(self, tables: Union[PandasLike, Dict], name=None) -> None:
```
hboisgon, line 85:
```suggestion
    def set(self, tables: Union[PandasLike, Dict], name:Optional[str]=None) -> None:
```

docs/dev/migrating-to-v1.rst
hboisgon: On top I wonder if we don't keep the plural here: ``tables: TablesComponent`` to mention that this component actually contains several tables/pd.Dataframe. Maybe we will/can add a singular one ``table: TableComponent`` for single and direct self.data: pd.Dataframe.
Same question for the current ``geoms``.
@DirkEilander what do you think?

hboisgon (changes requested):
Thanks for adding this @savente93 :) still a couple of small things that should be adressed before merging but shouldn't be too much!

hydromt/components/table.py
savente93: not functionally, but it helps the type checkers

hydromt/_typing/type_def.py
savente93: I can see what you mean, in this case it might not be worth it since pandas is a relatively small bit of the code base, so I'll revert this.

docs/dev/migrating-to-v1.rst
savente93: I like the table, I think that's a good idea so I'll add that. I'm not a fan of adding singular versions on top of the plural, but I don't really care much about which is will be, so I'm happy to defer there

savente93: I've added the table so what remains is the plural/singular. @hboisgon @DirkEilander what do you think?

hydromt/components/table.py
deltamarnix: You can also `assert self._data is not None`

deltamarnix: And it is also never true, because `_initialize_tables()` will always set `self._data` to an empty dict I believe

hydromt/components/table.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def _initialize_tables(self, skip_read=False) -> None:
+        """Initialize the model tables."""
+        if self._data is None:
+            self._data = dict()
+            if self._root.is_reading_mode() and not skip_read:
+                self.read()
+
+    @hydromt_step
+    def write(self, fn: str = "tables/{name}.csv", **kwargs) -> None:
```
deltamarnix, line 60: In my PRs I have been trying to make some consts for the file names. I would recommend in the top of this file for a `DEFAULT_FN = "tables/{name}.csv"`

hydromt/components/table.py
```diff
@@ -0,0 +1,126 @@
...
+        """Write tables at <root>/tables."""
+        self._root._assert_write_mode()
+        if self.data:
```
deltamarnix, line 63: `self.data` will always initialize to an empty dict. So it will always write. Not sure if this is the correct check. Perhaps check on `self._data`

hydromt/components/table.py
```diff
@@ -0,0 +1,126 @@
...
+        Parameters
+        ----------
+        tables : pandas.DataFrame, pandas.Series or dict
+            Table(s) to add to model.
+            Multiple tables can be added at once by passing a dict of tables.
+        name : str, optional
+            Name of table, by default None. Required when tables is not a dict.
+        """
+        self._initialize_tables()
+        self._data = cast(Dict[str, Union[pd.DataFrame, pd.Series]], self._data)
```
deltamarnix, line 103: I would opt for `assert self._data is not None`

hydromt/components/table.py
savente93: Indeed though my type checker doesn't seem to pick up on that fact

hydromt/components/table.py
savente93: good point

hboisgon: So there's a reason why we check on ``self.data`` and not ``self._data`` here.
In case of update, ``self.tables`` might never be updated so may never be initialized. When a user specifically wants to write, then it should write without the user having to manually call ``self.tables.read`` or just ``self.tables`` first. This led to a lot of unexpected bugs in the past.
What we use to do but was maybe not well updated in core in Model.write is that in that function we check on _tables to only write components that actually were initialized. See the hydromt_wflow example:

```
    def write(self):
        """Write the complete model schematization and configuration to file."""
        self.logger.info(f"Write model data to {self.root}")
        # if in r, r+ mode, only write updated components
        if not self._write:
            self.logger.warning("Cannot write in read-only mode")
            return
        self.write_data_catalog()
        if self.config:  # try to read default if not yet set
            self.write_config()
        if self._grid:
            self.write_grid()
        if self._tables:
            self.write_tables()
        if self._geoms:
            self.write_geoms()
        if self._forcing:
            self.write_forcing()

```

So please revert back to a check on ``self.data`` but maybe a better check that the dict is not empty.

hydromt/components/table.py
hboisgon: I meant the same as @deltamarnix that the check should not be necessary. Though then I don't understand your answer @savente93. Is there an actual bug or error if you remove that check? Seems a bit worrisome...

hydromt/components/table.py
```diff
@@ -0,0 +1,133 @@
...
+        if self._data is None:
+            self._data = dict()
+            if self._root.is_reading_mode() and not skip_read:
+                self.read()
+
+    @hydromt_step
+    def write(self, fn: str = DEFAULT_TABLE_FILENAME, **kwargs) -> None:
+        """Write tables at <root>/tables."""
+        self._root._assert_write_mode()
+        if len(self.data) > 0:
```
hboisgon, line 62: Nice you implemented it this way :) you can ignore my comments in the other thread then.

hydromt/components/table.py
```diff
@@ -0,0 +1,133 @@
...
+                raise ValueError("name required when tables is not a dict")
+            else:
+                tables_to_add: Dict[str, Union[pd.DataFrame, pd.Series]] = {
+                    name: tables
+                }
+        else:
+            tables_to_add: Dict[str, Union[pd.DataFrame, pd.Series]] = tables
+
+        for df_name, df in tables_to_add.items():
+            if not (isinstance(df, pd.DataFrame) or isinstance(df, pd.Series)):
```
hboisgon, line 116: To be honest, I find all these type checking a tad heavy... And are they really needed, you check higher up on dict and later you check then on pd.DataFrame or pd.Series so this all seems redundant to me. If not I really do need my type checking course ;)

```suggestion
        if not isinstance(tables, dict):
            if name is None:
                raise ValueError("name required when tables is not a dict")
            else:
                tables_to_add = {name: tables}
        else:
            tables_to_add = tables

        for df_name, df in tables_to_add.items():
            if not (isinstance(df, pd.DataFrame) or isinstance(df, pd.Series)):
```

hydromt/components/table.py
```diff
@@ -0,0 +1,133 @@
...
+                raise ValueError(
+                    "table type not recognized, should be pandas DataFrame or Series."
+                )
+            if df_name in self._data:
+                if not self._root.is_writing_mode():
+                    raise IOError(f"Cannot overwrite table {df_name} in read-only mode")
+                elif self._root.is_reading_mode():
+                    self._root.logger.warning(f"Overwriting table: {df_name}")
```
hboisgon, line 124: I think the logger is more a property of model than root no? We have a lot of shortcuts so this is a getting a bit confusing. Maybe we should add again the checks on read and write mode in ``ModelComponent`` then you know most of the time you can just call the ``_`` property to avoid getting these types of confusion.
```suggestion
                    self._logger.warning(f"Overwriting table: {df_name}")
```

hboisgon (commented):
Almost there! Nothing crucial so mainly just a comment here.

hydromt/components/table.py
```diff
@@ -0,0 +1,133 @@
...
+        self._data: Optional[Dict[str, Union[pd.DataFrame, pd.Series]]] = None
+        super().__init__(model=model)
+
+    @property
+    def data(self) -> Dict[str, Union[pd.DataFrame, pd.Series]]:
+        """Model tables."""
+        if self._data is None:
+            self._initialize_tables()
+
+        assert self._data is not None
```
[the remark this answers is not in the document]
savente93: it is not for functionality as this will always be true, it is just to help automated checkers like mypy make deductions it can't otherwise make. I don't really understand why including it is something that you are worried about.

docs/dev/migrating-to-v1.rst
savente93: then we will keep the plural

hydromt/components/table.py
savente93: Some of these checks are not for us but for the type checker. Because python is such a free and dynamic language it can sometimes not make deductions that are easy to see for us. I moved the type definition to the start so we only have to do it once, so hopefully that will make it a bit easier for you.

docs/dev/migrating-to-v1.rst
savente93: Dirk agreed to keep the plural as well

hydromt/components/table.py
```diff
@@ -0,0 +1,133 @@
...
+    Dict,
+    Optional,
+    Union,
+)
+
+import pandas as pd
+
+from hydromt.components.base import ModelComponent
+from hydromt.hydromt_step import hydromt_step
+from hydromt.utils.constants import DEFAULT_TABLE_FILENAME
```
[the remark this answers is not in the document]
savente93: I disagree, I think that making it a class variable actual makes it less clear. Because you cannot give class variables as default arguments, so then you get more hidden functionality. Plugins can still override this by supplying a different default argument of their own. I'm fine with defining the constant in this file if you find that more clear, but I don't think making it a class variable makes sense

docs/api.rst
```diff
@@ -515,6 +515,38 @@ Setup methods
    GridComponent.add_data_from_raster_reclass
    GridComponent.add_data_from_geodataframe
 
+TableComponent
```
hboisgon, line 518: I thought we said to use the plural for this one in the end?
```suggestion
TablesComponent
```

docs/dev/migrating-to-v1.rst
```diff
@@ -233,9 +233,14 @@ Once a component has been added, any component (or other object or scope that ha
...
-In the core of HydroMT, the available components are (list or maybe table):
-  - `GridComponent` for data on a regular grid
-  - etc.
+In the core of HydroMT, the available components are:
++-----------------------+-----------------+-----------------------------------------------------------------+
+| v0.x Model Attribute  | Component       | Description                                                     |
++=======================+=================+=================================================================+
+| model.tables          | TableComponent  | Component for storing non-geospatial data in pandas DataFrames  |
+| model.grid            | GridComponent   | A component for managing regular grids and adding data to it    |
+| model.geoms('region') | RegionComponent | A component for managing the area of interest for the model.    |
```
hboisgon, line 242: For consistency:
```suggestion
| model.tables          | TablesComponent  | Component for managing non-geospatial data in pandas DataFrames  |
| model.grid            | GridComponent   | Component for managing regular gridded data in single hydromt RasterDataset    |
| model.geoms('region') | RegionComponent | Component for managing the area of interest for the model in a geopandas GeoDataFrame.    |
```

docs/dev/migrating-to-v1.rst
```diff
@@ -361,6 +366,10 @@ has not been changed compared to the GridModel.
 | model.setup_grid_from_*(...) | model.grid_component.add_data_from_*(...) |
 +------------------------------+-------------------------------------------+
 
+TableComponent
+^^^^^^^^^^^^^^
+
+Same as Grid, but it's called Table component and has less methods.
```
hboisgon, line 372: The previous ``Model.tables`` is now replaces by a ``TablesComponent`` that can used to store several non-geospatial tabular data into a dictionnary of pandas DataFrames. The TablesComponent for now only contains the basic methods such as ``read``, ``write`` and ``set``.

hydromt/components/table.py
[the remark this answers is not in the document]
hboisgon: Then let's leave it outside of the ModelComponent class. But maybe then define the variable directly in this file instead of in hydromt.utils.constant.
It would be great to have this consistent for all ModelComponents which is currently not the case. May I suggest the following:
```suggestion
DEFAULT_FILENAME = "tables/{name}.csv"
```
@Tjalling-dejong and @deltamarnix also need your opinion on this for your respective PRs.

hydromt/components/table.py
```diff
@@ -0,0 +1,133 @@
...
+
+if TYPE_CHECKING:
+    from hydromt.models.model import Model
+
+__all__ = ["TableComponent"]
+
+
+class TableComponent(ModelComponent):
+    """Table Component."""
+
```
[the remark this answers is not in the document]
hboisgon: Then not this way.

hydromt/components/table.py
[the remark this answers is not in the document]
hboisgon: It became more clear with the mypy discussion we had. What makes me worried is more the burden on the developer to support type checking. I doubt plugins will do all this (mypy, pydantic, type_checking) unless they get more capacity/budget.

hydromt/components/table.py
hboisgon, line 1: Also rename the file to tables.py

hboisgon (approved):
All good now! Just a final rename to use the plural everywhere (TablesComponent instead of TableComponent).
And good if you could all agree on the default filename, @savente93 @Tjalling-dejong and @deltamarnix.

I approve already as the changes left are small but please rename to the plural before merging :)

hydromt/components/table.py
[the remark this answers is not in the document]
savente93: As marnix is still on holiday, I will be resolving this for now, but I've made a ticket for it (#864 ) and we will pick that up when he comes back

PR #853, opened 2024-03-27
Move grid region parsing functionality to region component

### LINKED ISSUE #847: Move region specific code from GridComponent to ModelRegionComponent
### Kind of request

### Enhancement Description

GridComponent in the v1 branch contains quite some code that should actually be part of ModelRegionComponent. The code should be separated more so that it is easier to reuse.

- `region` is still a default component in `Model`. So it will always be present.
- `region.create` will only be called implicitly if a region is passed as an argument in the command line.
- `region.create` is still a valid step in the configuration yml
- `grid.create` has an optional argument `region` to initialize the referenced region. Which is by default `region`, but that can be altered via the `global.components` input in the yml.

### Use case

GridComponent should be able to reference to different region objects. For example, when we have a subgrid with a subregion

```yml
model_type: Model
global:
  components:
    region:
      type: ModelRegionComponent
    subgridregion:
      type: ModelRegionComponent
    grid:
      type: GridComponent
      region: region
    subgrid:
      type: GridComponent
      region: subgridregion
```

The region should be specified when setting up the `GridComponent`, so it can always reference to the right region.

The user should be able to specify the region directly via `region.create`, or via `grid.create`.

```yml
steps:
  - grid.create:
      region:
        bbox: [1,2,3,4]
# OR
  - region.create:
      region:
        bbox: [1,2,3,4]
```

### Additional Context

Setting up these different components is something that could also be done within a model class when making a plugin like so:

```py
class MyPluginModel(Model):
  def __init__(self):
    super().__init__()
    self.add_component("grid", GridComponent(self, region="region"))
    self.add_component("subgridregion", ModelRegionComponent(self))
    self.add_component("subgrid", GridComponent(self, region="subgridregion"))
```

`GridComponent` will keep a `str` to look up the component when required to prevent the garbage collector from cleaning up the code when needed.

```py
class GridComponent(ModelComponent):
  def __init__(self, region: str):
    self.region_component = region

  def create(self, ...):
    region_component = self._model.get_component(self.region_component, ModelRegionComponent)
```

deltamarnix:
## Issue addressed
Fixes #847 

## Explanation
Move region code parsing from GridComponent to ModelRegionComponent.
Attaches the region to the grid via the name id of the region component in the constructor of grid.
Removes region from the command line arguments.

## Checklist

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -424,13 +424,13 @@ Attributes
    ModelComponent.logger
    ModelComponent.model_root
 
-ModelRegionComponent
+RegionComponent
 ====================
 
 .. autosummary::
    :toctree: _generated
 
-   ModelRegionComponent
+   RegionComponent
 
 Attributes
 ----------
@@ -438,13 +438,13 @@ Attributes
 .. autosummary::
    :toctree: _generated
 
-   ModelRegionComponent.data
-   ModelRegionComponent.model
-   ModelRegionComponent.crs
-   ModelRegionComponent.bounds
-   ModelRegionComponent.data_catalog
-   ModelRegionComponent.logger
-   ModelRegionComponent.model_root
+   RegionComponent.data
+   RegionComponent.model
+   RegionComponent.crs
+   RegionComponent.bounds
+   RegionComponent.data_catalog
+   RegionComponent.logger
+   RegionComponent.model_root
 
 General Methods
 ---------------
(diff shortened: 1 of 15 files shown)


deltamarnix:
TODO: 

- [ ] Fix the open TODOs in region.py
- [ ] Add functionality from `MeshComponent` to `RegionComponent` once `MeshComponent` PR #792 is finished.

PR #854, opened 2024-03-27
Vector component

### LINKED ISSUE #818: `VectorModelComponent` class
### Kind of request

None

### Enhancement Description

Currently the vector model class does not have its corresponding component yet. this should be added. 

### Use case

_No response_

### Additional Context

_No response_

deltamarnix:
## Issue addressed
Fixes #818 

## Explanation
Changes the VectorModel and VectorMixin to a VectorComponent.

## Checklist


hydromt/test_utils/test_equal.py
```diff
@@ -0,0 +1,69 @@
...
+    for c in components:
+        component_equal, component_errors = check_components_equal(
+            getattr(one, c), getattr(other, c), c
+        )
+        is_equal &= component_equal
+        errors.update(**component_errors)
+    return is_equal, errors
+
+
+def check_components_equal(
```
deltamarnix, line 45: We will have to think this through later. How can we ensure that other components created by plugins are also testable.

hydromt/components/vector.py
```diff
@@ -162,20 +169,25 @@ def read_vector(
...
+            )
             # check if ds is empty (default fn has a value)
             if len(ds.sizes) == 0:
                 fn = None
-        if fn_geom is not None and isfile(join(self.root.path, fn_geom)):
-            gdf = gpd.read_file(join(self.root.path, fn_geom))
+        if fn_geom is not None and isfile(join(self._root.path, fn_geom)):
+            gdf = gpd.read_file(join(self._root.path, fn_geom))
             # geom + netcdf data
+            # TODO: What if ds is None?
```
hboisgon, line 190: Isn't it taken into account ? lines 185-186 makes sure ds is not empty and I assume read_nc would complain if file not found. Extra I think that the default for data_vars is None anyway.

hydromt/components/vector.py
```diff
@@ -250,164 +263,63 @@ def write_vector(
...
-        gdf_other = geods_other.vector.geometry.to_frame("geometry")
-        errors.update(_check_equal(gdf, gdf_other, name="vector.geometry"))
-        # test vector data only
-        if geods.vector.geom_format == "xy":
-            drop_vars = [geods.vector.x_name, geods.vector.y_name]
-        else:
-            drop_vars = [geods.vector.geom_name]
-        ds = geods.drop_vars(drop_vars)
-        ds_other = geods_other.drop_vars(drop_vars)
-        errors.update(_check_equal(ds, ds_other, name="vector.data"))
```
hboisgon, line 411: This part was not transferred. Maybe each component should have its own test_equal function that can be called from Model.test_equal ? That decision would impact other components @Tjalling-dejong and @savente93

hydromt/models/model.py
```diff
@@ -1466,39 +1464,6 @@ def crs(self) -> CRS:
         """Returns coordinate reference system embedded in region."""
         return self.region.crs
 
-    # test methods
-    def _test_equal(self, other, skip_component=None) -> Tuple[bool, Dict]:
-        """Test if two models including their data components are equal.
```
hboisgon, line 1471: This method is very useful so I would like to see it transfered. Maybe here just for each component create their own ``test_equal`` method and here like ``Model.write`` just loop through the components.

hydromt/test_utils/test_equal.py
hboisgon: I suggest adding a test_equal method do each component and then keep the test_equal method in Model to loop over components.

docs/api.rst
```diff
@@ -533,30 +531,27 @@ Components and attributes
...
 -------------
 
 .. autosummary::
    :toctree: _generated
 
-   VectorModel.setup_config
-   VectorModel.setup_region
-   VectorModel.setup_maps_from_rasterdataset
-   VectorModel.setup_maps_from_raster_reclass
+   VectorModel.create
```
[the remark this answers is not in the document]
hboisgon: Update: indeed I does not exist.

hydromt/components/vector.py
```diff
@@ -97,43 +103,44 @@ def set_vector(
...
-                    self._vector[dvar] = data[dvar]
+                    self.data[dvar] = data[dvar]
                 else:
                     raise ValueError(
                         f"Index coordinate of data variable {dvar} "
                         "does not match vector index coordinate"
                     )
 
-    def read_vector(
+    def read(
```
hboisgon, line 139: I forgot in my review but please add @hydromt_step to read and write functions

hydromt/components/vector.py
savente93: So this depends on how we want to define equality. if it is just "has the same components with the right name and type" then model can just do this on it's own. if equality means "has the same data" then yet, each component will need to implement their own equality check. However, if we want this then this equality check will have to become a required part of the component API. is it important enough to us to for plugins to do this? (genuine question, no sarcasm intended). I am okay with this, but we should be aware and document properly that this could be a very expensive operation depending on how much data is involved.

deltamarnix: I was also wondering what this is used for. And it's mainly for unit tests to see if reading and writing is implemented correctly. For that reason I would opt for moving it to the test utilities, which I did (see test_equal.py)

hydromt/test_utils/test_equal.py
deltamarnix: Discussed with the team. We will make the `test_equal` functionality a separate module for testing utils.

hydromt/components/vector.py
```diff
@@ -1,49 +1,55 @@
...
+from hydromt.io.readers import read_nc
+from hydromt.io.writers import write_nc
+from hydromt.models.model import Model
 
-__all__ = ["VectorModel"]
-logger = logging.getLogger(__name__)
+__all__ = ["VectorComponent"]
 
+DEFAULT_FN = "vector/vector.nc"
+DEFAULT_FN_GEOM = "vector/vector.geojson"
```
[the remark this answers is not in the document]
deltamarnix: Discussed with the team. Defaults go into the constructor, `write` and `read` functions will get optional arguments.

hydromt/components/vector.py
```diff
@@ -232,11 +249,11 @@ def write_vector(
...
+        ds = self.data
         if len(ds) == 0:
-            self.logger.debug("No vector data found, skip writing.")
+            self._logger.debug("No vector data found, skip writing.")
             return
-        self._assert_write_mode()
+        self._root._assert_write_mode()
 
         # If fn is None check if vector contains only 1D data
         if fn is None:
```
deltamarnix, line 259: @hboisgon I was looking through this function and I figure that `fn` will never ever be `None`, not even in the old version of this function. I am wondering if there is a bunch of dead code here.

hydromt/components/vector.py
deltamarnix: I've put this code back where it belongs.

hydromt/components/vector.py
deltamarnix: Fixed by making fn optional

docs/api.rst
```diff
@@ -608,30 +607,24 @@ Components and attributes
 .. autosummary::
    :toctree: _generated
 
-   VectorModel.vector
-   VectorModel.crs
-   VectorModel.region
+   VectorComponent.data
+   VectorComponent.geometry
+   VectorComponent.index_dim
+   VectorComponent._model
```
hboisgon, line 613:
```suggestion
   VectorComponent.index_dim
   VectorComponent.crs
   VectorComponent._model
```

hydromt/models/model.py
```diff
@@ -455,6 +453,38 @@ def write_data_catalog(
 
             cat.to_yml(path, root=root)
 
+    def test_equal(self, other: "Model") -> tuple[bool, dict[str, str]]:
```
hboisgon, line 456: I would add a skip_components argument here like before.

tests/models/test_model.py
```diff
@@ -549,90 +547,109 @@ def test_gridmodel_setup(tmpdir):
...
+    # automatic split
+    vector_component.write(
+        filename=None, geometry_filename="vector/vector_split.geojson"
+    )
+    assert isfile(join(vector_model_no_defaults.root.path, "vector", "vector_split.nc"))
+    assert not isfile(
+        join(vector_model_no_defaults.root.path, "vector", "vector_all.nc")
+    )
+    # geojson 1D data only
+    vector_component._vector = vector_component._vector.drop_vars("zs").drop_vars(
```
hboisgon, line 620:
```suggestion
    vector_component._data = vector_component._data.drop_vars("zs").drop_vars(
```

tests/models/test_model.py
```diff
@@ -549,90 +547,109 @@ def test_gridmodel_setup(tmpdir):
...
+    vector_model1.vector._vector = None
+    # geojson only
+    # automatic split
+    vector_model1.vector.read(
+        filename="vector/vector_split.nc",
+        geometry_filename="vector/vector_split.geojson",
+    )
+    vector1 = vector_model1.vector.data
+    assert len(vector1["zs"].dims) == 2
+    vector_model1.vector._vector = None
```
hboisgon, line 646:
```suggestion
    vector_model1.vector._data= None
```

tests/models/test_model.py
```diff
@@ -549,90 +547,109 @@ def test_gridmodel_setup(tmpdir):
...
+    )
+
+    # test read vector
+    vector_model1 = Model(root=str(tmpdir), mode="r")
+    vector_model1.add_component("vector", VectorComponent(vector_model1))
+    # netcdf only
+    vector_model1.vector.read(filename="vector/vector_full.nc", geometry_filename=None)
+    vector0 = vector_model1.vector.data
+    assert len(vector0["zs"].dims) == 2
+    vector_model1.vector._vector = None
```
hboisgon, line 637:
```suggestion
    vector_model1.vector._data = None
```

hboisgon (changes requested):
Thanks for re-adding the test equal functionality @deltamarnix !
I think still a few minor comments before it's ready to be merged.

hydromt/components/tables.py
```diff
@@ -141,3 +142,28 @@ def get_tables_merged(self) -> pd.DataFrame:
         return pd.concat(
             [df.assign(table_origin=name) for name, df in self.data.items()], axis=0
         )
+
+    def test_equal(self, other: ModelComponent) -> tuple[bool, dict[str, str]]:
```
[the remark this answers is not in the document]
deltamarnix: The parent class defines the definition for this function. It is not possible to just change the type of the parameter. mypy doesn't like that. Also, the type checking is done on runtime in the parent class anyways. So we're good.

hydromt/components/vector.py
```diff
@@ -97,109 +125,127 @@ def set_vector(
...
+                if dvar in self.data:
+                    self._logger.warning(f"Replacing vector variable: {dvar}")
                 # check on index coordinate before merging
                 dims = data[dvar].dims
                 if np.array_equal(
-                    data[dims[0]].values, self._vector[self.index_dim].values
+                    data[dims[0]].values, self.data[self.index_dim].values
                 ):
-                    self._vector[dvar] = data[dvar]
+                    self.data[dvar] = data[dvar]
```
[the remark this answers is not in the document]
deltamarnix: `self.data` is a `Dataset` object. When you ask for `data`, you retrieve a reference of the dataset, not a copy. So in this case `self._data` and `self.data` are the same, right? The only difference for the type checker is that we are certain that `self.data` is not `None`, so it is easier to work with.

I can also write this, if you prefer:

```py
assert self._data is not None
self._data[dvar] = data[dvar]
```

hydromt/models/model.py
deltamarnix: Let's add it once we need it. I haven't found any code that uses it.

deltamarnix:
@savente93 Could you write a `test_equal` function for ConfigComponent?

hydromt/components/vector.py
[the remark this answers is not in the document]
hboisgon: I prefer this yes. You can add the assertion right after the call to self._initiliaze

hydromt/models/model.py
hboisgon: Plugins ? :)
But in a way now by default you check on the component type which was not there before so maybe implementing a skip argument is less needed than it used to be.

hboisgon (approved):
All good now!
Please still update the ``set`` function to update the ``_data`` object but after that you can merge so I approve already :)

hydromt/components/vector.py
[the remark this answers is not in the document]
deltamarnix: I will change it, no problem. But apart from preference, is there a functional difference that I miss?

hydromt/models/model.py
savente93: Perhaps a nice middle way is to only check equality on components that are loaded? Just an idea, if it doesn't work, feel free to ignore.

tests/models/test_model.py
```diff
@@ -144,7 +146,7 @@ def test_model(model, tmpdir):
     # check if equal
     model._results = {}  # reset results for comparison
     with pytest.deprecated_call():
```
savente93, line 148: I think this should no longer be deprecated now as we decided to keep it?

tests/models/test_model.py
```diff
@@ -549,90 +547,107 @@ def test_gridmodel_setup(tmpdir):
...
+    model1.read()
+    equal, errors = vector_model.test_equal(model1)
+    assert equal, errors
+
+
+def test_vectormodel_vector(vector_model_no_defaults, tmpdir, geoda):
+    # test set vector
+    testds = vector_model_no_defaults.vector.data.copy()
+    vector_component = vector_model_no_defaults.get_component("vector", VectorComponent)
+    # np.ndarray
```
savente93, line 572: I think it would be nice to subdivide these into multiple tests, but you can also leave that for later if you want, as it's not critical and this PR is already quite large.

savente93 (approved):
Very nice Marnix. Good to see we were renable some tests again! I had a few suggestions, but nothing important. It's all good to go IMO.

hydromt/models/model.py
deltamarnix: There is no real way to check if a component is loaded, as it is lazily loaded. Whenever `data` is called, it will be loaded.

tests/models/test_model.py
deltamarnix: not now :)

PR #863, opened 2024-03-29
Add `ConfigComponent`

### LINKED ISSUE #799: `ConfigModelComponent` should be able to write simulation config files in `toml` format
### Kind of request

### Enhancement Description

currently toml is supported as a format for both reading and writing of all config files except the data catalog. as described in https://github.com/Deltares/hydromt/pull/791#discussion_r1499481644 there should be toml support but only for writing the simulation config files, and not for any reading. 

### Use case

_No response_

### Additional Context

_No response_

### LINKED ISSUE #823: ``ModelConfigComponent`` class
### Kind of request

None

### Enhancement Description

Move ``Model.config`` to a ``ModelConfigComponent``

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #823 
Fixes #799 

## Explanation
The implementation is almost identical to the original. The only thing that really changed was the name, which I put my reasoning for in #862 

## Checklist


hydromt/models/model.py
```diff
@@ -99,9 +98,9 @@ def __init__(
             Model root, by default None
         mode : {'r','r+','w'}, optional
             read/append/write mode, by default "w"
-        config_fn : str, optional
-            Model simulation configuration file, by default None.
-            Note that this is not the HydroMT model setup configuration file!
+        kernel_config_path : str, optional
+            Path to where the configuration file for the kernel/simulation
+            that is going to consume the model should be written.
```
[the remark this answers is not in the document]
hboisgon: And you are actually not even using this argument anymore, are you?

tests/components/test_kernel_config_component.py
```diff
@@ -15,7 +19,6 @@ def test_config_dict():
     return {
         "section1": {
             "list": [1, 2, 3],
-            # "tuple": (1, "b"), # yaml cannot deal with tuple
```
hboisgon, line 18: But did it use to then? Is that something we should keep? 
For Model ConfigComponent, I don't know a lot of models that use yaml but for the hydromt build/update config, this means that none of the @hydromt_step functions can have tuple in their arguments (also valid for plugins).

hboisgon (changes requested):
Good start @savente93 but we are not there yet.
First on the name of this component. I think ConfigComponent is good enough as hydromt config is not a component of model.

Then on the functionality:

- one function was not transferred: ``Model.setup_config``
- the behavior of ``Model.config`` is not properly transferred

I elaborated a bit more in the comments but config was a particular object in Model that was always created on the fly from a template instead of the user specifically creating it. I think if we want to achieve full flexibility, we can simplify the behavior and harmonize with other components. My suggestion would be:

- remove ``config_fn`` from ``Model.__init__``
- keep allowing to create a config using a template file. This can be achieved with either (1) add the config filename to ``ConfigComponent.__init__`` and transfer the rest of the behavior as in main. Or (2) add a ``create`` function that allows to start a config from a template config file (more in line with other components but a bigger change for the user).

I think I would prefer option 2. @DirkEilander any thoughts on this?

savente93:
@hboisgon I see. I'm happy to update the functionality I'll get to work on that, but the name I feel a bit more strongly about. I don't really care what the affix will be (kernel/simulation/output/etc.) but I do feel quite strongly about it not being just `ConfigComponent` as this has caused me quite some confusion even while writing this PR. I'm happy to see if we can find a middle ground, but `ConfigComponent` is too generic in my opinion.

hydromt/models/model.py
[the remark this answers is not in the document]
savente93: good point, I'll remove it

hydromt/components/kernel_config.py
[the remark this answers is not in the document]
savente93: `set` has become this no? Well okay, I admit that `set` does not take a dictionary, I'll add that one, but aside from that all the functionality is there since `Model.setup_config` simply set the value for each key value pair in the provided dictionary. As for naming, `update` is doable, tough it isn't necessarily an update since it can also create new ones. how about `set` for the dictionary one and `set_value` for the single key-value pair?

hydromt/components/kernel_config.py
```diff
@@ -0,0 +1,177 @@
...
+
+from hydromt.components.base import ModelComponent
+from hydromt.io.path import make_config_paths_relative
+from hydromt.io.readers import read_yaml
+from hydromt.io.writers import write_toml, write_yaml
+
+if TYPE_CHECKING:
+    from hydromt.models import Model
+
+DEFAULT_KERNEL_CONFIG_PATH = "kernel_config.yml"
```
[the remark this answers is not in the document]
savente93: good point, that is my mistake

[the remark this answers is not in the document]
savente93: see #864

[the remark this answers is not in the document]
savente93: As you correctly noted, it was no longer necessary to take this as an argument in the init, so I've just removed that and resolved this conversation, as I think that resolved the issue, but feel free to let me know if I've misunderstood.

hydromt/components/kernel_config.py
```diff
@@ -0,0 +1,177 @@
...
+            raise RuntimeError("Could not load data for kernel config component")
+        else:
+            return self._data
+
+    def _initialize_kernel_config(self, skip_read=False) -> None:
+        """Initialize the kernel configs."""
+        if self._data is None:
+            self._data = dict()
+            if self._root.is_reading_mode() and not skip_read:
+                self.read()
```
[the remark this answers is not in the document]
savente93: ah I see, I understand better now. I'm honestly happy with whatever you think is best. At the very least I'll add the functionality to read from a template as that will be necessary anyway.

tests/components/test_kernel_config_component.py
savente93: I don't know, this comment was added in a PR before I joined (#292 ) What I could do is cast tuples to lists, when writing to file as these are functionally identical in python, although this would require a lot more work as then I'd have to traverse the whole dictionary. Trying to write tuples to yaml produces a valid but weird syntax and trying to read tuples leads it to just becoming a string: 
```python
In [10]: yaml.safe_load(r"""a:
    ...:     (1,2,3)""")
Out[10]: {'a': '(1,2,3)'}

In [11]: yaml.dump({'a':(1,2,3)})
Out[11]: 'a: !!python/tuple\n- 1\n- 2\n- 3\n'
```

Users can still set tuples in their python code through the `set` function, but given that it's not a feature of yaml I don't think this should be our concern.

hydromt/components/kernel_config.py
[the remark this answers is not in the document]
savente93: I've added the functions, all that remains is to see if you like the names.

docs/api.rst
```diff
@@ -516,6 +516,50 @@ Setup methods
...
+
+
+
+Setup methods
+-------------
+
+.. autosummary::
+   :toctree: _generated
+
+   KernelConfigComponent.create
```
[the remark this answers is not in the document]
savente93: resolved in other comment.

hydromt/components/kernel_config.py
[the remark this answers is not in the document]
savente93: I've added the reading regardless of mode and written some tests for it. From these comments I also deduced that a model should have a kernel config component by default so I added it during init, please let me know if I have misunderstood. I think this also is a bit closer now to what you wanted re it being overridable for plugins since they can now just change the default string that gets added in the model init either by subclassing the component or by just providing a different string in their model init. Is this what closer to what you wanted?

tests/components/test_kernel_config_component.py
hboisgon: If this was already not supported before then we can ignore :) Just wanted to make we were not removing existing functionality with the transfer to components

hydromt/components/kernel_config.py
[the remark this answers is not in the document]
hboisgon: I would like to have Dirk advice on how to deal with config before merging.
I still don't like the ``DEFAULT_KERNEL_CONFIG_PATH``. I think most models actually don't need to subclass ``ConfigComponent`` as functionality is pretty standard and we support reading and writing from TOML. But if I'm wflow, then I still need to subclass to only change the ``DEFAULT_KERNEL_CONFIG_PATH``.... That users can overwrite the template is good but plugins should be able to overwrite the real default value as it does not make sense for them to use the hydromt one.

hydromt/components/kernel_config.py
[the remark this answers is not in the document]
hboisgon: So convention wise for the names:

- ``GridComponent.set`` is equivalent to ``KernelConfigComponent.set``. It is used to read if needed, validate data to add and actually add to ``self._data``
- ``GridComponent.add_data_from_rasterdataset`` is equivalent to ``KernelConfigComponent.setup_config`` in the terms that this is a hydromt_step that user can call from hydromt config to modify the config. I mentioned ``update`` because ``set`` is the method that creates new ones (by calling read or starting an empty dict) while ``setup_config`` is not there to set up from scratch but just to change values in config.
- ``KernelConfigComponent.get_value`` we do because it's more complex than just ``GridComponent.data[''variable"]``

So could you change back ``set_value`` to ``set`` to respect conventions with other components? And then add a hydromt_step function to update config from a dictionary? I'm never good with ones so this one is up to you.

hydromt/components/kernel_config.py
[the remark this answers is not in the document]
savente93: This is true for the code that this comment is attached to, that I understand. However, in the newest version I changed this, so that the component actually indeed stores the path that can be given at model init, meaning that all a plugin has to do is changed what they pass into the config component. Hopefully this is better, but I'm okay with waiting for what @DirkEilander thinks.

hydromt/components/kernel_config.py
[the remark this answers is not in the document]
savente93: I think `set` and `update` are a good one, so I went with that.

docs/api.rst
```diff
@@ -516,6 +516,50 @@ Setup methods
    GridComponent.add_data_from_geodataframe
 
 
+KernelConfigComponent
+=========
```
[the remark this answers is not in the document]
DirkEilander: I think `ConfigComponent` is most consistent with the other names. Basically all `ModelComponent` classes are model or kernel components. I think renaming the HydroMT config to HydroMT workflow would be a better way to avoid confusion with the `ConfigComponent`

hydromt/components/kernel_config.py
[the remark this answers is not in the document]
DirkEilander: Regarding `DEFAULT_KERNEL_CONFIG_PATH`. From the code I don't understand how the functionality you describes actually works @savente93. Could you give a short code example of how this would look like for a plugin? We used to this this with class properties as mentioned by @hboisgon. I don't really care how, but it would be good if plugins don't need to subclass to only change one class property.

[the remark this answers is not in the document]
DirkEilander: > So even in write/build mode, we would read from a template config file generally stored in the plugin repositories. Hence even in write mode, there is always a config to read so we did not add a check on self._root.is_reading_mode() only on skip_read. This is also why the config_fn filename was the only filename argument in Model.__init__.

> We could decide to make this behavior more transparent by adding a create method to ConfigComponent which creates a config from a template. Later this config can be updated by the not yet transfered setup_config method. It's a little bit of a breaking change for users but I think in a way better for transparency. @DirkEilander what do you think?

 I agree with @hboisgon that a `create` method makes sense and is consistent with other model components.

hydromt/components/config.py
```diff
@@ -0,0 +1,215 @@
...
+if TYPE_CHECKING:
+    from hydromt.models import Model
+
+DEFAULT_CONFIG_PATH = "config.yml"
+
+
+class ConfigComponent(ModelComponent):
+    """A component to write configuration files for simulations/kernels."""
+
+    def __init__(self, model: "Model", config_fn: Optional[str] = None):
```
deltamarnix, line 23:
```suggestion
    def __init__(self, model: "Model", *, config_fn: Optional[str] = None):
```

hydromt/components/config.py
```diff
@@ -0,0 +1,215 @@
...
+    ) -> None:
+        """Write kernel config at <root>/{path}."""
+        self._root._assert_write_mode()
+        if self.data:
+            if path is not None:
+                p = join(self._root.path, path)
+            else:
+                p = join(self._root.path, self._config_fn)
+
+            write_path = join(self._root.path, p)
```
deltamarnix, line 69: This looks a bit weird. We just did a join with `_root.path`, and now we do it again? That can never be right.

hydromt/components/config.py
```diff
@@ -0,0 +1,215 @@
...
+
+    @hydromt_step
+    def read(self, path: Optional[str] = None) -> None:
+        """Read kernel config at <root>/{path}."""
+        self._initialize(skip_read=True)
+        # if path is abs, join will just return path
+        if path is not None:
+            p = path
+        else:
+            p = self._config_fn
```
deltamarnix, line 93:
```suggestion
        p = path pr self._config_fn
```

hydromt/components/config.py
```diff
@@ -0,0 +1,215 @@
...
+            elif ext == ".toml":
+                write_toml(write_path, write_data)
+            else:
+                raise ValueError(f"Unknown file extention: {ext}")
+
+        else:
+            self._model.logger.debug("No kernel config found, skip writing.")
+
+    @hydromt_step
+    def read(self, path: Optional[str] = None) -> None:
```
deltamarnix, line 86: Instead of optional, I would just put DEFAULT_PATH here, like I did in my PR. Then you can get rid of config_fn in the constructor. Unless that's the way we want to do it of course.

```suggestion
    def read(self, path: str = DEFAULT_PATH) -> None:
```

hydromt/components/config.py
```diff
@@ -0,0 +1,215 @@
...
+        if isfile(read_path):
+            self._model.logger.info(f"Reading kernel config file from {read_path}.")
+        else:
+            self._model.logger.warning(
+                f"No default kernel config was found at {read_path}. It wil be initialized as empty dictionary"
+            )
+            return
+
+        ext = splitext(p)[-1]
+        if ext in [".yml", ".yaml"]:
```
deltamarnix, line 104: So we can write to toml, but not read from toml?

hydromt/components/config.py
```diff
@@ -0,0 +1,215 @@
...
+        --------
+        >> self.set({'a': 1, 'b': {'c': {'d': 2}}})
+        >> self.data
+            {'a': 1, 'b': {'c': {'d': 2}}}
+        >> self.set({'a.d.f.g': 1, 'b': {'c': {'d': 2}}})
+        >> self.data
+            {'a': {'d':{'f':{'g': 1}}}, 'b': {'c': {'d': 2}}}
+        """
+        for k, v in data.items():
+            self.update(k, v)
```
deltamarnix, line 129: Sounds like a good place for pydantic.v1.utils.deep_update? Or the recursive version that you wrote yourself in another PR.

hydromt/components/config.py
```diff
@@ -0,0 +1,215 @@
...
+        current = self.data
+        for i, part in enumerate(parts):
+            if part not in current:
+                current[part] = {}
+            if i < num_parts - 1:
+                current = current[part]
+            else:
+                current[part] = value
+
+    def get_value(self, key: str, abs_path: bool = False):
```
deltamarnix, line 166:
```suggestion
    def get_value(self, key: str, abs_path: bool = False) -> Any:
```

hydromt/components/config.py
```diff
@@ -0,0 +1,215 @@
...
+        >> 1
+
+        >> get_config_value('b.c.d')
+        >> 2
+
+        >> get_config_value('b.c')
+        >> {'d': 2}
+
+        """
+        self._initialize()
```
deltamarnix, line 198: No need to initialize, `self.data` does that.

hydromt/components/config.py
savente93: good point, this stemmed from a missunderstanding on my part but it should be fixed now

hydromt/components/config.py
savente93: good point

hydromt/components/config.py
savente93: to be discussed tomorrow

hydromt/components/config.py
savente93: you're correct, that was a copy paste error on my part.

hydromt/components/config.py
deltamarnix: There is this new function `rgetattr` that I built that does exactly that expansion. That might help, right?

savente93: it would for the `get` that's true, but for the update the dictionaries need to be created which it doesn't do I believe

hboisgon:
Hi @savente93 I added the changes to have config behave like other component and a method to create from template. I did not have time yet to create tests for it. Could you look at the implementation I made and finalize from there?

hydromt/components/config.py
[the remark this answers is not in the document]
savente93: Resolving as we discussed this during the refinement today

hydromt/components/config.py
savente93: Resolving as we discussed this during the refinement today

savente93:
@hboisgon I think it should all be good now. I added the note about our discussion during our last refinement in the migration guide instead of the plugin dev because that is still referencing the state on main so I didn't think there was a good place to put it there without being very out of place or rewriting the entire page, hope this is okay to you. If not please let me know where you think it should go and I'll put it there. Other than that, I think we're all in agreement now. If so, can you give the final approve?

hydromt/components/config.py
```diff
@@ -70,7 +71,7 @@ def _initialize(self, skip_read=False) -> None:
         """Initialize the model config."""
         if self._data is None:
             self._data = dict()
-            if self._root.is_reading_mode() and not skip_read:
+            if not skip_read:
```
hboisgon, line 74: Can you re-add the sefl._root.is_reading_mode? I added it again because we are changing behavior and enforcing a create method in write mode for config.

hydromt/components/config.py
```diff
@@ -102,10 +103,9 @@ def write(
     @hydromt_step
     def read(self, path: Optional[str] = None) -> None:
         """Read model config at <root>/{path}."""
-        self._root._assert_read_mode()
         self._initialize(skip_read=True)
         # if path is abs, join will just return path
-        p = path or self._filename
+        p = path or self._default_template_filename or self._filename
         read_path = join(self._root.path, p)
```
hboisgon, line 109: Can you rechange? We decided with Dirk to add a create method rather than keep reading a template by default for config in write mode for more transparency

hydromt/components/config.py
```diff
@@ -236,7 +234,7 @@ def create(
             {'a': 1, 'b': {'c': {'d': 2}}}
         """
         # Check if self.data is not empty
-        if len(self.data) == 0:
+        if self._data is not None and len(self._data) > 0:
```
hboisgon, line 237: Good catch! But you should do the check on ``self.data`` an not ``self._data`` to force read if you have read rights here.

hydromt/components/config.py
```diff
@@ -250,12 +248,23 @@ def create(
             template = self._default_template_filename
             prefix = "default"
         else:
-            self._logger.warning("No template provided and no default template found.")
+            self._logger.warning(
+                f"No default model config was found at {template}. "
+                "It wil be initialized as empty dictionary"
+            )
+            self._data = {}
```
hboisgon, line 255: This is not needed as self._data should already be an empty dict at that stage

hydromt/components/config.py
```diff
@@ -250,12 +248,23 @@ def create(
...
+            return
+
+        if not isfile(template):
+            self._logger.warning(
+                f"Template was provided but file did not exist: {template}"
+                "It wil be initialized as empty dictionary"
+            )
+            self._data = {}
```
hboisgon, line 263: same as above

hydromt/components/config.py
```diff
@@ -250,12 +248,23 @@ def create(
...
             return
 
+        template = Path(template)
         # Here directly overwrite config with template
-        if not template.is_file():
-            raise FileNotFoundError(f"Template file not found: {template}")
```
hboisgon: I prefer to raise an error if the user provided a file and it's not found rather than a warning. Unless we want to add the NoData policy thing here with error by default

hydromt/models/model.py
```diff
@@ -135,7 +137,11 @@ def __init__(
         self._components: Dict[str, ModelComponent] = {}
         self._add_components(components)
 
-        config_component: ConfigComponent = ConfigComponent(self, config_fn=config_fn)
+        config_component: ConfigComponent = ConfigComponent(
+            self,
+            filename=config_filename,
+            default_template_filename=default_config_template_filename,
+        )
```
hboisgon, line 144: I don't know if we want to add a config component by default to Model, especially some model might decide to create a DelwaqConfigComponent just because they have to subclass as the format is not toml or yaml.

hboisgon (changes requested):
Thanks for updating and adding tests @savente93 !
Because we have a ``create`` function now, read for config can work similar to other components ie we dont read in write only mode.
Also I don't think we should add ConfigComponent by default in Model as the ConfigComponent will not suit every plugins. Also the filename can be changed through the components arguments by the user so they dont need to be arguments of model init

hydromt/models/model.py
savente93: sure I will remove it

savente93:
@hboisgon should be all good now!

savente93:
oh wow good catch! thanks for that, you're right!

PR #866, opened 2024-04-02
plugins for drivers

### LINKED ISSUE #432: Support custom drivers to connect to data APIs

Jaapel:
## Issue addressed
Fixes #432 

## Explanation
Added plugin behaviour for drivers. To load drivers, you only need to import them, as the validators search for subclasses of the Driver Component. To keep the plugin consistent with others, I did keep the `PLUGINS` API similar to the other plugins, although it is not used in the `BaseDriver` class.

The main advantage of the current `BaseDriver` class is that you could define a `Driver` inline, and still use `BaseDriver.model_validate`, or `create_source({"driver": {"name": "my_plugin_driver"}})` without having to do an editable install.

## Checklist


hydromt/plugins.py
```diff
@@ -71,6 +75,25 @@ def component_plugins(self) -> dict[str, type["ModelComponent"]]:
...
+            raise RuntimeError("Could not load any driver plugins")
+
+        drivers: dict[str, Type["BaseDriver"]] = cast(
+            Dict[str, Type["BaseDriver"]],
+            {name: value["type"] for name, value in self._driver_plugins.items()},
+        )
+        # Do not return ABCs, such as BaseDriver, or RasterDatasetDriver
+        return {
+            key: value for key, value in drivers.items() if ABC not in value.__bases__
+        }
```
deltamarnix, line 95: https://docs.python.org/3/library/inspect.html#inspect.isabstract

deltamarnix, line 95:
```suggestion
        return {name: value["type"] for name, value in self._driver_plugins.items() if not inspect.isabstract(value["type"])}
```

hydromt/drivers/__init__.py
deltamarnix, line 1: I see that only the drivers are in this list of `__all__`, but there are some other python files that are now not exposed. When someone wants to create its own driver, you might want to open up preprocessor.py.

And also `BaseDriver` is now part of the list of drivers, although it is an abstract class.

Jaapel: How do we see `__all__`? Just for the entrypoints, or just as a way to shorten import paths?

Jaapel:
@deltamarnix @savente93 Added: https://github.com/Deltares/hydromt/issues/872

hydromt/_typing/type_def.py
```diff
@@ -59,7 +56,7 @@ def _validate_bbox(
 RasterDatasetSource = StrPath
 DatasetSource = StrPath
 
-Bbox = Tuple[float, float, float, float]
+# Bbox = Tuple[float, float, float, float]
```
savente93, line 59: if it's no longer necessary I'd remove it.

hydromt/drivers/preprocessing.py
savente93, line 1: Sonar cube is flagging that you coppied some code here from data dapter without removing it from the original place. Can you remove it in either place? Same for Pygrio_driver and gis_utils

hydromt/metadata_resolver/convention_resolver.py
```diff
@@ -76,7 +75,7 @@ def resolve(
         timerange: Optional[TimeRange] = None,
         bbox: Optional[Bbox] = None,
         # TODO: align? -> from RasterDatasetAdapter
```
savente93, line 77: What about this TODO? Can we remove it or either link an issue?

savente93 (commented):
Couple of small comments, but otherwise LGTM, once you solve those I'll be happy to approve. Tests look good so if they pass I think it's all good enough to go

hydromt/drivers/preprocessing.py
Jaapel: It is currently present in the unmigrated version of the `DataAdapter`. I can remove it if you want, but that makes the unmigrated code harder to read.

savente93: Then it can stay, we'll just remove it when we migrate that part of the code.

savente93 (approved):
LGTM!

PR #867, opened 2024-04-02
Add `GeomsComponent`

### LINKED ISSUE #825: ``ModelGeomsComponent`` class
### Kind of request

None

### Enhancement Description

Move ``Model.geoms`` to ``ModelGeomsComponent``

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #825 

## Explanation
TBC

## Checklist


savente93:
I think the component itself now functions mostly as intended. However to keep my momentum going I did a very quick implementation of what is almost but not entirely in `ConventionResolver` @Jaapel could you help me see how we can merge the two so we don't have duplicated work?

Jaapel:
> I think the component itself now functions mostly as intended. However to keep my momentum going I did a very quick implementation of what is almost but not entirely in `ConventionResolver` @Jaapel could you help me see how we can merge the two so we don't have duplicated work?

If I understand correctly `capture_glob` tries to read a format string and fills any named captures with `*` and returns a regex to match / replace. The `ConventionResolver` (before resolve_paths), does a similar thing, but also has some default keywords that it fills in like day, month, variable and zoom_level

savente93:
@Jaapel  the point was mainly to find the values of any placeholders that were filled in, so the extra keywords don't really matter to me. I integrated my code into yours and they don't seem to affect each other functionalities at all which is nice. The question that does remain is whether `_expand_uri_placeholders` should remain private now that I'm using it. I think we should not expose it outside hydromt, but I'm not sure what our stance is on using private attributes of other submodules. What do you think?

docs/dev/migrating-to-v1.rst
```diff
@@ -377,6 +377,11 @@ TablesComponent
 
 The previous `Model.tables` is now replaces by a `TablesComponent` that can used to store several non-geospatial tabular data into a dictionary of pandas DataFrames. The `TablesComponent` for now only contains the basic methods such as `read`, `write` and `set`.
 
+GeomsComponent
+^^^^^^^^^^^^^^
+
+The previous `Model.geoms` is now replaces by a `GeomsComponent` that can used to store several geospatial geometry based data into a dictionary of geopandas GeoDataFrames. The `GeomsComponent` for now only contains the basic methods such as `read`, `write` and `set`.
```
hboisgon, line 383:
```suggestion
The previous `Model.geoms` is now replaced by a `GeomsComponent` that can be used to store several geospatial geometry based data into a dictionary of geopandas GeoDataFrames. The `GeomsComponent` for now only contains the basic methods such as `read`, `write` and `set`.
```

hydromt/components/geoms.py
```diff
@@ -0,0 +1,171 @@
+"""Table component."""
```
hboisgon, line 1:
```suggestion
"""Geoms component."""
```

hydromt/components/geoms.py
```diff
@@ -0,0 +1,171 @@
...
+from hydromt.metadata_resolver import ConventionResolver
+
+if TYPE_CHECKING:
+    from hydromt.models.model import Model
+
+_DEFAULT_GEOMS_FILENAME = "geoms/{name}.geojson"
+
+
+class GeomsComponent(ModelComponent):
+    """A component to manage geo-spatial geometries."""
```
hboisgon, line 29: I wonder which docstrings is the best place to describe the component. Is it the class or the __init__ one?
I would be nice if at least mentions the data property and what it returns (dict of geopandas.GeoDataFrame).

Ideally we would have something like [geopandas](https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.html) or [xarray](https://docs.xarray.dev/en/stable/generated/xarray.Dataset.html) too at some point.

hydromt/components/geoms.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(self, model: "Model", filename: str = _DEFAULT_GEOMS_FILENAME):
+        """Initialize a GeomComponent.
+
+        Parameters
+        ----------
+        model: Model
+            HydroMT model instance
+        filename: str
+            The path to use for reading and writing of component data by default.
+            by default DEFAULT_GEOMS_FILENAME
```
hboisgon, line 40: I dont think ``DEFAULT_GEOMS_FILENAME`` will print in this case. Can you use the actual value of the variable "geoms/{name}.geojson" ?
```suggestion
            by default "geoms/{name}.geojson" ie one file per geodataframe in the data dictionnary.
```

hydromt/components/geoms.py
```diff
@@ -0,0 +1,171 @@
...
+        """
+        self._data: Optional[Dict[str, Union[GeoDataFrame, GeoSeries]]] = None
+        self._filename = filename
+        super().__init__(model=model)
+
+    @property
+    def data(self) -> Dict[str, Union[GeoDataFrame, GeoSeries]]:
+        """Model geometries.
+
+        Return dict of geopandas.GeoDataFrame or geopandas.GeoDataSeries
```
hboisgon, line 50:
```suggestion
        Return dict of geopandas.GeoDataFrame or geopandas.GeoSeries
```

hydromt/components/geoms.py
```diff
@@ -0,0 +1,171 @@
...
+                )
+            )
+
+            self._logger.debug(f"Writing file {write_path}")
+
+            write_folder = dirname(write_path)
+            if not isdir(write_folder):
+                os.makedirs(write_folder, exist_ok=True)
+
+            if to_wgs84:
```
hboisgon, line 166: This is only to respect GeoJSON file conventions so this is why we would only like to apply it for geojson files and not other extension as you would be loosing the original CRS which is quite dangerous...
So could add gain the old checks on geojson as well?

```suggestion
            if to_wgs84 and (
                kwargs.get("driver") == "GeoJSON"
                or str(fn).lower().endswith(".geojson")
            ):
```

hydromt/models/model.py
```diff
@@ -71,7 +70,6 @@ class Model(object, metaclass=ABCMeta):
     _API = {
         "crs": CRS,
         "config": Dict[str, Any],
```
hboisgon, line 72: This one should also have been removed no?

hydromt/models/model.py
```diff
@@ -109,6 +108,8 @@ def __init__(
             {"region": {"type": "ModelRegionComponent"}}, components
         )
 
+        self.target_crs = CRS.from_user_input(target_model_crs)
```
hboisgon, line 111: I would say that this will become model CRS, it's the target for the create methods as it can differ from the region crs but in the end it will be model crs. I would also allow to use the keyword "utm"  to let hydromt find the closest projected CRS using crs = gis_utils.parse_crs(crs, bbox=geom.total_bounds).
But you do need a total_bounds for that which you don't have at that stage. Let's leave it at this for now but we will need that when working on the final region PR with Marnix.

tests/components/test_geoms_component.py
```diff
@@ -0,0 +1,63 @@
...
+    bbox = box(*[4.221067, 51.949474, 4.471006, 52.073727], ccw=True)
+    geom = gpd.GeoDataFrame(geometry=[bbox], crs=4326)
+    geom.to_crs(3857, inplace=True)
+    assert geom.crs.to_epsg() == 3857
+
+    write_folder = join(tmpdir, "geoms")
+    write_path = join(write_folder, "test_geom.geojson")
+    geom_component.set(geom, "test_geom")
+    geom_component.write()
+    region_geom = gpd.read_file(write_path)
```
hboisgon, line 61: Could you maybe also test the ``read`` method? I don't think it's done.

tests/components/test_grid_component.py
```diff
@@ -98,9 +98,6 @@ def test_create_grid_from_bbox_rotated(mock_model):
     assert "xc" in grid_component.data.coords
     assert grid_component.data.raster.y_dim == "y"
     assert np.isclose(grid_component.data.raster.res[0], 0.05)
-    # Check set_geoms call
-    geom = grid_component._model.method_calls[0][1][0]
-    assert isinstance(geom, gpd.GeoDataFrame)
```
hboisgon, line 103: You did update that region would be added to geoms though no? Can you check that here?

hboisgon (approved):
Nice job @savente93.
I think mainly docstrings updates and one modification of the ``write`` function are required.
So I approve already so that you can merge after your changes.

tests/components/test_geoms_component.py
savente93: good that you suggested this because it actually caught a bug!

hydromt/models/model.py
savente93: yes good catch!

hydromt/components/geoms.py
savente93: I think the class docstring is better since that will be what is provided on a hover in the code, but we can certainly expand or debate this definition at some point. I've added a mention of the dict of geodataframes for now at least.

tests/components/test_grid_component.py
savente93: I have lost track of the discussion around the region somewhat but my understanding was that the region(s) wouldn't be part of the geoms component anymore since they have their own specialised behaviour

savente93: I'll be merging this one now, but if you think this needs changing we can pick that up in the region PR from Marnix as we do the rest of the things you mentioned.

PR #871, opened 2024-04-08
Remove Network model

### LINKED ISSUE #817: `NetworkModelComponent` class
### Kind of request

### Enhancement Description

currently network models are not implemented, but we do have a class for it, so either the class should move to the component architecture, or w3e should remove the class. 

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #817 

## Explanation
We currently don't have an implementation or a model to base a potential implementations on, so it doens't make much sense to keep this around until we do. Therefore it's better to remove it it and work on it when we have an actual use case. As we haven't been touching the main documentation throughout v1 I haven't updated it there either, we will tackle that when we adress the documentation more directly .

## Checklist


hboisgon (approved):
LGTM!

PR #885, opened 2024-04-12
fix error message for 0D geometry variable in GeoDataset

### LINKED ISSUE #830: handling of single-point geometries by update_geometry() in vector.py
### HydroMT version checks

### Reproducible Example

I am using a python script to create boundary conditions (for Delft3D-FM) using HydroMT. It consists of the code below. A NetCDF is read, a geometry is read, and both are combined into a Dataset with geospatial coordinates. The example below works correctly (as expected).

```
    results = xr.load_dataset(path_netcdf_wflow_results)

    Qup_coord = "Qbound_gauges_fm"
    Qup_data = results["Qbound"].rename({Qup_coord: "index"})
    Qup_data.name = 'Qupstream'

    Qup_gdf = gpd.read_file(path_geojson).copy().set_index("id")
    Qup_gdf.rename(columns={"id": "index"})

    Qup_data = Qup_data.vector.from_gdf(
                gdf=Qup_gdf, data=Qup_data, index_dim='index')

    # Qup_data = Qup_data.sel(index=1)
    Qup_fn = os.path.join(
            path_folder_output, f"{filename_output}_Q_upstream.nc")
    Qup_data.vector.to_netcdf(Qup_fn)
```
However, problems occur when creating a netcdf after indexing Qup_data (see commented line). When indexing a single point, the associated geometry changes from a GeoSeries or GeoDataFrame object into a pandas Series object (see also https://stackoverflow.com/questions/46046000/maintaining-the-geostructure-when-indexing-a-geodataframe). This in itself is not problematic, but an error is raised when using a vector.to_netcdf methods. As a result, the geometry type is not recognized as a GeoSeries and a ValueError is raised due to the following check in update_geometry: (https://github.com/Deltares/hydromt/blob/202874eb4fe3415d0608ea81cd61620af6f5816a/hydromt/vector.py#L308C1-L309C70):

```
        elif not isinstance(geometry, GeoSeries):
            raise ValueError("geometry should be a GeoSeries object")
```
I would say that this is a fault caused by how geopandas works. Still, it affects how HydroMT seems to handle single-point geometries, for example when creating a single-point upstream boundary condition in the form of a netcdf. Currently, I use xarrays .to_netcdf as a workaround instead of vector.to_netcdf

### Current behaviour

Raises a ValueError when dealing with single-point geometries

### Desired behaviour

Handling of single-point geometries by (for example?) converting the Series back to a GeoSeries or GeoDataFrame before raising the ValueError

### Additional context

_No response_

DirkEilander:
## Issue addressed
Fixes #830 

## Explanation
Singe geometry geodatasets with zero dimensions are not discovered. While we can fix this it provides many other issues downstream as we also need a index dimension name which does not exist in this case. Hence I went with improving the GeoDataset typing and error messages.

## Checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -22,6 +22,7 @@ Fixed
 - Bug in `MeshModel.get_mesh` after xugrid update to 0.9.0. (#848)
 - Bug in `raster.clip_bbox` when bbox doesn't overlap with raster. (#860)
 - Allow for string format in zoom_level path, e.g. `{zoom_level:02d}` (#851)
+- Provide better error message for 0D geometry arrays in GeoDataset (#885)
 
 v0.9.4 (2024-02-26)
 ===================
--- a/hydromt/vector.py
+++ b/hydromt/vector.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 import logging
-from typing import Any, Union
+from typing import Any, Hashable, Optional, Union
 
 import geopandas as gpd
 import numpy as np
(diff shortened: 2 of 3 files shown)


hboisgon (approved):
LGTM! And as mentioned in the issue happy with sel using a list as a workaround and only improving the error message here.

PR #888, opened 2024-04-16
Grid inherits from SpatialModelComponent

### LINKED ISSUE #813: Move `setup_grid` and `setup_mesh2d` to region component.
### Kind of request

### Enhancement Description

Currently GridModel and MeshModel have their own fucntionatlity to setup the region. When the new Region component and the Grid/Mesh components are merged, a new PR should be made to move this functionality from the components into the region component so that all the region definitoin logic is in one place. 

### Use case

This should make documentation and organization clearer. 

### Additional Context

_No response_

### LINKED ISSUE #815: Deriving model regions from non-hydromt models
### Kind of request

Adding new functionality

### Enhancement Description

Currently, if we cannot read the model region, we try to derive it from existing model files. We agreed that the components should be able to derive their own region so the users/plugins can overwrite this functionality if necessary. However we must still decide how to decide which components we should derive the region from if we cannot read. 
Ideally which component to derive the region from should be done by the model, but it is unclear yet how this should be specified. 

### Use case

This is to make it possible to hydromt to read modes that were not built by hydromt or where the region file was corrupted or missing. 

### Additional Context

_No response_

### LINKED ISSUE #847: Move region specific code from GridComponent to ModelRegionComponent

deltamarnix:
## Issue addressed
Fixes #847 
Fixes #813 
Fixes #815

## Explanation
Region component is now a new base class that inherits from ModelComponent. Any component that should also have its own region can inherit from SpatialModelComponent. The `_region_data` property can be overridden to return the region that fits the data on that component. Like the region over a grid, or a mesh.

The `region` argument is removed from the command line.

Users can set up their region in a component via the arguments in `create`. `parse_region` can be found as a workflow.

```yaml
global:
  components:
    grid:
      type: GridComponent
steps:
  - grid.create:
      region:
        bbox: [1,2,3,4]
```

The `Model.region` is determined by the only `SpatialModelComponent` in the model. If there are more, it should be specified in the config via `region_component`.

```yaml
global:
  components:
    grid:
      type: GridComponent
    subgrid:
      type: GridComponent
  region_component: grid
```

## Checklist


deltamarnix:
@hboisgon and @savente93 Here is my draft PR. The following things are not yet in:

- [ ] Assigning another region to reference to via the constructor of a component. This might go to a next PR?
- [ ] `Model.crs` seems to have changed in `v1`, and I'm not sure if we should keep this new `target_crs`.
- [ ] `MeshComponent` is not yet a `SpatialModelComponent`, since I want to wait for #792 .

docs/api.rst
```diff
@@ -422,46 +422,6 @@ Attributes
    ModelComponent.logger
    ModelComponent.model_root
 
-ModelRegionComponent
```
deltamarnix, line 425: I should add this back for `SpatialModelComponent`

docs/dev/migrating-to-v1.rst
deltamarnix, line 1: I should update this document

hydromt/components/grid.py
```diff
@@ -155,8 +154,9 @@ def write(
     @hydromt_step
     def read(
         self,
-        fn: str = "grid/grid.nc",
+        filename: str = "grid/grid.nc",
```
deltamarnix, line 157: Is there still another PR that will look into these filenames? Otherwise I have to change it here to `Optional[str] = None`

hydromt/components/grid.py
```diff
@@ -323,23 +306,12 @@ def create(
...
-                ds=ds_hyd,
-                kind=kind,
-                logger=self._logger,
-                **region,
-            )
-            # get ds_hyd again but clipped to geom, one variable is enough
+            # get ds_hyd but clipped to geom, one variable is enough
             da_hyd = self._data_catalog.get_rasterdataset(
-                hydrography_fn, geom=geom, variables=["flwdir"]
+                hydrography_fn, geom=self.region, variables=["flwdir"]
```
deltamarnix, line 312: Instead of `self.region`, I could also use `geom` from line 266

hydromt/components/spatial.py
```diff
@@ -28,27 +29,34 @@
...
+        region_filename: StrPath = DEFAULT_REGION_FILENAME,
     ) -> None:
         super().__init__(model)
-        self._data: Optional[GeoDataFrame] = None
+        self.__filename: StrPath = region_filename
+        self.__data: Optional[GeoDataFrame] = None
 
-    @hydromt_step
-    def create(
+    def create_region(
```
deltamarnix, line 49: Discussed a bit with @hboisgon if these functions should be public or protected (with underscore). In a way this is [protected](https://stackoverflow.com/a/1020776/363224), since it can only be called from child classes. But I understand that it's part of the public API for a python package, but only for plugin developers, and not from a user's perspective.

For now we have gone for public, although my gut feeling says we should make them protected.

hydromt/components/grid.py
```diff
@@ -482,6 +443,7 @@ def data(self) -> xr.Dataset:
         """Model static gridded data as xarray.Dataset."""
         if self._data is None:
             self._initialize_grid()
+        assert self._data is not None
         return self._data
```
deltamarnix, line 448: The `set_crs` function could be a `@crs.setter`. Should we only have this for `GridComponent`, or should every `SpatialModelComponent` be able to do this?

hydromt/components/spatial.py
```diff
@@ -246,20 +210,117 @@ def write(
...
+        elif "geom" in region:
+            geom = region["geom"]
+            if geom.crs is None:
+                raise ValueError('Model region "geom" has no CRS')
+
+        if crs is not None and geom is not None:
+            crs = gis_utils.parse_crs(crs, bbox=geom.total_bounds)
+            geom = geom.to_crs(crs)
+
+        # TODO: This won't be called when subclassed. Should this move to _create_region?
```
deltamarnix, line 312: This TODO should be picked up before merging, although it's only logging.

hydromt/components/spatial.py
```diff
@@ -276,13 +337,13 @@ def test_equal(self, other: ModelComponent) -> tuple[bool, dict[str, str]]:
         eq, errors = super().test_equal(other)
         if not eq:
             return eq, errors
-        other_region = cast(ModelRegionComponent, other)
+        other_region = cast(SpatialModelComponent, other)
 
         if self.kind != other_region.kind:
             errors["kind"] = f"kind {self.kind} != {other_region.kind}"
```
deltamarnix, line 343: self.kind doesn't exist anymore, remove!

hydromt/models/model.py
```diff
@@ -97,10 +98,6 @@ def __init__(
         """
         # Recursively update the options with any defaults that are missing in the configuration.
         components = components or {}
-        components = deep_merge(
-            {"region": {"type": "ModelRegionComponent"}}, components
-        )
-
         self.target_crs = CRS.from_user_input(target_model_crs)
```
deltamarnix, line 101: I'm not sure if we want to have this `target_crs`. Remove and use `region_component.region` instead?

tests/components/test_grid_component.py
```diff
@@ -106,19 +112,19 @@ def test_create_grid_from_bbox(mock_model):
     grid_component.create(
         region={"bbox": bbox},
         res=0.05,
-        add_mask=False,
+        add_mask=True,
```
deltamarnix, line 115: It was unclear from the test if `add_mask` was part of the spec of this test, but otherwise I would retrieve a GeoDataFrame without data and couldn't check the region.

hydromt/components/spatial.py
```diff
@@ -291,6 +352,7 @@ def test_equal(self, other: ModelComponent) -> tuple[bool, dict[str, str]]:
         return len(errors) == 0, errors
 
 
+# TODO: Remove when migrating MeshComponent
```
deltamarnix, line 355: Wait for #792 to implement this part

hydromt/components/spatial.py
```diff
@@ -128,115 +132,75 @@ def create(
...
         **read_kwargs,
-    ):
-        """Read the model region from a file on disk."""
+    ) -> None:
+        """Read the model region from a file on disk. Sets it on the region attribute.
+
+        This function should be called from within the `read` function of the component inheriting from this class.
+        """
         self._root._assert_read_mode()
+        filename = filename if filename is not None else self.__filename
```
savente93, line 189:
```suggestion
        filename = filename or self.__filename
```

hydromt/components/spatial.py
```diff
@@ -246,20 +210,138 @@ def write(
...
                 gdf = gdf.to_crs(4326)
 
             gdf.to_file(write_path, **write_kwargs)
 
+    def parse_region(
+        self,
+        region: dict,
+        *,
+        hydrography_fn: str,
+        basin_index_fn: str,
```
savente93, line 232: shouldn't these be optional so we can use the default ones?

savente93:
I also checked the sonar cloud logs, and they are either about the TODO comments or about the deeply nested structure that was already there. I think this is ok for now, it's better to keep that the way it is and refactor that once we know everything is working well, so it shouldn't not block from merging imo

hydromt/components/spatial.py
deltamarnix: The optional part happens in `create_region`. I really need the files in this function.

hydromt/components/spatial.py
deltamarnix: If the implementor inherits `parse_region`, and it does a early return, because e.g. `if kind == "grid"`, then we don't come through any of this code. And the logging seems to be a part that we would want to have, even if the result was a grid.

tests/models/test_region_parsing.py
```diff
@@ -4,17 +4,11 @@
...
+from hydromt.components.spatial import _parse_region, _parse_region_value
 from hydromt.workflows.basin_mask import _check_size
 
 
-def test_region_from_geom(world):
-    _kind, region = _parse_region({"geom": world})
-    assert isinstance(region["geom"], gpd.GeoDataFrame)
-
-
+# TODO: All these tests should move to test_spatial_component
```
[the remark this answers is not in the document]
deltamarnix: Yes, should be done once we have all those raster data sets and data catalogs in.

hydromt/components/grid.py
```diff
@@ -806,3 +777,35 @@ def test_equal(self, other: ModelComponent) -> tuple[bool, dict[str, str]]:
...
+        """Parse the region dict into a GeoDataFrame.
+
+        Overrides the base functionality to also support "grid".
+        Default options are still available from the base functionality.
+        """
+        kwargs = region.copy()
+        kind = next(iter(kwargs))  # first key of region
+        if kind == "grid":
+            value0 = kwargs.pop(kind)
+            ds = self._data_catalog.get_rasterdataset(value0, driver_kwargs=kwargs)
```
hboisgon, line 798: Allowing for driver_kwargs in region dict is new and should be documented. Also driver_kwargs is used for the data catalog but I don't think this is the argument to use for get_rasterdataset. Unless v1 changed? I couldn't find the get_data method for rasterdataset...

hydromt/components/grid.py
```diff
@@ -806,3 +777,35 @@ def test_equal(self, other: ModelComponent) -> tuple[bool, dict[str, str]]:
...
+            assert ds is not None
```
hboisgon, line 799: again... use NoDataStrategy?

hydromt/components/grid.py
```diff
@@ -806,3 +777,35 @@ def test_equal(self, other: ModelComponent) -> tuple[bool, dict[str, str]]:
...
+            if crs is not None and ds is not None:
```
hboisgon, line 800: didn't you just check that ds was not None?

hydromt/components/grid.py
```diff
@@ -806,3 +777,35 @@ def test_equal(self, other: ModelComponent) -> tuple[bool, dict[str, str]]:
...
+                crs = gis_utils.parse_crs(crs, bbox=ds.total_bounds)
+                ds = ds.to_crs(crs)
```
hboisgon, line 802: This method does not exist :) it's not that simple to reproject a dataset, you need to use the ds.raster.reproject method. Also if you look up, we did not allow a crs different than the grid/ds

hydromt/components/spatial.py
```diff
@@ -28,27 +29,36 @@
...
-DEFAULT_REGION_FILE_PATH = "region.geojson"
 
-
-class ModelRegionComponent(ModelComponent):
+class SpatialModelComponent(ModelComponent, ABC):
     """Define the model region."""
 
+    DEFAULT_REGION_FILENAME = "region.geojson"
+    DEFAULT_HYDROGRAPHY_FILENAME = "merit_hydro"
+    DEFAULT_BASIN_INDEX_FILENAME = "merit_hydro_index"
```
hboisgon, line 38: In other components, the defaults were written outside of the class.
Also hydrography and basin index should not have defaults.

hydromt/components/spatial.py
```diff
@@ -28,27 +29,36 @@
...
     def __init__(
         self,
         model: "Model",
+        region_component: Optional[str] = None,
+        region_filename: StrPath = DEFAULT_REGION_FILENAME,
     ) -> None:
         super().__init__(model)
-        self._data: Optional[GeoDataFrame] = None
+        self.__filename: StrPath = region_filename
+        self.__data: Optional[GeoDataFrame] = None
```
hboisgon, line 48:
```suggestion
        self._filename: StrPath = region_filename
        self._data: Optional[GeoDataFrame] = None
```

hydromt/components/spatial.py
```diff
@@ -28,27 +29,36 @@
...
+        self._region_component = region_component
 
-    @hydromt_step
-    def create(
+    def create_region(
         self,
+        *,
```
hboisgon, line 53: In other methods the * happens after region

hydromt/components/spatial.py
```diff
@@ -28,27 +29,36 @@
...
         region: dict,
-        hydrography_fn: str = "merit_hydro",
-        basin_index_fn: str = "merit_hydro_index",
-    ) -> Dict[str, Any]:
-        """Check and return parsed region arguments.
+        hydrography_fn: str = DEFAULT_HYDROGRAPHY_FILENAME,
+        basin_index_fn: str = DEFAULT_BASIN_INDEX_FILENAME,
```
hboisgon, line 56: Remove defaults

hydromt/components/spatial.py
```diff
@@ -128,115 +134,99 @@ def create(
             * {'interbasin': [xmin, ymin, xmax, ymax], 'xy': [x, y]}
 
             * {'interbasin': /path/to/polygon_geometry, 'outlets': true}
-        logger:
-            The logger to use.
-
-        Returns
-        -------
-        kind : {'basin', 'subbasin', 'interbasin', 'geom', 'bbox', 'grid'}
-            region kind
```
hboisgon, line 137: You still return a geodataframe

hydromt/components/spatial.py
hboisgon: I still think this method should be an external function ie used to be a workflow or a class as Sam started to work on.
It does a lot of nice things for region delineation which could actually be useful outside of a model component: 

- for CST to delineate a region before actually building a model 
- in export_data function to clip data from a data catalog ie no model needed)
- for clipping model (clip command), we also need to derive the region to clip to (which will then become the future model.region after clipping of all components)

hydromt/components/spatial.py
hboisgon: Oh I get it. I think the double underscore is confusing. I'd rather have the spatial component use different properties for the region data and filename else this gets confusing with the actual model data and filename. Maybe more self._region_filename and self.region_data or self._region is maybe more clear. Still not sure I like this definition of region, it conflicts with grid.region definition for example.

hydromt/components/spatial.py
hboisgon: Maybe better in create region indeed

hydromt/components/spatial.py
```diff
@@ -276,21 +384,36 @@ def test_equal(self, other: ModelComponent) -> tuple[bool, dict[str, str]]:
...
+            if region_component.region is None:
+                raise ValueError(
+                    f"Unable to get region from the referenced region component: '{self._region_component}'"
+                )
+            return region_component.region
+        return None
+
 
+# TODO: Remove when migrating MeshComponent
 def _parse_region(
```
hboisgon, line 417: Do you still need this one?

hydromt/models/model.py
hboisgon: I think it's better yes

hydromt/models/model.py
```diff
@@ -121,10 +118,10 @@ def __init__(
         # file system
         self.root: ModelRoot = ModelRoot(root or ".", mode=mode)
 
-        self._components: Dict[str, ModelComponent] = {}
+        self._components: dict[str, ModelComponent] = {}
```
hboisgon, line 121: I thought we needed to use typing and not python types?

hydromt/models/model.py
```diff
@@ -133,6 +130,38 @@ def __init__(
...
 
+        self._region_component_name = self._determine_region_component(region_component)
+
+    def _determine_region_component(self, region_component: Optional[str]) -> str:
+        if region_component is not None:
+            if region_component not in self._components:
+                self.logger.warning(
+                    f"Component {region_component} not found in components."
+                    "You can add it afterwards with add_component."
+                )
```
hboisgon, line 141: I wonder if this is behavior we want to encourage. I think after model init all required components should be initialized even if their data may stay empty.
But maybe we should allow for non spatial models (if they exist?) so we can keep this warning but change the message?

hydromt/models/model.py
```diff
@@ -133,6 +130,38 @@ def __init__(
...
+            has_region_components = [
+                (name, c)
+                for name, c in self._components.items()
+                if isinstance(c, SpatialModelComponent)
+            ]
+            if len(has_region_components) > 1:
+                raise ValueError(
+                    "Multiple region components found in components. "
+                    "Specify region_component."
+                )
```
hboisgon, line 159: Maybe plugins will want to overwrite this (eg model region is actually the total bounds of several components). Then the method should be public and documented.

hydromt/models/model.py
```diff
@@ -157,17 +186,19 @@ def __getattr__(self, name: str) -> ModelComponent:
         return self._components[name]
 
     @property
-    def region(self) -> ModelRegionComponent:
-        """Return the model region component."""
-        return self.get_component("region", ModelRegionComponent)
+    def region(self) -> SpatialModelComponent:
```
hboisgon, line 189: I wonder if here we don't want to return a geodataframe similar to the spatialcomponent.region property? I dont think we need the other properties of the component now anymore (eg _data etc, also as crs can be derived from the geodataframe which maybe our model should still have? a crs property I mean)

tests/components/test_grid_component.py
hboisgon: But the case of add_mask is False should still be handled and should return a return (in that case just raster.box) Which is why I doubt the set_region method and region should really just be derived from the final component geometry.

hydromt/models/model.py
savente93: > I think after model init all required components should be initialized even if their data may stay empty.
Then people will no longer be able to use the Model in ad hoc ways and only plugins can customise the behaviour of the model beyond what we provide at init. Is that what we want?

hydromt/models/model.py
savente93: since python 3.9 they are equivalent, Though I personally prefer using the Typing library since I think it reads a bit better. I think it's mostly due to convention though. We haven't made an agreement on this yet. Do others have a preference?

hydromt/components/grid.py
savente93: I think we should not introduce new functionalities at this point, so if this is new we should remove it again. Besides, any driver kwargs that are needed by default should be listed in the data catalog themselves if I remember correctly, so they are not really necessary at this point, though please correct me if I'm wrong.

docs/dev/migrating-to-v1.rst
```diff
@@ -351,45 +347,44 @@ Making the model region its own component
 The model region is a very integral part for the functioning of HydroMT. Additionally
```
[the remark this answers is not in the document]
deltamarnix: I have tried to update it. Do you like it better this way?

docs/api.rst
```diff
@@ -422,45 +422,41 @@ Attributes
...
-   ModelRegionComponent.bounds
-   ModelRegionComponent.data_catalog
-   ModelRegionComponent.logger
-   ModelRegionComponent.model_root
-
-General Methods
----------------
+   SpatialModelComponent.model
+   SpatialModelComponent.data_catalog
+   SpatialModelComponent.logger
```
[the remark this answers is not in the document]
deltamarnix: You are right, but I also think we are not consistent at the moment. We decided a couple of weeks ago that we should go for `_data_catalog`, because a normal user should retrieve it via `Model.data_catalog` (which is the same reference). Plugin developers can still retrieve this protected property.

But now for `create_region`, we have decided that it should be public, because of plugin developers. We should discuss this again and be consistent.

hydromt/components/spatial.py
savente93: Happy to have more discussions about this, but I don't think that this should be the case. 

I have a couple of problems with putting it outside of the components: 
1. In the past putting this outside of the components caused a lot of problems with circular imports because the components need it, but it in turn uses information from the components. It's not impossible to fix, but it makes the imports very fragile which I don't like
2. I think it makes the logic much harder to follow, since then in that create region we have to do many checks for every possible region, whereas the components can just focus on their specific region kinds as well as create regions specific to them as needed. 
3. If this function is going to be part of our Public API then I would like to write down and agree on what we do and don't allow in there, since now I find it quite fluid. 
4. We're moving in the direction of making the components more responsible for these things, which has the added benefits that if plugins want to add their own region kind they can implement it in their components (perhaps delf3dfm might want to create their own `pipes` region?)

hydromt/models/model.py
deltamarnix: I'm okay using Typing library.

hydromt/components/grid.py
```diff
@@ -36,17 +34,21 @@
...
-class GridComponent(ModelComponent):
+class GridComponent(SpatialModelComponent):
     """ModelComponent class for grid components.
 
     This class is used for setting, creating, writing, and reading regular grid data for a
     HydroMT model. The grid component data stored in the ``data`` property of this class is of the
     hydromt.gis.raster.RasterDataset type which is an extension of xarray.Dataset for regular grid.
     """
 
+    DEFAULT_FILENAME = "grid/grid.nc"
```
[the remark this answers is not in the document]
deltamarnix: What would be the advantage of one over the other? I have also seen some that we prefixed with an underscore.

I chose for this because I can simply call it `DEFAULT_FILENAME` and any inheriting class can access it via `GridComponent.DEFAULT_FILENAME`.

We can also consider making it a property, so that inheriting classes (`MySpecialGridComponent`) can overwrite it as follows:

```py
class GridComponent
  @property
  def default_filename(self):
    return "grid/grid.nc"

  def __init__(self, filename: Optional[str] = None):
    self.__filename = filename or self.default_filename()

class MySpecialGridComponent(GridComponent):
  @property
  def default_filename(self):
    return "some_other_place/special.nc"
```

hydromt/components/grid.py
```diff
@@ -261,34 +271,15 @@ def create(
...
-                bbox = region["bbox"]
-                geom = gpd.GeoDataFrame(geometry=[box(*bbox)], crs=4326)
-            elif kind == "geom":
-                geom = region["geom"]
-                if geom.crs is None:
-                    raise ValueError('Model region "geom" has no CRS')
-            if crs is not None:
-                crs = gis_utils.parse_crs(crs, bbox=geom.total_bounds)
-                geom = geom.to_crs(crs)
             # Generate grid based on res for region bbox
```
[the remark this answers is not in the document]
deltamarnix: no, it's in the workflow

hydromt/components/grid.py
```diff
@@ -350,17 +330,11 @@ def create(
...
-                    align=align,
-                )
+                da_hyd = da_hyd.raster.reproject(dst_crs=crs, dst_res=res, align=align)
+                assert da_hyd is not None
             # Get xycoords, geom
             xcoords = da_hyd.raster.xcoords.values
             ycoords = da_hyd.raster.ycoords.values
-            if geom.crs != da_hyd.raster.crs:
-                crs = da_hyd.raster.crs
-                geom = geom.to_crs(crs)
```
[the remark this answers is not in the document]
deltamarnix: it's in the workflow

hydromt/components/grid.py
```diff
@@ -371,13 +345,11 @@ def create(
             # Get xycoords, geom
             xcoords = da_like.raster.xcoords.values
             ycoords = da_like.raster.ycoords.values
-            geom = da_like.raster.box
```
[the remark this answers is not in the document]
deltamarnix: this is done in the workflow

hydromt/workflows/region.py
```diff
@@ -0,0 +1,293 @@
...
+        elif isinstance(value0, (xu.Ugrid1d, xu.Ugrid2d)):
+            kwarg = dict(
+                mesh=xu.UgridDataset(value0.to_dataset(optional_attributes=True))
+            )
+        else:
+            raise ValueError(
+                f"Unrecognized type {type(value0)}."
+                "Should be a path, data catalog key or xugrid object."
+            )
+        kwargs.update(kwarg)
```
deltamarnix, line 157: There are no real tests where kind is "mesh" I believe. We should check if this is actually working.

hydromt/components/spatial.py
deltamarnix: For now I have moved the code to a `workflow`. A plugin could still decide to parse custom regions within custom components before called `parse_region`.

```py
PipesComponent(SpatialModelComponent):
  def create(self, region):
    if next(iter(region)) == "pipes":
      # do custom stuff
      geom = custom_pipes_region(region)
    else:
      geom = parse_region(region)
```

hydromt/components/spatial.py
deltamarnix: outdated comment

tests/components/test_grid_component.py
deltamarnix: Setting `add_mask=False` gives me a `self.grid` with:

```
<xarray.Dataset> Size: 188B
Dimensions:      (y: 7, x: 16)
Coordinates:
  * y            (y) float64 56B 45.63 45.58 45.53 45.48 45.43 45.38 45.33
  * x            (x) float64 128B 12.08 12.13 12.18 12.23 ... 12.73 12.78 12.83
    spatial_ref  int32 4B 0
Data variables:
    *empty*
```

The functions for `bounds` and `region` in GridComponent now check for `if len(self.data) > 0:`.
This might actually be removed? `self.data.raster.bounds` does contain a value: `(12.05, 45.3, 12.850000000000001, 45.650000000000006)`

hydromt/components/spatial.py
savente93: That is true. If you think it works then I'm happy to go along with it.

tests/models/test_region_parsing.py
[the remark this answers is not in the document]
deltamarnix: outdated

hydromt/components/spatial.py
deltamarnix: outdated

docs/api.rst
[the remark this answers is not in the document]
deltamarnix: The function moved to a workflow, so it's now public anyways

hydromt/components/grid.py
```diff
@@ -177,12 +186,13 @@ def read(
         """
         self._root._assert_read_mode()
         self._initialize_grid(skip_read=True)
+        self.read_region(region_filename, **kwargs)
```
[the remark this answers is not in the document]
deltamarnix: I now have a better understand of this after having a call with @hboisgon . We have decided that read_region is not needed at all. After the region geom is created, it is only used to create the actual data for a grid or a mesh. After that the data is discarded.

Also `write_region` only writes the region data that the grid or mesh is providing. It is not connected to the region that is provided by the user.

hydromt/models/model.py
```diff
@@ -157,17 +183,28 @@ def __getattr__(self, name: str) -> ModelComponent:
...
+        return (
+            cast(
+                SpatialModelComponent, self._components[self._region_component_name]
+            ).region
+            if self._region_component_name in self._components
+            else None
+        )
+
+    @property
+    def crs(self) -> CRS:
```
savente93, line 197: Return type should be `Optional`

hydromt/workflows/region.py
```diff
@@ -0,0 +1,293 @@
...
+    if kind in PLUGINS.model_plugins:
+        model_class = PLUGINS.model_plugins[kind]
+        other_model = model_class(root=value0, mode="r", logger=logger)
+        value0 = other_model.region
+        kwargs = dict(geom=other_model.region)
+        kind = "geom"
+    elif kind == "mesh":
+        if not _compat.HAS_XUGRID:
+            raise ImportError("xugrid is required to read mesh files.")
+
```
savente93, line 143: Since #792 I beleive, xugrid has become a required dependency so this can go.

savente93 (approved):
Looking good, two small pedantic comments, but that's all for me :)

hydromt/components/grid.py
deltamarnix: I have split the driver kwargs and the region_kwargs. Yaml will show:

```ymal
steps:
- grid.create:
    region_options:
      filename: region.geojson
      # other kwargs for the region driver
    # other kwargs for the grid driver
```

hydromt/components/grid.py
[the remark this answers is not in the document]
deltamarnix: @savente93 What do you think? How should we handle `_DEFAULT_FILENAME` or `cls.DEFAULT_FILENAME` or `self.default_filename()`?

hydromt/models/model.py
deltamarnix: No, let's implement a good component that can for example:

```yaml
global:
  region_component: region
  components:
    region:
      type: CompoundRegionComponent
      regions:
        - grid
        - vector
        - subgrid
    grid:
      type: GridComponent
    # etc
```

And implement it like so:

```py
class CompoundRegionComponent(SpatialModelComponent):
  def __init__(model, regions):
    pass

  def region(self):
    # pseudo-code
    expand([r.bounds for r in self.regions])
```

hydromt/workflows/region.py
deltamarnix: pixi.toml and pyproject.toml tell us something else. As long as you don't import a MeshComponent it should be fine, right?

hydromt/workflows/region.py
```diff
@@ -0,0 +1,293 @@
...
+    elif kind == "grid":
+        assert data_catalog is not None
+        ds = data_catalog.get_rasterdataset(value0, kwargs=kwargs)
+        assert ds is not None
+        if crs is not None:
+            logger.warning(
+                "For region kind 'grid', the grid's crs is used and not"
+                f" user-defined crs '{crs}'"
+            )
+        geom = ds.raster.box
```
deltamarnix, line 200: Should "grid" stay in this workflow? Or is it something that `GridComponent` could handle before calling `parse_region`?

hydromt/workflows/region.py
```diff
@@ -0,0 +1,293 @@
...
+    kind = next(iter(kwargs))  # first key of region
+    value0 = kwargs.pop(kind)
+
+    if kind in PLUGINS.model_plugins:
+        model_class = PLUGINS.model_plugins[kind]
+        other_model = model_class(root=value0, mode="r", logger=logger)
+        value0 = other_model.region
+        kwargs = dict(geom=other_model.region)
+        kind = "geom"
+    elif kind == "mesh":
```
deltamarnix, line 140: Should `parse_region("mesh")` be supported? Or is it something that `MeshComponent` should handle instead?

hydromt/workflows/region.py
savente93: Then we forgot to update those and should do so now

hydromt/components/grid.py
```diff
@@ -323,23 +314,12 @@ def create(
...
-                kind=kind,
-                logger=self._logger,
-                **region,
-            )
-            # get ds_hyd again but clipped to geom, one variable is enough
+            # get ds_hyd but clipped to geom, one variable is enough
             da_hyd = self._data_catalog.get_rasterdataset(
                 hydrography_fn, geom=geom, variables=["flwdir"]
             )
+            assert da_hyd is not None
```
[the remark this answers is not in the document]
deltamarnix: mypy doesn't understand, but I think it's good to add this to some of the assertions, as you want to let the user know when the data is incorrect.

hydromt/workflows/grid.py
```diff
@@ -21,6 +26,133 @@
...
+    align: bool = True,
+    dec_origin: int = 0,
+    dec_rotation: int = 3,
+) -> xr.Dataset:
+    # Derive xcoords, ycoords and geom for the different kind options
+    if kind in ["bbox", "geom"]:
+        if not isinstance(res, (int, float)):
+            raise ValueError("res argument required for kind 'bbox', 'geom'")
+        # Generate grid based on res for region bbox
+        # TODO add warning on res value if crs is projected or not?
```
savente93, line 49: Should we do anything with this TODO? Is there an issue?

hydromt/components/grid.py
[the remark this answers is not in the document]
savente93: I think we already have a lot of properties so I'd like to not add more unless really necessary. I don't think there is much bennefit of one over the other, but I agree that we should be consistent. I think the class variable is slightly better in the case of inheritance as Marnix mentioned so unless others have reasons not to, let's go with that. If we do, we should also bring the other components in line

hydromt/workflows/region.py
savente93: I've been thinking about this a bit, but it seems that we'll need a bidirectional relationship between grids and regions (regions can be derived from a grid and vise versa) so I think this is an okay way of doing it for bootstrapping that relationship. It's fairly simple code as well, so I don't think it's a huge issue.

hydromt/workflows/region.py
savente93: As far as I can tell the "mesh" code path doesn't even return a `GeoDataframe`. if that is the case then I think it should either not be supported here or we should find a way to make it compatible with that return type

hydromt/workflows/region.py
savente93: I agree though I don't have enough experience with `xugrid` to do this, so if we decide to keep this I think we should ask  Hélène (or someone else that has experiences with meshes) to do it.

hydromt/workflows/grid.py
deltamarnix: This was already in the original code

hydromt/workflows/region.py
```diff
@@ -0,0 +1,287 @@
...
+            f'Region value "{value0}" for kind={kind} not understood, '
+            f'provide one of "{v_lst}"'
+        )
+    kwargs.update(kwarg)
+
+    if kind in ["basin", "subbasin", "interbasin"]:
+        # retrieve global hydrography data (lazy!)
+        assert data_catalog is not None
+        assert hydrography_fn is not None
+        assert basin_index_fn is not None
```
hboisgon, line 169: No mypy here.
Get_rasterdataset will take care of it if they are None.
AND basin_index_fn is actually allowed to be None!!

hydromt/workflows/region.py
```diff
@@ -0,0 +1,287 @@
...
+        _update_crs(geom, crs)
+    elif kind == "geom":
+        geom = kwargs["geom"]
+        if geom.crs is None:
+            raise ValueError('Model region "geom" has no CRS')
+        _update_crs(geom, crs)
+    elif kind == "grid":
+        assert data_catalog is not None
+        ds = data_catalog.get_rasterdataset(value0, kwargs=kwargs)
+        assert ds is not None
```
hboisgon, line 188: No mypy needed here the NoDataStrategy will take care of this

hydromt/workflows/region.py
hboisgon: And more changes... So actually before kinds model, grid and mesh actually did not return a geodataframe but the object you were asking for ie a grid, a mesh or the model. And the other kinds were also not returning a geodataframe but a dict and only geom option was returning a geodataframe in the dict....

So this is quite a big change. For most kinds a good change but then we have to deal with the other kinds like model, grid and mesh:

- model: as returning the entire model is quite specific for now to Delwaq, I agree to return the model.region and I'll deal with it in the delwaq plugin
- grid and mesh: I really think you should get a grid and mesh objects back here for it to be useful for plugins ie be able to re-use the schematisation and not just the bounding box. This said maybe some models only want it for the extent and want their own shematisation so it may be okay to just return the bounding box after all. But then we will have to duplicates some code again in the plugins if they do want to get the xarray object of gird and mesh... Maybe add two other functions like parse_region_grid and parse_region_mesh that would return xarray objects and in parse_region we can call then and then just create the geom from the bounding box?

hydromt/workflows/region.py
```diff
@@ -0,0 +1,287 @@
...
+            v = f"GeoDataFrame {v.total_bounds} (crs = {v.crs})"
+        elif isinstance(v, xr.DataArray):
+            v = f"DataArray {v.raster.bounds} (crs = {v.raster.crs})"
+        kwargs_str.update({k: v})
+    logger.debug(f"Parsed region (kind={kind}): {str(kwargs_str)}")
+
+    return geom
+
+
+def write_region(
```
hboisgon, line 208: move to io and do not use ModelRoot argument. Check for write permission in the method in SpatialComponent and only pass the root.path here

hydromt/workflows/region.py
```diff
@@ -0,0 +1,287 @@
...
+            write_kwargs.get("driver") == "GeoJSON"
+            or str(filename).lower().endswith(".geojson")
+        ):
+            gdf = gdf.to_crs(4326)
+
+        gdf.to_file(write_path, **write_kwargs)
+
+
+def _update_crs(
+    geom: Optional[gpd.GeoDataFrame], crs: Optional[Union[CRS, int]]
```
hboisgon, line 244: Should optional geom really be allowed?

tests/components/test_grid_component.py
hboisgon: Yes so we need better checks on the grid property in case mask is False indeed

hboisgon (changes requested):
Nice work @deltamarnix ! We are really getting there!
A couple of main things:

- You forgot to update some components that are spatial: geoms, vector (to come: datasets from @savente93 PR)
- The parse_region method is quite different than what is used to be. I made a few suggestions on how we can handle this case, not sure if you agree? Also document the move in the migration guide.

hydromt/components/grid.py
```diff
@@ -13,50 +13,49 @@
...
+        region_component: Optional[str] = None,
+        region_filename: StrPath = SpatialModelComponent.DEFAULT_REGION_FILENAME,
     ):
-        """Initialize a GridComponent.
-
-        Parameters
-        ----------
-        model: Model
-            HydroMT model instance
-        """
```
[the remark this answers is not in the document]
deltamarnix: Because it doesn't really say anything.

hydromt/components/spatial.py
```diff
@@ -0,0 +1,138 @@
...
+from typing import TYPE_CHECKING, Dict, Optional, Tuple, cast
+
+import geopandas as gpd
+import numpy as np
+from geopandas import GeoDataFrame
+from pyproj import CRS
+
+from hydromt._typing.type_def import StrPath
+from hydromt.components.base import ModelComponent
+from hydromt.workflows.region import write_region
```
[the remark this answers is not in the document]
deltamarnix: @savente93 Where should I leave `write_region`? In workflows or in io?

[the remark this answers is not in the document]
savente93: I think `io` is better.

hydromt/models/model.py
deltamarnix: If you are a CLI user, you are unlikely to get this error, as it is common that these users are plugins. Therefore the model should be in place for them.

We might want to make this into an error, but it's quite useful to have this for testing.

hydromt/models/model.py
deltamarnix: We will get conflicts during loading. @savente93 made error handling for this, but it's not possible to load two `GridComponent`s in at the moment.

hydromt/workflows/grid.py
```diff
@@ -21,6 +26,133 @@
...
+        da_hyd = data_catalog.get_rasterdataset(
+            hydrography_fn, geom=geom, variables=["flwdir"]
+        )
+        if da_hyd is None:
+            _exec_nodata_strat(
+                f"No data available in hydrography_fn '{hydrography_fn}' on variable 'flwdir'",
+                NoDataStrategy.RAISE,
+                logger,
+            )
+        assert da_hyd is not None
```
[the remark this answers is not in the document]
deltamarnix: mypy doesn't see that the no-data-strategy will raise and still expects the code to continue. I want to get rid of error lines, so I put the assert there.

hydromt/workflows/grid.py
```diff
@@ -21,6 +26,133 @@
...
+        if res != da_hyd.raster.res and crs != da_hyd.raster.crs:
+            da_hyd = da_hyd.raster.reproject(dst_crs=crs, dst_res=res, align=align)
+            assert da_hyd is not None
+        # Get xycoords, geom
+        xcoords = da_hyd.raster.xcoords.values
+        ycoords = da_hyd.raster.ycoords.values
+    elif kind == "grid":
+        # Get xycoords
+        xcoords = geom.raster.xcoords.values
+        ycoords = geom.raster.ycoords.values
```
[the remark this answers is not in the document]
deltamarnix: Which file exactly?

hydromt/workflows/grid.py
```diff
@@ -21,6 +25,127 @@
...
+        if res is not None:
+            logger.warning(
+                "For region kind 'grid', the grid's res is used and not"
+                f" user-defined res {res}"
+            )
```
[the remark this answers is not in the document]
deltamarnix: There is no test for grids.

Also, I don't think I really understand what's going on. I read in the old code that fn comes from `region[kind]`, which means `region["grid"]`. But that information is no longer available in this function.

How could we implement this?

region.py/parse_region already does the `geom=ds.raster.box`, so I believe that the geom is correct. But you say I should read the file again.

tests/components/test_grid_component.py
deltamarnix: For me it is unclear what I should change now

deltamarnix: Should we check for `if self.data.raster is not None`, or `if len(self.data.coords)>0`?

hydromt/models/model.py
deltamarnix: I will try to make it an error to raise

hydromt/workflows/region.py
```diff
@@ -0,0 +1,287 @@
...
+from os import makedirs
+from os.path import basename, exists, join
+from pathlib import Path
+from typing import Any, Dict, Optional, Union, cast
+
+import geopandas as gpd
+import numpy as np
+import xarray as xr
+import xugrid as xu
+from genericpath import isdir, isfile
```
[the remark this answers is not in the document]
deltamarnix: done

hydromt/workflows/region.py
deltamarnix: I want to let mypy do its thing, otherwise there are red squiggles everywhere. I moved the assert of basin_index_fn inside the if statement.

And get_rasterdataset does not accept null as data_like

hydromt/workflows/region.py
deltamarnix: mypy doesn't know about NoDataStrategy and cannot infer the raise of the error.

hydromt/workflows/region.py
deltamarnix: I have moved this, but region is actually the only piece of code that checks for `overwrite_mode` and now my tests are failing. I am wondering if I should remove the check. Why can't the root folder exist when you want to write only the region?

hydromt/components/vector.py
```diff
@@ -65,10 +66,16 @@ def data(self) -> xr.Dataset:
         assert self._data is not None
         return self._data
 
+    @property
+    def _region_data(self) -> Optional[gpd.GeoDataFrame]:
+        raise AttributeError(
+            "region cannot be found in vector component. Meaning that the region_component is not set or could not be found in the model."
+        )
```
deltamarnix, line 73: This should implement: https://github.com/Deltares/hydromt/blob/f1fad0ef457ce9623fb5969f52bd8bf6f28140f6/hydromt/models/model_vector.py#L376

deltamarnix: Also check test_equal for vector.

hydromt/components/geoms.py
```diff
@@ -62,9 +68,15 @@ def _initialize(self, skip_read=False) -> None:
         """Initialize geoms."""
         if self._data is None:
             self._data = dict()
-            if self._root.is_reading_mode() and not skip_read:
+            if self.root.is_reading_mode() and not skip_read:
                 self.read()
 
+    @property
+    def _region_data(self) -> Optional[GeoDataFrame]:
```
deltamarnix, line 75: Should the region of a geoms component be the union of all bounds?

deltamarnix:
@hboisgon I made changes to the workflow functions for regions and grids. Workflow functions for grid only accept the basic data types, not `dict`. The `parse_region` function is split into multiple functions and components are now in charge of what `kind` it is before calling the correct function. I could even split the basin functions a bit further.

To me it makes sense to have more separate functions as most of the parameters are optional parameters that are only used for specific cases.

I would still love to get rid of `_parse_region_value`, but I first want to know what you think of the new setup.

I did my best to keep all functionality as implemented in main, but please do take a close look to all the different parse_region functions that I made and the way grid and mesh components are implementing `create`.

hboisgon:
@deltamarnix thanks for the last changes! I am now busy reviewing the PR so please do not change anything anymore before I am done! Thanks :)

hboisgon:
Hi @deltamarnix @savente93 and @Tjalling-dejong 

Many many thanks to you all for all the work you did on restructuring the model structure of hydromt! I think it looks really good now!

Why I now tag you all: while reviewing the changes of @deltamarnix for missing docstrings or bugfixing etc I also decided to harmonize and have a second look at all model components and the restructure. So here is a couple of things I would like to draw your attention to and a list of remaining todos.

### Changes 

For all:

- **Model and workflows API docs** is up-to date. Because of the new ``hydromt_step`` functionality, I split components functions into a CLI category (ie available from command line and hydromt workflow file) and the rest.
- The **migration guide** for model should now be up-to-date and I hope will help plugins get started. We can split it for different users after alpha. I did not change much here, just made sure it was complete :)
- added ``test_equal`` methods to all components but mesh
- I decided to document ``Model.components`` property. It's just nice as a python user to glance at your available model components.

For @deltamarnix : 
- **Region_component of grid and mesh**: I think we/I were too harsh when I said that grid region will always come from grid and same for mesh. So I re-added for ``GridComponent`` and ``MeshComponent`` that ``region_component`` can point to another component. What I did instead is that for the ``create`` method, it raises an error if ``self._region_component`` is not None. And in the ``add_data_from_`` methods, I now added some functions to get the grid like or mesh like data either from the component itself or from the reference. I hope you are okay with these changes? This way we do allow a forcing=GridComponent(model, region_component=grid) for which forcing.create_from_region won't work but forcing.add_data_from_rasterdataset will.
- **region_filename**: I changed that each spatial component has its own name for the region filename to avoid conflicts.
- **GeomsComponent**: implemented ``_region_data``
- **DatasetsComponent and SpatialDatasetsComponent**: because of the ``add_raster_data_from`` methods, we needed to have a region component for Datasets. Instead of making ``Datasets`` a ``SpatialComponent``, I decided to split them. Hope you agree. See the line in the comments for Sam for extra details.
- **grid/mesh workflows**: I really like that you split the grid/mesh creation in several functions and decided to make some of them public. I added an extra argument ``region_crs`` to the create method for the user to be able to specify the region crs in case of a bbox or if it cant be inferred from file.
- **parse_region**: I really like the split :) I changed a little though so that the returned region will still be in its original crs. This mimics the data catalog behavior, where we do not reproject data but users can pass the crs of the data in case it cant be inferred from file. I think it is more intuitive that way if the behavior looks a bit like the data catalog.

For @savente93 :
- **DatasetsComponent**: Because of the new SpatialComponent, I decide to split the dict of xarray object into ``DatasetsComponent`` if you do not need a region and ``SpatialDatasetsComponent`` if you do need a region. This is because the ``add_raster_data_from_`` methods do need a region so these have been moved to ``SpatialDatasetsComponent``. The one thing I do not like so much is that again the set/read/write methods are duplicated but I don't think I can easily have ``SpatialDatasetsComponent`` inherit from both ``DatasetsComponent`` and ``SpatialComponent`` no? Else I also added some tests and one is now skipped waiting for artifact data catalog.
- **VectorComponent**: as vector was already different than the rest for default filename definition, I decided to go back to the behavior in main and remove the ``self._filename`` and ``self._geometry_filename`` properties. So filenames and the defaults are just defined in the ``vector.read`` and ``vector.write`` functions.

### TODO

It would be good if you could have a look at the changes and hopefully you agree with these and else we can still discuss here before closing this PR. @deltamarnix could you still check the failing SonarCloud analysis and see if you can improve?

Before alpha release, there are still two main todos on the model side I think:

- generate the docs to see if the API and migration guide print nicely for the plugins to get started with v1
- when artifact data is enabled: **update the mock tests for model and components** --> while implementing my own test for ``SpatialDatasetsComponent``, I realize that the mock test (at least for grid) would fail with actual data. To ensure we have the same behavior as before, I would really like to get back to the model tests from main. For some tests we can use fixture data and this is fine as the get_data method do handle python objects but the create and add_data_from functions should not be mocked to ensure they are working the way we expect them to (eg variable selection, renaming, reprojection etc). Do you also see this?

Once these are in, I think we are finally done with Model!!
Again great job everyone and I am really happy with the progress and how the restructure looks like!

savente93:
I did not review the whole PR again, but I'm happy with the changes as described. As far as I'm concerned we're ready to merge, we can address the enabling of tests in a separate PR (imo) as this one is already quite massive.

hydromt/components/geoms.py
```diff
@@ -47,10 +54,22 @@ def __init__(
...
+            region. If None, the region will be set to the union of all geometries in
+            the data dictionary.
+        region_filename: str, optional
+            The path to use for writing the region data to a file. By default
+            "geoms/geoms_region.geojson".
         """
         self._data: Optional[Dict[str, Union[GeoDataFrame, GeoSeries]]] = None
         self._filename: str = filename or self.__class__.DEFAULT_FILENAME
-        super().__init__(model=model, region_component=region_component)
+        region_filename = region_filename or self.__class__.DEFAULT_REGION_FILENAME
```
deltamarnix, line 67: Because `DEFAULT_REGION_FILENAME` is overriding the original `DEFAULT_REGION_FILENAME`, there should be no need for this line 67 because the base class should pick up the right one.

hydromt/components/grid.py
```diff
@@ -681,3 +719,23 @@ def test_equal(self, other: ModelComponent) -> Tuple[bool, Dict[str, str]]:
             errors["data"] = str(e)
 
         return len(errors) == 0, errors
+
+    def _get_grid_data(self) -> Union[xr.DataArray, xr.Dataset]:
+        """Get grid data as xarray.DataArray from this component or the reference."""
+        if self._region_component is not None:
+            region_component = cast(
+                GridComponent, self.model.get_component(self._region_component)
+            )
```
deltamarnix, line 728: I don't think you can just assume that the `region_component` used is another grid. If that is the case I would change the name of the variable in `__init__` to `grid_reference` and also perform a check here with: `if not isinstance(grid_component, GridComponent): raise ValueError`

hydromt/components/mesh.py
```diff
@@ -680,6 +719,50 @@ def _grid_is_equal(self, grid_name: str, data: xu.UgridDataset) -> bool:
...
 
+    def _get_mesh_grid_data(self, grid_name: str) -> Union[xu.Ugrid1d, xu.Ugrid2d]:
+        if self._region_component is not None:
+            region_component = cast(
+                MeshComponent, self.model.get_component(self._region_component)
+            )
+            if region_component is None:
+                raise ValueError(
+                    f"Unable to find the referenced region component: '{self._region_component}'"
+                )
```
deltamarnix, line 730: Same as in Grid, I think it's good to check if the component is actually a mesh and raise if it's not

hydromt/components/vector.py
```diff
@@ -175,8 +188,8 @@ def set(
     def read(
         self,
         *,
-        filename: Optional[str] = None,
-        geometry_filename: Optional[str] = None,
+        filename: Optional[str] = "vector/vector.nc",
+        geometry_filename: Optional[str] = "vector/vector.geojson",
```
deltamarnix, line 192: This is old style again. We have come to the decision that the optionals are by default None and will then use the self.filename, which is by default DEFAULT_FILENAME

hydromt/workflows/grid.py
```diff
@@ -110,16 +113,25 @@ def create_grid_from_region(
 
     kind = next(iter(region))
     if kind in ["bbox", "geom"]:
-        if not res:
+        if not isinstance(res, (int, float)):
```
deltamarnix, line 116: What is the difference  to check if it is int or float? The spec says it's an optional float, so it can be either float or None. Can you elaborate on why this check is better?

hydromt/workflows/mesh.py
```diff
@@ -84,38 +88,65 @@ def create_mesh2d_from_region(
...
+            clip_to_geom = False
+        else:
+            geom = parse_region_geom(region, crs=region_crs, data_catalog=data_catalog)
+            clip_to_geom = True
+        if crs is not None:
+            crs = utils.parse_crs(crs, bbox=geom.total_bounds)
+            geom = geom.to_crs(crs)
+        return create_mesh2d_from_geom(
+            geom, res=res, align=align, clip_to_geom=clip_to_geom
+        )
```
deltamarnix, line 113: This is way harder to read than what I had.

In that case I would try to go for 

```py
if kind in ["bbox", "geom"]:
  if not res:
    raise

  geom = parse_region_bbox() if kind == "bbox" else parse_region_geom()
  clip_to_geom = kind == "geom"

  if crs is not None:
    crs = ...
    geom = geom.to_crs(crs)

  return create_mesh2d_from_geom()
```

hydromt/workflows/mesh.py
```diff
@@ -84,38 +88,65 @@ def create_mesh2d_from_region(
...
 
     raise ValueError(f"Unsupported region kind '{kind}' found in grid creation.")
 
 
-def _create_mesh2d_from_mesh(
+def create_mesh2d_from_mesh(
```
deltamarnix, line 118: I see you made these functions public, should they go into `__all__` as well?

hydromt/components/geoms.py
hboisgon: Ok then we can remove indeed

hydromt/components/grid.py
hboisgon: For the methods called add_data_from I do need a grid for reprojection. I thought that calling cast would do the type checking for me already? If not what does cast actually do then?

hydromt/components/mesh.py
hboisgon: Same reply as above. What does cast actually do if not type checking already?

hydromt/components/vector.py
hboisgon: It's back to old style because for these components, we should allow for None in one of the files (ie output file can be netcdf only or geometry only or netcdf+geometry). With the default filenames, it is too tough to specify that you want None for one of them so I went back to the old behavior to be sure a user can overwrite defaults easily if he is not happy with them without having to change the component instantiation in the yaml which I find complex.

hydromt/workflows/grid.py
hboisgon: The spec says but not type check is actually done by the spec no?

hydromt/workflows/mesh.py
hboisgon: Sure that also works. I just am not too found of if and else statement in one line but it works if you prefer it.

hydromt/workflows/mesh.py
hboisgon: Good one!

hydromt/components/grid.py
deltamarnix: `cast()` is an empty method that only helps the static type checker to understand that this value is, according to the developer, indeed of that type.

It forces `region_component` to be of `GridComponent` for the type checker, but the function `cast` is an empty function and during runtime it doesn't do anything.

So if you want to know if the value is of a correct type and otherwise use `None`, or `raise`, you must use `isinstance` instead.

hydromt/components/vector.py
deltamarnix: Alright, that makes sense. In that case I think we should continue with this in all other components as well and make some of them non-optional. @savente93 , do you agree?

hydromt/workflows/grid.py
deltamarnix: Correct, but are you expecting someone to put a `string` there? Otherwise we have to check all incoming arguments all the time.

`res` is an `optional[float]`. That's what we expect, and we want to know if it's not `None`. If you expect an incoming `int`, then we should change the argument spec to `Optional[Union[int,float]]`. But in runtime I would only check `if not None`.

savente93 (approved):
I'm okay with the current changes.

tests/components/test_grid_component.py
deltamarnix: @hboisgon This is still a thing after all the refactors. The code in `GridComponent._region_data` still checks for `if len(self.data) > 0:`. What should we make of this?

hboisgon (approved):
Good to go now!!

tests/components/test_grid_component.py
hboisgon: I think we decided to leave it for now as it did not work for main already and open an issue to fix it for later as it may not be fully straightforward to fix

savente93:
whoooo! well done everyone! great work <3

PR #894, opened 2024-04-18
Dataset component

### LINKED ISSUE #824: ``ModelXArrayDictComponent`` class (forcing, maps, results, states)
### Kind of request

None

### Enhancement Description

Move ``Model.forcing`` to ``ModelForcingComponent``

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #824 

## Explanation
The two `add_data` functions (used to be `setup_maps` functions) are not tested currently as these need the data catalog, but we'll test them again when we reenable all the integration tests. Hopefully that's okay for now. 

## Checklist
- [x] Branch is up to date with `v1


hydromt/models/model.py
```diff
@@ -54,22 +47,11 @@ class Model(object, metaclass=ABCMeta):
     Inherit from this class to pre-define mandatory components in the model.
     """
 
-    _DATADIR = ""  # path to the model data folder
     _NAME: str = "modelname"
-    _MAPS = {"<general_hydromt_name>": "<model_name>"}
-    _FOLDERS = [""]
```
hboisgon, line 60: This is behavior that we are deleting so we should updtae the changelog / migration guide.
Any subfolders listed was generated automatically during model __init__ so for plugins, this may break some of their writing methods in case they do not create these subfolders.

hydromt/models/model.py
```diff
@@ -54,22 +47,11 @@ class Model(object, metaclass=ABCMeta):
...
     # supported model version should be filled by the plugins
     # e.g. _MODEL_VERSION = ">=1.0, <1.1"
     _MODEL_VERSION = None
 
-    _API = {
-        "maps": XArrayDict,
-        "forcing": XArrayDict,
-        "results": XArrayDict,
-        "states": XArrayDict,
-    }
```
hboisgon, line 71: Do we completely delete? Or leave the dict empty in core? Maybe these API tests are not so needed anymore because we use pydantic and do better type checking in core?

hydromt/models/model.py
```diff
@@ -161,19 +137,6 @@ def region(self) -> ModelRegionComponent:
...
-        _api = cls._API.copy()
-
-        # reversed is so that child attributes take priority
-        # this does mean that it becomes imporant in which order you
-        # inherit from your base classes.
-        for base_cls in reversed(cls.__mro__):
-            if hasattr(base_cls, "_API"):
-                _api.update(getattr(base_cls, "_API", {}))
-        return _api
-
```
hboisgon, line 176: Not sure if it was in the API docs but as the plugins used it we should document. Maybe add a section in the migration with the properties of model that were deleted/updated etc for properties other than the components?

tests/models/test_model.py
```diff
@@ -79,24 +78,6 @@ def test_api_attrs():
 #         _MODELS["unknown"]
 
 
-def test_check_data(demda):
```
hboisgon, line 82: Was this test transferred somewhere else prior to this PR?

hboisgon (changes requested):
Nice @savente93 ! Just a couple of comments but nothing major I think.
I'm hesitating on the name to use the plural to be more in check with other components like tables and geoms. What do you think?

docs/api.rst
```diff
@@ -545,6 +497,38 @@ General methods
    TablesComponent.write
    TablesComponent.read
 
+DatasetComponent
+==============
+
+.. autosummary::
+   :toctree: _generated
+
+    DatasetComponent
```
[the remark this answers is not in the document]
savente93: Good point, that was my oversight, I think you are correct. I'll change it.

hydromt/components/dataset.py
```diff
@@ -0,0 +1,405 @@
...
+from hydromt._typing.type_def import XArrayDict
+from hydromt.components.base import ModelComponent
+from hydromt.hydromt_step import hydromt_step
+from hydromt.io.readers import read_nc
+from hydromt.io.writers import write_nc
+
+if TYPE_CHECKING:
+    from hydromt.models.model import Model
+
+_DEFAULT_DATASET_FILENAME = "{name}.nc"
```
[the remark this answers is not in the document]
savente93: I did not do it because we can't give multiple folders here such as `forcings` as it's a generic component but if you think it's better I'm happy to add one.

hydromt/components/dataset.py
```diff
@@ -0,0 +1,405 @@
...
+    def _initialize(self, skip_read=False) -> None:
+        """Initialize geoms."""
+        if self._data is None:
+            self._data = dict()
+            if self._root.is_reading_mode() and not skip_read:
+                self.read()
+
+    def set(
+        self,
+        data: Union[Dataset, DataArray],
```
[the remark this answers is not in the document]
savente93: Shouldn't dataframes go in the `tables` component?

hydromt/components/dataset.py
```diff
@@ -0,0 +1,405 @@
...
+        ---------
+        data: xarray.Dataset or xarray.DataArray
+            New xarray object to add
+        name: str
+            name of the xaray.
+        """
+        self._initialize()
+        if isinstance(data, DataFrame):
+            data = data.to_xarray()
+        ds: XArrayDict
```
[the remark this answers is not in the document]
savente93: no preference, I'll use a cast instead, but the behaviour is the same

hydromt/components/dataset.py
```diff
@@ -0,0 +1,405 @@
...
+            Additional keyword arguments that are passed to the `to_netcdf`
+            function.
+        """
+        self._root._assert_write_mode()
+
+        if len(self.data) == 0:
+            self._logger.debug("No data found, skiping writing.")
+            return
+
+        kwargs = {**{"engine": "netcdf4"}, **kwargs}
```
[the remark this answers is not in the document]
savente93: if I remember correctly because tests fail without an engine being specified. So this is only in the case there the user did not specify a different driver, and I believe that `netcdf4` is the most common one? If you think this is incorrect I'm happy to see if we can find a different solution though.

hydromt/components/dataset.py
```diff
@@ -0,0 +1,405 @@
...
+            raster_fn,
+            geom=self._model.region.data,
+            buffer=2,
+            variables=variables,
+            single_var_as_array=False,
+        )
+        if isinstance(ds, DataArray):
+            ds = ds.to_dataset()
+
+        ds = cast(Dataset, ds)
```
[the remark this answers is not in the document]
savente93: In my case I got an error from mypy without it, though functionally it is not necessary. I also think there's not really a downside to it either though

hydromt/models/model.py
savente93: Indeed, these are not necessary any more as they are taken care of by the component architecture.

hydromt/models/model.py
```diff
@@ -54,22 +47,11 @@ class Model(object, metaclass=ABCMeta):
     Inherit from this class to pre-define mandatory components in the model.
     """
 
-    _DATADIR = ""  # path to the model data folder
     _NAME: str = "modelname"
-    _MAPS = {"<general_hydromt_name>": "<model_name>"}
```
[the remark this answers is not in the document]
savente93: I think you're correct that this should be better doccumented, I'll try and add something to the migration guide for that, though I also think it's good to notice that the functionality is not gone, it has just moved to the components.

hydromt/models/model.py
savente93: I've added it to the migration guide.

tests/models/test_model.py
savente93: As far as I understand it is no longer necessary as this does a lot of type checking and name checking that is now done by the component and therefore no longer necessary. The dataset component tests include some equality checking which should also ensure the names are correct (at least I would assume). So as far as I understand we've covered all the necessary functionality in other tests.

hydromt/io/readers.py
```diff
@@ -862,9 +862,9 @@ def parse_values(
 
 
 def read_nc(
-    fn: StrPath,
-    root,
-    logger: Logger,
+    filename_template: StrPath,
```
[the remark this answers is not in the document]
savente93: I disagree. I would argue that this is more accurate, as the template can specify multiple files, not just one. Could you explain a bit more why you think `fn` is more explicit?

savente93:
@hboisgon  thanks for the review! I have fixed most of your comments. There are still a few open conversations that we should come to an agreement on. We don't have our catchup this week, but if you want then you can schedule something with me at the usual time if you would like to discuss some of the things still open, so hopefully we can still merge this tomorrow and Marnix can continue with his part. Let me know what works for you, I'll check my calendar early tomorrow in case you want to schedule something.

docs/dev/migrating-to-v1.rst
```diff
@@ -506,6 +511,25 @@ of the `ModelComponent`: `MeshComponent`. The restructering of `MeshModel` follo
...
+	class MyModel(Model):
+	...
+	def __init__(self):
+		...
+		config_component = ConfigComponent(self, filename="my_plugin_default_config.yaml")
+		self.add_component("config", config_component)
+
+	```
+	When done correctly the behaviour should be the same as in v0.x in this regard.
+- **_FOLDERS**: Since the components are now responsible for creating their folders when writing, we no longer have a `_FOLDERS` attribute and the `Model` will no longer create the folders during model init. This was done to provide more flexibility in which folders need to be created and which do not need to be. Components should make sure that they create the necessary folders themselves during writing.
```
hboisgon, line 531: Shame, since the changes down are too big for tracking... mainly spelling mistakes and using rst file code-block for the python part. Content wise it's good :)

```suggestion
Below you will find a summary of the functionalities, features, attributes and other things that were removed from the `Model` class for v1 and how you can access their new equivalents.

- **api**: The `api` property and its associated attributes such as `_API` were previously provided to the plugins to enable additional validation. These have been superseded by the component architecture and have therefore been removed. Except in the case of equality checking (which will be covered separately below) plugins do not need to access any replacement functionality. All the type checking that was previously handled by the `api` property is now performed by the component architecture itself. If you use components as instructed they will take care of the rest for you.
- **_MAPS/_CONF/_GEOMS/etc.**: As most aspects are now handled by the components, their model level attributes such as `_CONF` or `_MAPS` have been removed. The same functionality/ convention can still be used by setting these in the components. For example, if you want to tell HydroMT that the default config for your plugin/model is called `my_plugin_default_config.yaml` you can do so by setting the `filename` attribute of the `ConfigComponent` at init like so:

.. code-block:: python

	class MyModel(Model):
	...
	def __init__(self):
		...
		config_component = ConfigComponent(self, filename="my_plugin_default_config.yaml")
		self.add_component("config", config_component)

	When done correctly the behaviour should be the same as in v0.x in this regard.
- **_FOLDERS**: Since the components are now responsible for creating their folders when writing, we no longer have a `_FOLDERS` attribute and the `Model` will no longer create the folders during model init. This was done to provide more flexibility in which folders need to be created and which do not need to be. Components should make sure that they create the necessary folders themselves during writing.
```

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+from hydromt.io.writers import write_nc
+
+if TYPE_CHECKING:
+    from hydromt.models.model import Model
+
+_DEFAULT_DATASET_FILENAME = "datasets/{name}.nc"
+
+
+class DatasetsComponent(ModelComponent):
+    """A component to manage colections of Xarray objects.
```
hboisgon, line 30:
```suggestion
    """A component to manage collections of Xarray objects.
```

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+            if self._root.is_reading_mode() and not skip_read:
+                self.read()
+
+    def set(
+        self,
+        data: Union[Dataset, DataArray],
+        name: Optional[str] = None,
+        split_dataset: bool = False,
+    ):
+        """Add data to the xrarray component.
```
hboisgon, line 77:
```suggestion
        """Add data to the xarray component.
```

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+    def _initialize(self, skip_read=False) -> None:
+        """Initialize geoms."""
+        if self._data is None:
+            self._data = dict()
+            if self._root.is_reading_mode() and not skip_read:
+                self.read()
+
+    def set(
+        self,
+        data: Union[Dataset, DataArray],
```
hboisgon, line 73: I guess the conversation was outdated but we can add DataFrame here and keep the lines that convert to_xarray

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+    ) -> None:
+        r"""Read model dataset files at <root>/<filename>.
+
+        key-word arguments are passed to :py:func:`hydromt.io.readers.read_nc`
+
+        Parameters
+        ----------
+        filename : str, optional
+            filename relative to model root. should contain a {name} placeholder
+            which will be used to determine the names/keys of the geometries.
```
hboisgon, line 125:
```suggestion
            which will be used to determine the names/keys of the datasets.
```

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+            if None, the path that was provided at init will be used.
+        **kwargs:
+            Additional keyword arguments that are passed to the
+            `hydromt.io.readers.read_nc` function.
+        """
+        self._root._assert_read_mode()
+        self._initialize(skip_read=True)
+        kwargs = {**{"engine": "netcdf4"}, **kwargs}
```
hboisgon, line 133: All the conversation got outdated but I like your explanation :) Maybe so that user can overwrite, can you instead check if engine not in kwargs then engine =netcdf4 or does this line takes care of that somehow? That the original kwargs get priority?

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+            Additional keyword arguments that are passed to the `to_netcdf`
+            function.
+        """
+        self._root._assert_write_mode()
+
+        if len(self.data) == 0:
+            self._logger.debug("No data found, skiping writing.")
+            return
+
+        kwargs = {**{"engine": "netcdf4"}, **kwargs}
```
hboisgon, line 190: Same remark here as above

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+        variables: Optional[List] = None,
+        fill_method: Optional[str] = None,
+        name: Optional[str] = None,
+        reproject_method: Optional[str] = None,
+        split_dataset: bool = True,
+        rename: Optional[Dict[str, str]] = None,
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Add data variable(s) from ``raster_filename`` to dataset object.
+
+        If raster is a dataset, all variables will be added unless ``variables``
```
hboisgon, line 258:
```suggestion
        If raster is a xarray dataset, all variables will be added unless ``variables``
```

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+        self,
+        raster_filename: Union[str, Path, Dataset],
+        variables: Optional[List] = None,
+        fill_method: Optional[str] = None,
+        name: Optional[str] = None,
+        reproject_method: Optional[str] = None,
+        split_dataset: bool = True,
+        rename: Optional[Dict[str, str]] = None,
+    ) -> List[str]:
+        """HYDROMT CORE METHOD: Add data variable(s) from ``raster_filename`` to dataset object.
```
hboisgon, line 256:
```suggestion
        """HYDROMT CORE METHOD: Add data variable(s) from ``raster_filename`` to datasets component.
```

hydromt/components/datasets.py
```diff
@@ -0,0 +1,407 @@
...
+        reclass_variables: List,
+        variable: Optional[str] = None,
+        fill_method: Optional[str] = None,
+        reproject_method: Optional[str] = None,
+        name: Optional[str] = None,
+        split_dataset: bool = True,
+        rename: Optional[Dict] = None,
+        **kwargs,
+    ) -> List[str]:
+        r"""HYDROMT CORE METHOD: Add data variable(s) to dataset object by reclassifying the data in ``raster_filename`` based on ``reclass_table_filename``.
```
hboisgon, line 330:
```suggestion
        r"""HYDROMT CORE METHOD: Add data variable(s) to datasets component by reclassifying the data in ``raster_filename`` based on ``reclass_table_filename``.
```

hydromt/io/readers.py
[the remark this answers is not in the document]
hboisgon: Indeed forgot we explicitly ask for {name} placeholder at least in ``write_nc`` to be in the path so filename_template sounds good.
For ``read_nc`` it's less clear though as it just says may contain wildcard and I think in main this just meant you can use * and the name is the basename of the file. Here is seems behavior may be different? Just we just be more explicit on a {name} placeholder like in write_nc ?

tests/models/test_model.py
hboisgon: Ah yes I did not realize you actually removed the corresponding function so this is fine!

hboisgon (approved):
Almost good to go @savente93 ! Thanks for the changes.
Mainly some last spelling mistakes ;) And for read_nc one final remark but I am happy to merge as is if you disagree.

hydromt/components/datasets.py
savente93: This does the same, if `engine` is in `kwargs` it will get priority. Though if you like I'm happy to do it the manual way.

hydromt/io/readers.py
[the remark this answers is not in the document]
savente93: well, I changed the behavior slightly since I wanted to make them more consistent. now `read_nc` actually also expects a `{name}` placeholder that will be used to determine the name of the dataset which will be used in the component. So now the same paths can be used for reading and writing. Therefore I think it's best to keep the names in both functions as well.

PR #904, opened 2024-04-22
add options for driver

### LINKED ISSUE #899: Replace `Driver` extra_fields with `options` that get passed like `driver_kwargs` in `v0.x`
### Kind of request

### Enhancement Description

Currently the idea was to put the previous `driver_kwargs` into the `Driver` model using [`extra=allow`](https://github.com/Deltares/hydromt/blob/v1/hydromt/drivers/base_driver.py#L23). It is better to make this behaviour explicit and call it `options`. We should document per driver where we pass these options, e.g. pass them to [`xr.read_zarr`](https://github.com/Deltares/hydromt/blob/v1/hydromt/drivers/zarr_driver.py#L63)

### Use case

Pass chunks, or other specific options to the driver

### Additional Context

_No response_

Jaapel:
## Issue addressed
Fixes #899

## Explanation
Added options in addition to runtime **kwargs.

## Checklist

### DIFF
--- a/docs/dev/migrating-to-v1.rst
+++ b/docs/dev/migrating-to-v1.rst
@@ -165,12 +165,12 @@ reading phase, the data catalog is updated accordingly:
 			name: netcdf
 			filesystem: local
 			metadata_resolver: convention
-			chunks:
-				latitude: 250
-				longitude: 240
-				time: 30
-			combine: by_coords
-			...
+			options:
+				chunks:
+					latitude: 250
+					longitude: 240
+					time: 30
+				combine: by_coords
 		data_adapter:
 			meta:
 				category: meteo
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -160,6 +160,7 @@ General
    :toctree: _generated
 
    drivers.base_driver.BaseDriver
+   drivers.base_driver.read
 
 RasterDataset
 -------------
@@ -167,6 +168,7 @@ RasterDataset
 .. autosummary::
    :toctree: _generated
    drivers.rasterdataset_driver.RasterDatasetDriver
+   drivers.rasterdataset_driver.RasterDatasetDriver.read
    drivers.rasterdataset_driver.RasterDatasetDriver.read_data
 
 ZarrDriver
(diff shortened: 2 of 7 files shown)


Jaapel:
@DirkEilander @savente93 I am wondering about how to deal with `MetaDataResolver.resolve` `**kwargs` and `Driver.read` `**kwargs`. do we just accept a `resolver_kwargs` `dict` and a `driver_kwargs` `dict` in the `DataSource.read`?

savente93:
if at all possible I think we should keep different sets of kwargs separate otherwise we're gonna end up with spaghetti kwargs all over the place. So I think what you suggest is fine, but I haven't looked at the PR yet which I'll do in a little bit.

docs/dev/migrating-to-v1.rst
```diff
@@ -165,12 +165,12 @@ reading phase, the data catalog is updated accordingly:
 			name: netcdf
 			filesystem: local
 			metadata_resolver: convention
-			chunks:
-				latitude: 250
-				longitude: 240
-				time: 30
-			combine: by_coords
-			...
+			options:
```
savente93, line 168: `driver_options`? At least I'd like to see some documentation clarifying the difference between `options`, `kwargs`, `driver_kwargs` and any other variation I might have forgotten.

Jaapel: I aim to replace driver_kwargs with options under the driver header, I do not like:
```yaml
driver:
  driver_options:
    ...
```

savente93: ah right, sorry did not properly read the whole context. I think you're right about your example, but I think we still need an internal guideline on how we talk about these because the different kinds is starting to get hard to keep in (human) memory

Jaapel: I updated the issue description to make the use case more clear

hydromt/metadata_resolver/convention_resolver.py
```diff
@@ -89,19 +90,24 @@ def _resolve_wildcards(
...
         variables: Optional[List[str]] = None,
         handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
         logger: Logger = logger,
-        **kwargs,
     ) -> List[str]:
         """Resolve the placeholders in the URI."""
+        warn_on_unused_kwargs(
+            self, dict(zip(("mask", "zoom_level"), (mask, zoom_level))), logger
+        )
+
```
Jaapel, line 110: @savente93 I used two methods of annotating unused kwargs. I can also put the `dict(zip` inside the warn function. Which one do you prefer and why?

savente93: sorry I don't quite understand the question, can you either rephrase or call me about this? thanks

Jaapel: @savente93 based on our discussion we chose to go for a simple function. I implemented this in the next commits.

savente93 (approved):
Nice work Jaap!

Issue #875, opened 2024-04-09
New `Driver` to read slippy map tile data


Jaapel:
### Kind of request

Changing existing functionality

### Enhancement Description

In https://github.com/Deltares/hydromt/blob/main/hydromt/data_adapter/rasterdataset.py#L360 there is still some custom validation for zoom_level that is not yet present in v1. This can better be moved to a new specific `Driver` or `URIResolver` that specializes in reading slippy map tiles.

In gis/raster.py there is a `to_slippy_tiles` function that results in a DataCatalog entry referring to the previous `raster` driver, which should refer to this new driver.

### Use case

Used if you want to match a certain zoom_level based on your search query.

PR #920, opened 2024-05-07
857/raster driver

### LINKED ISSUE #857: Create `raster` driver
### Kind of request

### Enhancement Description

Create a `raster` driver using the `rasterio` package.

Also: at https://github.com/Deltares/hydromt/blob/v0.9.4/hydromt/data_adapter/rasterdataset.py#L661 zoom levels are parsed based on the rasterio metadata.

In gis/raster.py there is a to_slippy_tiles function that results in a DataCatalog entry referring to the the previous raster driver, which should refer to a new driver #875 . This behaviour can be temporarily be deprecated.

### Use case

Open data using `rasterio`.

### Additional Context

_No response_

Jaapel:
## Issue addressed
Closes #857 

## Explanation
- Created the raster driver based on functionality previously in `io.open_raster` and `io.open_mfraster`
- Added `SourceMetadata` to the `read` and `read_data` signatures of drivers. The raster driver needs this, thus `metadata` is not only for the `data_adapter`.
- Also did some pre-work for #875 while migrating old code.

## Checklist


hydromt/drivers/dataframe/dataframe_driver.py
```diff
@@ -17,6 +21,7 @@ class DataFrameDriver(BaseDriver, ABC):
     def read(
         self,
         uri: str,
+        metadata: "SourceMetadata",
```
Jaapel, line 24: @savente93 I can also make the metadata optional (would just have to move the Pydantic model to circumvent cyclic imports), what would you prefer?

hydromt/metadata_resolver/slippy_map_tile_resolver.py
```diff
@@ -0,0 +1,121 @@
...
+        res = list(zls_dict.values())[0] / 2
+        zls = list(zls_dict.keys())
+        smaller = [x < (dst_res + res * 0.01) for x in zls_dict.values()]
+        zl = zls[-1] if all(smaller) else zls[max(smaller.index(False) - 1, 0)]
+    elif dst_crs is None:
+        raise ValueError("No CRS defined, hence no zoom level can be determined.")
+    else:
+        raise TypeError(f"zoom_level not understood: {type(zoom_level)}")
+    logger.debug(f"Using zoom level {zl} ({dst_res:.2f})")
+    return zl
```
savente93, line 121: Sonar cloud remarks that this is duplicated from `data_adapter/rasterdataset.py` if it's no longer needed there please remove it, or if it is, could we move it to a common place like the utils?

hydromt/metadata.py
savente93, line 1: Should this be here in the root or in `_type_def` ?

hydromt/config.py
savente93, line 1: is this necessary? seems a little overkill for something we don't really offer any customization on. Would love to understand the intention behind it.

hydromt/data_source/__init__.py
```diff
@@ -1,21 +1,19 @@
 """DataSources responsible for validating the DataCatalog."""
 
 # importing all data sources for discovery, factory needs to be imported last.
-from .data_source import DataSource, SourceMetadata  # noqa: I001
+from .data_source import DataSource  # noqa: I001
```
savente93, line 4: don't think the noqa is necessary anymore since we export it in the `__all__`  right?

tests/data_adapter/test_data_adapter.py
```diff
@@ -606,6 +606,7 @@ def test_dataframe_time(df_time, tmpdir):
 
 
 @pytest.mark.skip(reason="Needs implementation of all raster Drivers.")
+# FIXME: Partially tested in "test_rasterio_driver", just missing the cache=True option.
```
savente93, line 609: issue? :)

tests/drivers/geodataset/test_vector_driver.py
```diff
@@ -64,7 +64,7 @@ def example_vector_geods(self, geodf, tmp_dir: Path) -> Path:
 
     def test_read(self, geodf, example_vector_geods: Path):
         res = GeoDatasetVectorDriver(metadata_resolver=ConventionResolver()).read(
-            str(example_vector_geods)
+            str(example_vector_geods),
```
savente93, line 67: this seems accidental. Is it?

savente93 (commented):
I haven't really read into slippy tiles or map sctacks so I'm not totally sure what they do, but the tests look good, and I saw nothing weird so I'm good with it. THere are just a few small comments, but once those are resolved I'll be happy to approve

hydromt/metadata_resolver/slippy_map_tile_resolver.py
Jaapel: I was working on getting the slippy map tiling to work with the new architecture, but I found out while working that we decided to postpone to beta. So I already did some work, did not want to toss it.

hydromt/metadata.py
Jaapel: I can move it. It should have it's own file I think

hydromt/config.py
Jaapel: I think that certain things is HydroMT are settings. Pydantic Settings are easily settable using environment variables, so that is why I chose it. I think more settings belong here, definitively when HydroMT starts to do more (multiprocessing, secrets, etc.)

savente93: That makes sense, in that case we can keep it

hydromt/data_source/__init__.py
Jaapel: noqa is needed still, as the order of imports is important

hydromt/metadata.py
savente93: It's not a very big deal, so I'll leave it up to you

savente93 (approved):
well done!

PR #933, opened 2024-05-22
bugfix in hydrograph_1d method

DirkEilander:
## Issue addressed
An index error was raised when the number of peaks varies between stations. This could occur in real data and should be dealt with in the code.

## Explanation
Limit the maximum index per station to the number of peaks.

## General Checklist
- [x] Updated documentation
- [x] Updated changelog.rst


hboisgon (approved):
All good thanks for fixing this @DirkEilander !
It looks like the windows tests are failing for a different reason but we should look into that before we release...

PR #950, opened 2024-05-23
restructure V1 migration guide

DirkEilander:
## Explanation
I tried to group the items in the migration guide and order these in a logical way. We now have four sections. 

```md
## Package API
## Model
## DataCatalog
## Plugins
```

## General Checklist
- [ ] Updated documentation
- [ ] Updated changelog.rst

## TODO
- [x] carefully read through and check on naming consistency


DirkEilander:
Full table of content:

```markdown
# Migrating to v1

## Package API

## Model

### Moving from an inheritance to composition structure

### Implementing Model Components

#### Model region and geo-spatial components
#### GridComponent
#### VectorComponent
#### MeshComponent
#### TablesComponent
#### GeomsComponent
#### DatasetsComponent and SpatialDatasetsComponent
#### ConfigComponent

### Removed Model attributes

### Model configuration files

#### Removing support for `ini` and `toml`
#### Changes to the `yaml` file format

## DataCatalog

### Split the responsibilities of the `DataAdapter` into separate classes

#### DataSource
#### MetaDataResolver
#### Driver
#### DataAdapter

### Changes to the data catalog `yaml` file format

### Removing dictionary-like features for the DataCatalog

## Plugins
```

savente93:
Couple of thoughts to start off: 

- I'm not sure that every individual component needs it's own section, as they will be very similar, though if you feel you have things to say there it can stay. 
- I would actually put the yaml changes more up front as most users will be interested in that I presume. Not sure if the order was part of the proposal, but I feel like changes that basic users will care most about get buried a little bit. 
- We talked about wanting to split documentation up into sections for "lay" users, power/advanced users, and (plugin) developers. Do we want to start that trend here? 

Otherwise it looks generally okay. I'll add more things here if I think of them

Jaapel:
Looks good! If we want to do an excellent job, I would indeed specify changes from CLI -> High-Level DataCatalog API -> Plugin Devs, but this is better than what we had before.

Jaapel (approved):

DirkEilander:
I have applied your suggestions @Jaapel and @savente93.

docs/dev/migrating-to-v1.rst
```diff
@@ -1,79 +1,96 @@
...
-but given that YAML files are very model specific we've decided to make this available through the YAML file as well.
-- `global` is intended for any configuration for the model object itself, here you may override any default
-configuration for the components provided by your implementation. Any options mentioned here will be passed to the `Model.__init__` function
-- `steps` should contain a list of function calls. In pre-v1 versions this used to be a dictionary, but now it has become a list
-which removes the necessity for adding numbers to the end of function calls of the same name. You may prefix a component name
-for the step in a dotted manner to indicate the function should be called on that component instead of the model. In general any step
-listed here will correspond to a function on either the model or one of its components. Any keys that are listed under a step will be
-provided to the function call as arguments.
+
+- `modeltype` (optional) details what kind of model is going to be used in the model.
```
DirkEilander, line 82: @hboisgon @deltamarnix In my opinion `model_name` would be better. A type refers to a category of models, but here you want the actual model software name right?

```suggestion
- `model_name` (optional) the name of the model software.
```

DirkEilander: In addition, I think it would be nice to have a `Model.from_yml` class method to move the CLI functionality to the model class. In that way also python users can benefit from the yaml format. Something for v1-beta?

PR #951, opened 2024-05-24
Fix warnings

savente93:
## Explanation
Apparently pydantic doesn't like you overriding the class vars of parents. As I think it is important to stay warning free, I added a pixi test job and a (non-required) ci job to make sure that the test suite remains warning free.  We can also always allow/ignore warnings that we think are acceptable in the `pyproject.toml` 

## General Checklist
- [ ] Updated documentation
- [ ] Updated changelog.rst

### DIFF
--- a/.github/workflows/test_error_free.yml
+++ b/.github/workflows/test_error_free.yml
@@ -1,0 +1,61 @@
+name: Tests
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - v1
+    paths:
+      - tests/**/*
+      - hydromt/**/*
+      - data/**/*
+      - pyproject.toml
+      - pixi.lock
+      - pixi.toml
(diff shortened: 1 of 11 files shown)


Jaapel:
Regarding the pydantic thing, can we maybe change the class var annotation or update pydantic? It should be possible to inherit class vars according to: https://github.com/pydantic/pydantic/issues/2061

savente93:
Nice I didn't see that one, I'll see if that works

Jaapel:
Where do we use `supports_writing`? Otherwise try except NotImplementedError is also a valid way to check whether the driver supports writing

savente93:
That's true, I just thought it was a bit nicer of a way to document it, but I guess we can also just remove it if you're okay with that.

Jaapel (approved):
I approve of my own changes 👍

PR #971, opened 2024-05-31
restructure plugins

### LINKED ISSUE #958: expose class rather than submodule with plugin entrypoints
### Kind of request

### Enhancement Description

In v0.x we used the entrypoints to expose a class. In v1.x entrypoints expose a submodule. 

This has two disadvantages that require (large) changes from user and plugin that are not necessary in my opinion: 
- this poses strict requirements on a plugin structure, i.e. the model class must be in a submodule and should be the only thing exposed in that submodule. This is currently not the case for any plugin and also poses limitations on the file structure of the core.
- the name of the class is used as the name of the entrypoint which is used in our CLI, i.e. before the WflowModel class was know as the "wflow" entrypoint and hence the build command was `hydromt build wflow ...`. Now the entrypoint name is unused and the name becomes "WflowModel", hence the command is `hydromt build WflowModel ...`. I would like to keep the former for consistency and simplicity.

I couldn't find any discussion on the new implementation or big advantages,. One advantage I see is that that it is shorter in the toml file.

Unless there are other major advantages, I suggest to expose the class itself again with entrypoints like we did in v0.x.

### Use case

hydromt CLI

**v0.x (entrypoint exposed a class; this is how I suggest it functions)**

```toml
[project.entry-points."hydromt.models"]
wflow = "hydromt_wflow.wflow.WflowModel" # class in wflow.py
```

```bash
hydromt build wflow ...
```

**v1.x (entrypoint exposes a submodule; new implementation)**

```toml
[project.entry-points."hydromt.models"]
wflow = "hydromt_wflow.wflow"  # this is a script and not a subclass in hydromt_wflow; hence it wouldn't work in the current approach if I understand correctly
```

```bash	
hydromt build WflowModel ..
```

### Additional Context

We can slightly modify the v0.x `get_plugin_eps` function to update the v1 `_discover_plugins` function, see https://github.com/Deltares/hydromt/blob/ad2548e43f30b84488e55a4afe3414a0d1f7a87d/hydromt/models/model_plugins.py#L47

### LINKED ISSUE #960: create entrypoints for predefined catalogs
### Kind of request

Adding new functionality

### Enhancement Description

Currently plugins.py has entrypoints defined for Driver, Model and ModelComponent. 
To allow plugins to define their own catalogs, we can define entrypoints for catalogs as well.
In predefined_catalogs.py we define a list of catalogs that are available in the core package.

### Use case

Plugin developers can also maintain and expose their predefined catalogs to users

### Additional Context

This is pending the discussion in #958  about whether we should have entrypoints per submodule or per catalog class.

DirkEilander:
## Issue addressed
Fixes #958 
Fixes #960 

## Explanation
The entrypoints should now expose a submodule/script with a `__hydromt_eps__`. 
Since this is more specific than the python __all__ variable it poses less restrictions on the package structure.
The `__hydromt_eps__` should be a list of strings with names of classes that are exposed. 
The name of each class is first taken from the `name` attribute and otherwise the class name itself from `__name__`.
Note that both `Driver` and `Model` have a `name` attribute, `ModelComponent` does not (yet) have this attribute.
In addition I added a check to make sure the exposed classes are nonabstract subclasses of the base class that is associated with each entrypoint group.
Each entrypoint group now has it's own class to avoid code duplication.

## General Checklist
- [ ] Updated documentation
- [x] Updated changelog.rst

## Data/Catalog checklist
- [ ] None of the old `data_catalog.yml` files have been chagned


DirkEilander:
The migration guide has been updated in #950 to avoid merge conflicts

savente93 (approved):
Tests pass and are updated so LGTM

Issue #966, opened 2024-05-29
new `Model.from_yml` method


DirkEilander:
### Kind of request

None

### Enhancement Description

move build and update cli functionality to a Model method 

### Use case

- the cli should be minimally thin wrapper
- python users can more easily use the yaml files

Jaapel:
Perhaps useful to take into account when implementing: https://github.com/Deltares/hydromt/issues/872.
import plugings: https://github.com/Deltares/hydromt/blob/v1/hydromt/drivers/base_driver.py#L92
Scan subclasses and filter by class variable "name": https://github.com/Deltares/hydromt/blob/v1/hydromt/drivers/base_driver.py#L95

savente93:
currently fixed in main

Issue #886, opened 2024-04-15
Implement `DataCatalog` export with new class model.


Jaapel:
### Kind of request

Changing existing functionality

### Enhancement Description

With the new classes, the `DataCatalog`s export functionality needs to be implemented.

### Use case

Export HydroMT data.

hboisgon:
The command line interface to it is new but it's as old as the artifact data catalog :)
But indeed I think for alpha what is important is to see if the reshuffle works with plugins, not yet with the end/CLI users so if it's too long to implement then can be pushed to beta.

Jaapel:
We did change the `DataCatalog`s depending classes, so everywhere we currently interact with a `DataAdapter` should change to `DataSource` and underlying classes.

Jaapel:
Blocked by: https://github.com/Deltares/hydromt/issues/896

savente93:
fixed in #1052

PR #988, opened 2024-06-07
Cli integration tests

### LINKED ISSUE #943: Enable integration tests in `tests/cli/test_cli.py`
### Kind of request

None

### Enhancement Description

enable integration tests marked as skip or commented out in `tests/cli/test_cli.py` or add a further explanation why they can't be.

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #943 

## Explanation
I've decided to leave most of the cli code as is even though we wanted to reduce it, but the code for it works again. I've removed the tests that were purely about region arguments since we've decided to remove those. I haven't actually updated the code there yet, but I'll pick that up in #966 when the time comes, since I think that is closely related. 

I also can't quite remember why I thought that I'd needed a `from_dict` or `from_yml` function for this one, but given that I've written it and we plan to add it later I think we might as well just add it. I'll do more testing and integration in the ticket mentioned above. 

## General Checklist
- [ ] Updated documentation
- [ ] Updated changelog.rst


Jaapel (commented):
If I understand correctly, in the current component architecture, you would now not define your region using the `-r` argument, but you would add come `RegionComponent` to your model configuration, correct?

tests/cli/test_cli.py
```diff
@@ -180,25 +147,6 @@ def test_export_cli_no_data_ignore(tmpdir):
...
-    with pytest.raises(NotImplementedError):
-        _ = CliRunner().invoke(
-            hydromt_cli,
-            [
-                "export",
-                str(tmpdir),
-                "-s",
-                "hydro_lakes",
-                "-r",
-                "{'subbasin': [-7.24, 62.09], 'uparea': 50}",
```
Jaapel, line 194: How would we export for a certain region? I `hydromt export` should still be tested right?

savente93: The reason (as far as I understand) that the region got removed from the cli, is that the way we construct regions has become much more complicated and interwoven with the components. This means that regions can no longer be specified independent of what type of components you have. The `export` can be used without a region and then all the data will be exported. aside from that we also have a file format for specifying what data needs to get exported. I'm not sure if this is updated yet, but regions will have to be specified there. There should be at least one test that uses the file format.

savente93:
Most of the sonar clouds errors are introduced by touching code that isn't actually in the scope of this PR (i.e. refactoring `export_data`) so I think it's fine to leave it as is.

hydromt/data_catalog/data_catalog.py
```diff
@@ -1006,10 +1006,11 @@ def export_data(
         self,
         data_root: Union[Path, str],
         bbox: Optional[Bbox] = None,
-        time_tuple: Optional[TimeRange] = None,
+        time_range: Optional[TimeRange] = None,
```
Jaapel, line 1009: Seems like you partially updated the `export_data` function to v1, but not fully?

tests/cli/test_cli.py
```diff
@@ -56,51 +56,23 @@ def test_cli_clip_help():
...
-        "-r",
-        "{'bbox': [12.05,45.30,12.85,45.65]}",
         "--opt",
         "setup_grid.res=0.05",
         "-vv",
     ]
     _ = CliRunner().invoke(hydromt_cli, cmd)
 
 
-@pytest.mark.skip(
```
Jaapel, line 79: Why are the override tests skipped?

Jaapel:
Also SonarCloud makes a good point, would correct that one.

hydromt/data_catalog/data_catalog.py
savente93: Yeah, that's true, these are some left over changes from before I decided to leave the export_data function alone, I can remove them if you like

savente93:
You mean `export_data`?

tests/cli/test_cli.py
savente93: I'd have to write another yaml file for that since we don't have a way to specify model components from the CLI. I can do that if you like

savente93:
because that would mean basically fixing #886 as well in this pr

Jaapel:
> because that would mean basically fixing #886 as well in this pr

Yeah no need, just making sure I understand the changes

hydromt/data_catalog/data_catalog.py
Jaapel: no need

tests/cli/test_cli.py
Jaapel: Does the same `--fo` flag exist now?

savente93: yes

savente93:
There are still one or two issues that I can't figure out yet, I'll come back to those in an hour or so, I think I need to look at something else for a bit to recover my productivity

savente93:
Override test now works again, with one small addition. I couldn't figure out why the convention resolver couldn't find the `vito` file for the re class in the example, so I just edited that bit out. I'll have to add that bit back in at some point, but I don't think it's necessary for this PR, and belongs more in a PR to update the examples.

Jaapel:
In `ModelRoot` and in multiple functions in `cli.main` a `FileHandler` for logging is created, pointing to the same file. I took a few stabs at it, but I think solving it would require redoing logging in HydroMT, just relying on the logging hierarchy in the builtin `logging` module, instead of passing loggers around manually.

Jaapel (approved):

PR #992, opened 2024-06-12
data adapter docs

Jaapel:
## Explanation
Added docs on the `transform` and `slice_data` methods


Tjalling-dejong (approved):
Tests are failing, but looks like it is not due to the changes in this PR. So LGTM 👍

PR #1017, opened 2024-07-01
Update Naming conventions

### LINKED ISSUE #821: The `fn` naming convention should be revisited.
### Kind of request

### Enhancement Description

Currently `fn` is used throughout hydromt to mean `filename` however this comes with two problems:
1. `fn` is a generally accepted abrivation for `function` in other communities so it will be confusing for them
2. When something is called `fn` it can be very frequently something than a filename including but not limited to: a data source, a path relative to the root or several kinds of actual data. 

I would like to revisit this convention and come up with a better standard name for this kind of object. 

### Use case

_No response_

### Additional Context

_No response_

### LINKED ISSUE #834: Tracking issue for integrating and harmonising Model and components
### Kind of request

### Enhancement Description

# Requirements:
- Model needs to be able to add components etc. 
- #813 should be solved to be able to create grid components
- #821 so we can harmonize the components
- CLI/yaml users need a way to list the needed components
- Should user be able to determine model/target CRS at init? (#827 ) 
- What are the abstract methods of `ModelComponent` ? 

### Use case

_No response_

### Additional Context

```yaml
- init_model
	components:
		- grid
			kind: GridComponent
		- subgrid:
			kind: GridComponent
		- config:
			kind: ConfigComponent
		- geoms:
			kind: RegionComponent
```

savente93:
## Issue addressed
Fixes #821 
Fixes #834 

## Explanation
I think I've updated everything discussed in #862 this pr mostly updates `time_tuple -> time_range` and `fn -> path` or `fn -> data` as appropriate. I think with that we have fixed all the non data convention issues that we've discussed but feel freet o point out any I've missed. 

## General Checklist
- [ ] Updated documentation
- [x] Updated changelog.rst

## Data/Catalog checklist
- [ ] None of the old `data_catalog.yml` files have been chagned


Jaapel (approved):
LGTM 👍

PR #1019, opened 2024-07-03
Release flow

### LINKED ISSUE #972: Migrate release flow to pixi
### Kind of request

None

### Enhancement Description

To my dismay I found out that we haven't updated our release workflow when we moved to pixi. I'll do the release more "manually" now because I still want to release today, but the workflow should be updated. 

### Use case

_No response_

### Additional Context

_No response_

savente93:
## Issue addressed
Fixes #972 

## Explanation
I originally phrased this issue as "migrate release flow to pixi" because I was under the impression that pixi had a `publish` command. it does have a `upload` command for conda, which I still want to investigate. Perhaps this can be a more efficient way of dealing with the feedstock but given that this still needs to be investigated I decided to leave it out of scope for this one. Currently the release flow happens entirely within github actions (which has some limitations that I'll get to in a second) and I could move these to the pixi tasks, depending on what people prefer. The upside to doing it within github actions is that it generalises easier to projects which do now use pixi yet, but whether that trade off is worth it I'll leave up to the reviewer. 

With the changes in this PR the release flow will go as follows: 
1. Use the `prepare release` workflow on github. This will: 
    a. Create a release branch with the correct new version number
    b. update the `hydromt/__init__.py` and `docs/changelog.rst` with the new version and headers as necessary 
    c. Create a PR with these changes
 2. On every commit to the Release PR it will create test and upload
     a. docs
     b. docker image (incl one for binder integration)
     c. pypi package
     (d. conda package is excluded here because that has to happen externally through the feedstock) 
 3. Due to the fact that workflows cannot run other workflows (enforced by github at the platform level) a (possibly) empty commit will be necessary to start the artifact creation the first time. The github bot will post a comment explaining this and how to do it. Do not worry about the empty commit, the PR will be squash merged like any other at the end.  
 3. Each of  the release artifacts created can be downloaded and inspected manually if desired. The github bot will post links to the artifacts in a comment if they are successfully created
 4. Once everything looks okay, someone should approve the PR. Once this is done, a workflow will be started that will merge the PR, download all the previously created release artifacts and publish them (instead of regenerating them so you know the ones you inspected are the ones being released), in addition to creating a github release. Once all this is done it will create a commit on main, adding a new `unreleased` header in `docs/chagelog.rst` and setting the version in `hydromt/__init__.py` back to a dev version. 
 5. your release is now complete. 

Hope this was a good enough explanation, I know it's a bit hairy on the back end, but it makes the process of actually creating a release much more automated and inspect able I think. It's a very "set it and forget it" setup which I like. Let me know if any additional clarification is required 

## General Checklist
- [x] Updated documentation
- [x] Updated changelog.rst


savente93:
once this has had an initial review and people are happy with the approach I'll also update the docs with this explanation, but I wanted to get some feedback on the idea first.

savente93:
Oh, also, by it's very nature, this is pr is very hard to test, but I have developed this using this repo: https://github.com/savente93/ci-playground There things can be inspected and tested as much as I could figure out, so if reviewer wants to test/play with it first feel free to do it there.

Jaapel:
Seems like you can trigger workflows from workflows? Max 3 it seems in the notes. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

.github/workflows/prep-release.yml
```diff
@@ -0,0 +1,151 @@
...
+
+          set -e
+          # check if there already is a release PR open
+          # parsing of current version
+          export CURRENT_VERSION=$(grep "version" hydromt/__init__.py | cut -d= -f 2 | tr -d "\" ")
+          export CURRENT_MAJOR=$(echo $CURRENT_VERSION | cut -d'.' -f 1)
+          export CURRENT_MINOR=$(echo $CURRENT_VERSION | cut -d'.' -f 2)
+          export CURRENT_PATHCH=$(echo $CURRENT_VERSION | cut -d'.' -f 3)
+
+          BUMP="minor"
```
Jaapel, line 60: Seems like `BUMP` is always `"minor"`?

Jaapel:
I also do not understand your note about publishing to docker and pypi on every push to a release branch. Pypi at least does not support removing artifacts, I can't remember if testpypi does. But there should be some cleanup if you push a new version of the artifact.
If pixi can easily fix pushing artifacts to conda, I would happily support that. If not, then it is fine to include it in GitHub. Is there a GitHub action for it? The argument of sharing this with people is stronger if we create our own GH Action for it, as other developers will for sure want this as well.

savente93:
I think pixi might be able to also take on the release to conda which would mean we don't have to wait for the feedstock to wake up which would be nice, but this probably requires more research and development so I'm leaving this off for now. I've tested it as well as possible, but given that we can only really test it by actually releasing, it will have to be made more robust over time. I did add as much detail as I thought was appropriate to the docs, and the workflows are (hopefully) fairly well documented, so I think it's good enough for now.

.github/workflows/finish-release.yml
```diff
@@ -0,0 +1,126 @@
+
+name: Finalise a new release
+
+on:
+  pull_request:
+    types:
+    - closed
```
Jaapel, line 7: Should this not also include some filter on the release branch naming convention? Now this will also trigger on any PR being closed if I understand correctly.

.github/workflows/finish-release.yml
```diff
@@ -0,0 +1,126 @@
...
+              echo "detection result: $RELEASE_VERSION"
+              exit 1 # somthing went very wrong
+            fi
+            ## seting up git
+            git config user.name "GitHub Actions Bot"
+            git config user.email "<>"
+            git tag "v$RELEASE_VERSION"
+            # post release stuff so we don't forget
+            sed -i "s/v$RELEASE_VERSION/Unreleased\n==========\n\nNew\n---\n\nChanged\n-------\n\nFixed\n-----\n\nDeprecated\n----------\n\nv$RELEASE_VERSION/" docs/changelog.rst
+            sed -i 's/__version__.*=.*"\(.*\)".*/__version__ = "\1.dev0"/' hydromt/__init__.py
```
Jaapel, line 113: `__version__` seems hardcoded here?

.github/workflows/prep-release.yml
```diff
@@ -0,0 +1,138 @@
...
+          # [{\"name\":\"v$NEW_VERSION\",\"version\":\"$NEW_VERSION\",\"url\":\"https://deltares.github.io/hydromt/v$NEW_VERSION/\"}]
+          # an array with the new entry we want to add
+
+          # sort_by(.version|split(\".\")|map(tonumber))
+          # take the array, split the version field by the "." char and make numbers out of the components. then sort the whole array by these numbers
+          # this is why we had to remove the latest field earlier,
+          # otherwise the number conversion would fail here.
+
+          #| . + [{\"name\":\"latest\",\"version\":\"latest\",\"url\":\"https://deltares.github.io/hydromt/latest\"}]
+          # add teh latest field back in at the end
```
Jaapel, line 110:
```suggestion
          # add the latest field back in at the end
```

.github/workflows/prep-release.yml
```diff
@@ -0,0 +1,138 @@
...
+          	echo "invalid bump: $BUMP"
+          	exit 1
+          	;;
+          esac
+
+          ## seting up git
+          git config user.name "GitHub Actions Bot"
+          git config user.email "<>"
+
+          # # create release branch
```
Jaapel, line 79:
```suggestion
          ## create release branch
```

docs/dev/contributing.rst
```diff
@@ -422,18 +422,11 @@ A single test:
...
+1. Go to the `actions` tab on Github, select `Create a release` from the actions listen to the left, then use the `run workflow` button to start the release process. You will be asked whether it will be a `major`, `minor` or `patch` release. Choose the appropriate action.
+2. The action you just run will open a new PR for you with a new branch named `release/v<NEW_VERSION>`. (the `NEW_VERSION` will be calculated for you based on which kind of release you selected.) In the new PR, the changelog, hydromt version and sphinx `switcher.json` will be updated for you. Any changes you made to the `pyproject.toml` since the last release will be posted as a comment in the PR. You will need these during the Conda-forge release if there are any.
+3. Every commit to this new branch will trigger the creation (and testing) of release artifacts. In our case those are: Documentaiton, the PyPi package and docker image (the conda release happens seperately). After the artifacts are created, they will be uploaded to the repository's internal artifact cache. A bot will post links to these created artifacts in the PR which you can use to download and locally inspect them.
```
Jaapel, line 427:
```suggestion
3. Every commit to this new branch will trigger the creation (and testing) of release artifacts. In our case those are: Documentation, the PyPi package and docker image (the conda release happens separately). After the artifacts are created, they will be uploaded to the repository's internal artifact cache. A bot will post links to these created artifacts in the PR which you can use to download and locally inspect them.
```

docs/dev/contributing.rst
```diff
@@ -422,18 +422,11 @@ A single test:
...
+4. When you are happy with the release in the PR, you can simply merge it. We suggest naming the commit something like "Release v<NEW_VERSION>"
+5. After the PR is merged, a action should start (though it will not show up under the PR itself) that will publish the latest artifacts created to their respective platform. After this, a bot will add a final commit to the `main` branch, setting the hydromt version back to a dev version, and adding new headers to the `docs/changelog.rst` for unreleased features. It will also create a tag and a github release for you automatically. The release is now done as far as this repo is concerned.
+6. The newly published PyPi package will trigger a new PR to the `HydroMT feedstock repos of conda-forge <https://github.com/conda-forge/hydromt-feedstock>`_.
+   Here you can use the comment posted to the release PR to see if the `meta.yml` needs to be updated. Merge the PR to release the new version on conda-forge.
+7. celebrate the new release!
```
Jaapel, line 432: Still looking forward to this :)

Jaapel (changes requested):
Most of it looks good, only the branch matching comment is critical

.github/workflows/finish-release.yml
savente93: I get why you think that but it's not in sed `\1` is the code for "first capture group" so this will get substituted with what was captured in the `(.*)` earlier in the command

docs/dev/contributing.rst
savente93: celebrating the new release?

.github/workflows/finish-release.yml
savente93: oh that's a very good catch! (that's the downside of working in a simplified test environment I guess 😅). Will fix

savente93: I've fixed this now, though thinking about it, if this ran on a random PR, there wouldn't be any artifacts for it to download so it wouldn't have succeeded anyway. I guess that's one upside to doing it the way I have

Jaapel (approved):
👍

Issue #984, opened 2024-06-05
`zoom_level` argument still needed?


Jaapel:
### Kind of request

None

### Enhancement Description

`RasterDatasetAdapter` still uses `zoom_level`: https://github.com/Deltares/hydromt/blob/v1/hydromt/data_adapter/rasterdataset.py#L150

### Use case

Keep the API clean and stable before v1 release.

DirkEilander:
zoom_level is needed as an argument of the `DataCatalog.get_rasterdataset` method and should be passed to the Driver. I'm not sure it is required in the `RasterDatasetAdapter`

Jaapel:
Fixed in https://github.com/Deltares/hydromt/pull/1024

Issue #1023, opened 2024-07-08
Seperate URIResolver and Driver


Jaapel:
### Kind of request

Changing existing functionality

### Enhancement Description

The `URIResolver` and `Driver` classes were initially clumped together, as they were deemed to be "closely coupled". However, now every Abstract Driver type puts the two together in the same way: https://github.com/Deltares/hydromt/blob/233f95c4788ded2bad17d65759a68c207f74ab1b/hydromt/drivers/geodataframe/geodataframe_driver.py#L19-L55

I suggest to nicely seperate them, so that we can implement that behaviour in one place. This also opens better implementation of other features, such as zoom levels being handled at the `DataSource` level.

### Use case

Easier maintenance, better seperation of concerns, better reusability over `Driver`s and `URIResolver`s.

PR #1024, opened 2024-07-08
875/slippy map resolver

### LINKED ISSUE #875: New `Driver` to read slippy map tile data

Jaapel:
## Issue addressed
Fixes #875 
Fixes https://github.com/Deltares/hydromt/issues/984

Also:
- renamed `zoom_level` to either:
  - `zoom` when referring to either an overview level, or a combination of a resolution + unit
  - `overview_level` when only referring to an overview_level

## Explanation
Implemented zoom level in the `ConventionResolver` and the `RasterioDriver`.
I still think we should keep resolvers and driver apart (https://github.com/Deltares/hydromt/issues/1023), then we can implement this much cleaner in the `RasterDataSource`. Now I had to add `zls_dict` to metadata, pass metadata to `URIResolver.resolve` and do a lot of try/excepts to make it happen

## General Checklist
- [x] Updated documentation
- [ ] Updated changelog.rst


savente93:
does this also resolve #984 ?

Jaapel:
> does this also resolve #984 ?

Yes!

hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -72,24 +75,42 @@ def read_data(
...
+        # Then we can implement looking for a overview level in the driver.
+        def _open() -> Union[xr.DataArray, xr.Dataset]:
+            try:
+                return open_mfraster(uris, **kwargs)
+            except rasterio.errors.RasterioIOError as e:
+                if "Cannot open overview level" in str(e):
+                    kwargs.pop("overview_level")
+                    return open_mfraster(uris, **kwargs)
+                else:
+                    raise
```
savente93, line 111: should this be `raise e` ?

savente93 (approved):
can't really attest the logic of the map operations, but the code and tests look good to me, so I think this is fine

hydromt/data_catalog/drivers/raster/rasterio_driver.py
Jaapel: [You can use both: https://stackoverflow.com/questions/36153805/difference-between-raise-and-raise-e]

savente93: cool TIL

PR #1029, opened 2024-07-18
Seperate resolver from driver

### LINKED ISSUE #1023: Seperate URIResolver and Driver

Jaapel:
## Issue addressed
Fixes #1023

## Explanation
`URIResolver` and `Driver` were tightly coupled, as we designed it when v1 was first discussed. When implementing, we found out that this was not needed, and even worked around it by always coupling them in the same way on the `Driver` level. In this update I split them again, and manage their connection on the `DataSource` level.

More specifically:
the `BaseDriver` depended on the main abstract function `get_data`, which then called `URIResolver.resolve` followed by `self.read` and returned the results. Now the full chains lives in `DataSource`: 
- `URIResolver.resolve` -> `uris: List[str]`
- `BaseDriver.read(uris, **kwargs)` -> data
- `DataAdapter.transform(data, **kwargs)` -> data

This update also contains:
- Plugin behaviour for `URIResolver`: you can now use the `hydromt.uri_resolver` entrypoint to expose your own resolver (STAC?)
- A fix for serializing filesystems, as previously, only the protocol was serialized. Now we use the built-in `fsspec.AbstractFileSystem.to_dict` serializer (`password=False`) of course.

## General Checklist
- [ ] Branch is up to date with `v1`
- [ ] Updated documentation
- [ ] Updated changelog.rst

Updated the data catalogs, but again: v1 alpha, no need to bump versions


savente93 (approved):
Very nice Jaap, I think this looks a lot better organised

PR #1039, opened 2024-09-05
Docs/data sources

Jaapel:
## Issue addressed
Fixes part of the documentation

## Explanation
- Introduced doctest to the data types documentation.
- Updated `options` to be a property of the `URIResolver` instead of passing this at runtime. This allows for specifying certain resolver-specific options in the data catalog.
- Fixed a bug where `Dataset` was not a known data type. Apparently this was never tested, which is concerning.
- Updated some (quite outdated, even on v0) docstrings in the `DataCatalog` methods.


savente93 (approved):
LGMT

savente93 (changes requested):
looks good but tests should be fixed first

savente93 (approved):
All good now

PR #1042, opened 2024-09-11
Docs/hydromt python

Jaapel:
## Issue addressed
Fixes part of the v1 docs

# Explanation
Shifted the docs around what the module structure now looks like.
Some references are still broken, will fix those in a different docs PR first.

### DIFF
--- a/docs/guides/advanced_user/hydromt_python.rst
+++ b/docs/guides/advanced_user/hydromt_python.rst
@@ -3,11 +3,12 @@
 Using HydroMT in Python
 =======================
 
-As HydroMT's architecture is very modular, it is possible to use HydroMT in Python
-without using the command line interface (CLI). With Python, you actually have access to
-a lot of internal functionalities. This can be useful if you want to for example:
+As HydroMT's architecture is modular, it is possible to use HydroMT as a python library
+without using the command line interface (CLI). Using the libary, you actually have
+access to many internal functionalities. This can be useful if you want to for example:
 
-- build / update / clip models, check configurations or export data in Python instead of the CLI
+- build / update / clip models, check configurations or export data in Python instead of
+  the CLI
 - analyse model inputs or results
 - analyse and compare input data without connecting to a specific model
 - process input data for another reason than building a model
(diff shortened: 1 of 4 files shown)


docs/guides/advanced_user/hydromt_python.rst
```diff
@@ -3,11 +3,12 @@
 Using HydroMT in Python
 =======================
 
-As HydroMT's architecture is very modular, it is possible to use HydroMT in Python
-without using the command line interface (CLI). With Python, you actually have access to
-a lot of internal functionalities. This can be useful if you want to for example:
+As HydroMT's architecture is modular, it is possible to use HydroMT as a python library
+without using the command line interface (CLI). Using the libary, you actually have
+access to many internal functionalities. This can be useful if you want to for example:
```
savente93, line 8: Know this is from the previous version, but I'd rather not tell people "they have access to internal functionalities" we can say they have access to ones not necessarily available from the cli/yml but I think we should communicate something more along the lines of "internal means hands off or you're on your own"

docs/guides/advanced_user/hydromt_python.rst
```diff
@@ -79,29 +76,60 @@ You can find more detailed examples on using the DataCatalog and DataAdapter in
...
+
+The ``model`` module has a ``processes`` submodule. This module contains some functions
+to work with different kinds of model in- and ouput.
+
+* :ref:`grid <workflows_grid_api>`: generic workflows to prepare regular gridded data.
+    Used with the ``GridModel``. For example to prepare regular grid data from constant,
+    from RasterDataset (with or without reclassification) or from GeoDataFrame.
+* :ref:`mesh <workflows_mesh_api>:` generic workflows to prepare unstructured mesh
+    data. Used with the ``MeshModel``. For example to create a mesh grid or prepare
+    unstructured mesh data from RasterDataset.
```
savente93, line 124: We don't have `GridModel` or `MeshModel` anymore, just Models which may or may not have grid or mesh models

savente93 (commented):
two small nitpicks but otherwise it's looking good!

savente93 (approved):
nice!

PR #1052, opened 2024-09-17
Update export catalog

### LINKED ISSUE #886: Implement `DataCatalog` export with new class model.

Jaapel:
## Issue addressed
Fixes #886 

## Explanation
I moved the `DataCatalog.export_data` to the new class structure.
Next to using the correct classes, I had to:
- Fixed a bug in the migrate_datacatalog script to make sure that the `tileindex` `option` ended up in the resolver
- Changed the exported uri to something sensible if the format changes during export. This behaviour is included in the driver.
- Changed some catalog references to match the 0.9 and 1.0 references. I wonder how that ever worked in v0.

## General Checklist
- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist


savente93:
I'll give it an early review tomorrow

examples/data/geodataset_catalog.yml
```diff
@@ -1,4 +1,6 @@
 ---
+meta:
+  root: examples/data
```
savente93, line 3: so this is the old way of registering a root, now the format should be: 
```yml
meta:
    roots:
        - examples/data
```

I'm not sure we ever discussed whether we want to deprecate the old format or not. What do you think?

savente93 (commented):
I think this is the best implementation we have time/resources for now. Aside from the one comment about keeping the old root format in the yaml I think this is okay. Once we resolve that I'll be happy to approve

examples/data/geodataset_catalog.yml
Jaapel: These are still in the current data catalogs. We can decide to deprecate in another PR

savente93 (approved):
lets go

PR #1053, opened 2024-09-19
Update most examples for V1

savente93:
This PR updates the examples for the V1 release, and also fixes a few bugs along the way.


hydromt/_io/readers.py
Jaapel, line 1: Not in scope for this PR, but reading the `data_path` will never work with cloud files

Jaapel (approved):

PR #1059, opened 2024-09-24
Finish advanced user guide

savente93:
I added documentation showing how to use the components in more depth. There were files for `model_config` and `model_processes` but I actually have no idea what was supposed to go in those, so I just deleted them. I propose to just add those when people ask questions because then it will be clearer what should go in them.


Jaapel (approved):
Agree with the deletes, why document everything double...

PR #1070, opened 2024-09-27
fix catalog reference to correct hash

Jaapel:
## Issue addressed
Reference to main in old catalog format


savente93 (approved):
LGTM

veenstrajelmer:
Unfortunately, this change causes two tests to fail in the main branch. I have created a follow-up issue here: https://github.com/Deltares/hydromt/issues/1093

Issue #1121, opened 2025-02-20
Unexpected behaviour in raster.full_from_transform


roeldegoede:
### HydroMT version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

According to the documentation of raster.full_from_transform, the shape is defined as "Length along (dim0, x, y) dimensions, of which the first is optional" whereas it returns a DataArray with dimensions y, x (hence flipped).

When calling this function like this, where the mmax is nr of cells in x-direction, and nmax the nr of cells in y-direction, like is done in `hydromt.model.processes.grid.create_grid_from_region`.

```
raster.full_from_transform(
        transform,
        shape=(mmax, nmax),
        nodata=1,
        dtype=np.uint8,
        name="mask",
        attrs={},
        crs=geom.crs,
        lazy=False,
    )
```

### Current behaviour

This returns a raster with dimensions mmax by nmax (hence x, y) and not y, x!

### Desired behaviour

To properly work, we should change the code into:

```
raster.full_from_transform(
        transform,
        shape=(nmax, mmax),
        nodata=1,
        dtype=np.uint8,
        name="mask",
        attrs={},
        crs=geom.crs,
        lazy=False,
    )
```

and change the description of full_from_transform in raster.py

Issue #1120, opened 2025-02-20
Error in parsing crs="utm" in hydromt.model.processes.grid.create_grid_from_region


roeldegoede:
### HydroMT version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

Providing a region of kind "geom" with crs="utm" in `hydromt.model.processes.grid.create_grid_from_region` does not work.

```
import geopandas as gpd
import pandas as pd
import numpy as np
from shapely.geometry import Polygon
import hydromt

# Generate a random polygon in a reasonable UTM 33N range
x_min, x_max = 300000, 700000  # Easting range (meters)
y_min, y_max = 5000000, 6000000  # Northing range (meters)

# Create a simple quadrilateral polygon with random points
coords = [
    (np.random.uniform(x_min, x_max), np.random.uniform(y_min, y_max)),
    (np.random.uniform(x_min, x_max), np.random.uniform(y_min, y_max)),
    (np.random.uniform(x_min, x_max), np.random.uniform(y_min, y_max)),
    (np.random.uniform(x_min, x_max), np.random.uniform(y_min, y_max)),
    (np.random.uniform(x_min, x_max), np.random.uniform(y_min, y_max))  # Close the polygon
]

polygon = Polygon(coords)

# Create a GeoDataFrame
region= gpd.GeoDataFrame(pd.DataFrame({'id': [1]}), geometry=[polygon], crs=32633)

# Now finally call the function
hydromt.model.processes.grid.create_grid_from_region(
    region = {"geom": region},
    crs= "utm",
    res=100,
)
```

### Current behaviour

It tries to find a UTm zone somewhere on another planet, so it throws an error.

```
---------------------------------------------------------------------------
CRSError                                  Traceback (most recent call last)
Cell In[1], [line 26](vscode-notebook-cell:?execution_count=1&line=26)
     [23](vscode-notebook-cell:?execution_count=1&line=23) region= gpd.GeoDataFrame(pd.DataFrame({'id': [1]}), geometry=[polygon], crs=32633)
     [25](vscode-notebook-cell:?execution_count=1&line=25) # Now finally call the function
---> [26](vscode-notebook-cell:?execution_count=1&line=26) hydromt.model.processes.grid.create_grid_from_region(
     [27](vscode-notebook-cell:?execution_count=1&line=27)     region = {"geom": region},
[...]

Issue #1119, opened 2025-02-20
data catalog is not working ...


roeldegoede:
### HydroMT version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

mamba create -n hydromt -c conda-forge hydromt
mamba activate hydromt

or just in any environment with pip/mamba

pip/mamba install hydromt

```
from hydromt.model import Model
mod = Model(data_libs=["artifact_data"])

```

### Current behaviour

**Slightly different errors with pip or mamba install:

MAMBA:**

```
Traceback (most recent call last):
  File "C:\Users\goede_rl\AppData\Local\miniforge3\envs\hydromt\Lib\site-packages\fsspec\registry.py", line 246, in get_filesystem_class
    register_implementation(protocol, _import_class(bit["class"]))
                                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\goede_rl\AppData\Local\miniforge3\envs\hydromt\Lib\site-packages\fsspec\registry.py", line 281, in _import_class
    mod = importlib.import_module(mod)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\goede_rl\AppData\Local\miniforge3\envs\hydromt\Lib\importlib\__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "C:\Users\goede_rl\AppData\Local\miniforge3\envs\hydromt\Lib\site-packages\fsspec\implementations\http.py", line 9, in <module>
    import aiohttp
ModuleNotFoundError: No module named 'aiohttp'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #1122, opened 2025-02-20
potential (very small) fixes for issues #1120 and #1121

roeldegoede:
Code that locally solved the raised issues. Before finalizing, I thought let's share it first to see whethet it makes any sense


savente93:
Thanks for the suggested implementations Roel. I've taken your suggestions and split them into #1154  so that I could add a test and some docs (I didn't want to mess with you fork) #1119 was previously fixed in #1145. As soon as this gets reviewed we'll merge it and then you should be good.

savente93:
I'll be closing this PR as I've merged your solutions across other PRs

Issue #1105, opened 2025-01-07
Release `pyarrow<16` constraint


aufdenkampe:
### Kind of request

Changing existing functionality

### Enhancement Description

The [hydromt/pyproject.toml](https://github.com/Deltares/hydromt/blob/main/pyproject.toml) file specifies:

```yaml
"pyarrow<16",             # prevents incompat with pandas.
```

However, since Pandas [2.1.0 (Aug 30, 2023)](https://pandas.pydata.org/docs/dev/whatsnew/v2.1.0.html#what-s-new-in-2-1-0-aug-30-2023), pyarrow has been tightly integrated with Pandas.  

To take advantage of many recent updates and improvements to Pandas and GeoPandas, especially for IO with Parquet/GeoParquet files and GDAL 3.9 (via pyogrio), it would be very helpful to release your no-longer-necessary version constraint on pyarrow. For examples of some of the benefits of these advances, see https://github.com/WikiWatershed/global-hydrography/issues/1

PR #1130, opened 2025-03-13
Update python versions

### LINKED ISSUE #540: Add support for Python 3.12
### Kind of request

Adding new functionality

### Enhancement Description

On 02/10/2023 python 3.12 was released. To make sure we keep up with the newest developments we should add support and testing for this new version as well.

### Use case

_No response_

### Additional Context

_No response_

deltamarnix:
## Issue addressed
Fixes #540 

## Explanation
Add support for python 3.12 and 3.13

## General Checklist
- [x] Updated documentation
- [x] Updated changelog.rst

## Data/Catalog checklist
- [x] None of the old `data_catalog.yml` files have been chagned
- [x] `data/chagnelog.rst` has been updated

### DIFF
--- a/.github/workflows/create-release-artifacts.yml
+++ b/.github/workflows/create-release-artifacts.yml
@@ -22,11 +22,6 @@ jobs:
         with:
           fetch-depth: 0
 
-      - uses: actions/setup-python@v5
-        name: Install Python
-        with:
-          python-version: '3.13'
-
       - name: generate env
         run: |
           pip install tomli flit twine
--- a/.github/workflows/create-release-artifacts.yml
+++ b/.github/workflows/create-release-artifacts.yml
@@ -25,7 +25,7 @@ jobs:
       - uses: actions/setup-python@v5
         name: Install Python
         with:
-          python-version: '3.11'
+          python-version: '3.13'
 
       - name: generate env
         run: |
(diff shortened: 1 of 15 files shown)


.github/workflows/create-release-artifacts.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - uses: actions/setup-python@v5
         name: Install Python
         with:
-          python-version: '3.11'
```
deltamarnix, line 28: I wonder if we need to install python. Can't we do with the python that is inside of ubuntu?

deltamarnix:
Fixes #1105

.github/workflows/create-release-artifacts.yml
savente93: The docker image that github uses itself it's available anywhere externally as far as I can tell, however, but the community image that `act` uses which is engineerd to be as close to the GHA version as possible does have both python and pip, although the base ubuntu version does not. So it's difficult to say, but I think it should work without the python setup if we don't need a specific version (which I don't think we do)

tl;dr I think it should work without the install python step.

savente93 (approved):
Thanks for this Marnix! good work. I'll leave it up to you if you want to see if you can do without the python install step but aside from that it looks good to me

.github/workflows/create-release-artifacts.yml
deltamarnix: I removed it and we will see if it actually works.

PR #1148, opened 2025-03-27
Make tests run on minimal hydromt installation

### LINKED ISSUE #1147: Make a separate test bench for minimal installed hydromt
### Kind of request

Adding new functionality

### Enhancement Description

We should run a separate set of tests via GitHub actions where only a minimal set of hydromt + test is installed and tested.

```toml
[tool.pixi.environments]
py39-test = { features = ["py39", "test"], solve-group="py39" }
```

We have seen that some dependencies can cause problems when not fully installed. For example in #1145 , where there seemed to be a discrepancy between installing with mamba and installing with pip.

And the tests need to be run in the test matrix with the different python versions.

### Use case

_No response_

### Additional Context

_No response_

deltamarnix:
Simplify the environments
Run tests in minimal installation too
Skip tests that are dependent on optionals

#1147

## Issue addressed

Fixes #1147

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst


savente93 (commented):
Looks good overall but I have two questions: 
1. Given that we've moved almost everything to pypi dependencies and also test minimal verisons, do we still need the pip test run? 
2. I don't think the docker setup QEMU flow does anything on it's own as a seperate job, so if the others don't need it I think we can just remove that

savente93 (approved):
Good job :D

PR #1169, opened 2025-04-04
Fix release workflows

### LINKED ISSUE #1083: Release flow should update symlinks
### Kind of request

Adding new functionality

### Enhancement Description

release workflow curretnly updated the `switcher.json` of the docs, but doesn't update the symlinks

### Use case

automated release flow

### Additional Context

should be very easy with some `ln -s` action

### LINKED ISSUE #1138: Doc publishing workflow does not produce correct structure in docs
I'm basically making this issue to record some context for myself when I return on Monday.

During the release of v1.0.1 I noticed that the docs weren't being published correctly, ans the deploy stage seemed to be failing. After some fiddling I found out this was because the folder structure in the gh-pages branch wasn't correct. Probably the building docs flow will need to publish to a different dir. For my future reference, here is a screen shot of what the layout should look like in a sucessful docs deployment 

Image

I will look into this and hopefully fix the workflows.

savente93:
## Issue addressed

Fixes #1138 
Fixes #1083 

## Explanation
This PR addresses several issues: 
- the docs were being put in the place because I copy pasted from the other docs flow but forgot the calculation of the `DOC_VERSION` variable. this is now fixed
- In addition the release flow also updates the symlink in the gh-pages so we don't forget. I _think_ the `switcher.json` is updated by the deployment workflow, but I'm not 100% sure. If that turns out to be necessary I'll do that in a separate PR
- Last time the final creation of the release failed because the workflow tried to push to main which is protected. To remedy this I have opted to just create a PR which has to be manually merged, but it still takes care of the most boring and forgettable parts 

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist


.github/workflows/finish-release.yml
```diff
@@ -87,6 +85,23 @@ jobs:
...
+    - name: Determine release version
+      run: |
+        export CURRENT_VERSION=$(grep "__version__" hydromt/__init__.py | cut -d= -f 2 | tr -d "\" ")
+        echo "DOC_VERSION=v$CURRENT_VERSION" >> "$GITHUB_ENV"
+    - name: update doc symlinks
+      run: |
+        git config user.name "GitHub Actions Bot"
+        git config user.email "<>"
+        git pull
+        git checkout gh-pages
```
deltamarnix, line 98: Why not perform a checkout action instead?

.github/workflows/finish-release.yml
```diff
@@ -109,7 +124,7 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Finalise PR
+      - name: Finalise Release
         run: |
             git pull
             git checkout main
```
deltamarnix, line 130: Is this needed? actions/checkout@v4 is already run before

.github/workflows/finish-release.yml
```diff
@@ -123,6 +138,14 @@ jobs:
             git config user.name "GitHub Actions Bot"
             git config user.email "<>"
             git tag "v$RELEASE_VERSION"
+            git push origin tag "v$RELEASE_VERSION"
+            gh release create "v$RELEASE_VERSION" -F release_notes.rst -t "v$RELEASE_VERSION"
```
deltamarnix, line 142: I think I'd rather see that you make use of the github actions from the marketplace instead of using the git commands, so it's clear per step which step went wrong with additional logging if the actions provide it.

savente93: That's fair, I'll have a look at that

deltamarnix:
Could you try this out on a fork and see if the release process goes as expected?

.github/workflows/finish-release.yml
```diff
@@ -109,33 +125,50 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-      - name: Finalise PR
+        with:
+          fetch-depth: 0 # get the whole history not just the one commit
+          ref: main
+
+      - name: Determine relase version
```
Tjalling-dejong, line 132: small typo: relase -> release

Tjalling-dejong (approved):
I cant really test this, but it looks good to me.  I only spotted a typo that you should fix.

savente93:
I've tested this on my fork where it works well. I'll be adding some quick documentation and merging this tomorrow, then I'll release v1.1

PR #1179, opened 2025-04-10
Add tomlkit

Tjalling-dejong:
## Issue addressed

Fixes Deltares/hydromt_wflow#79

## Explanation

Replaced tomli and tomli-w with tomlkit

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist


hboisgon:
Nice addition! Please note that this will not fix the issue in hydromt_wflow as hydromt_wflow implemented its own ``write_config`` and ``_configwrite``. This can be harmonised when we work on the config component or on the value line from the setup_constant_pars issue in hydromt_wflow but for safety I would either not close the issue in hydromt_wflow when this PR gets merged or create straightaway a new issue.

savente93 (commented):
code looks good, but I'd like to see some more documentation. If this is public behavior we're committing to, I think it would be good to document which structures we generally uphold (e.g. order of tables and comments, but not necessary order of array items, (this is just an example, I don't know if this is true)). In addition I think it's also good to mention that as far as I know, toml is the only format for which we can do this for?

tests/io/test_io.py
```diff
@@ -241,3 +242,35 @@ def test_open_mfcsv_by_var(tmpdir, dfs_segmented_by_vars):
...
+
+    """
+    test_toml_fp = tmp_path / "test.toml"
+    with open(test_toml_fp, "w") as f:
+        f.write(toml_content)
+
+    # Read toml
+    config = _read_toml(test_toml_fp)
+
+    # Check if structure and comment is preserved
```
deltamarnix, line 270: Could you also test the cases where there are different levels that we want to preserve? There was this issue (can't find it anymore) where we said we need to keep the structure of:

```toml
[input]
vertical.foo = "bla"
horizontal.bar = "fez"
```

Instead of that it would become:

```toml
[input.vertical]
foo = "bla"

[input.horizontal]
bar = "fez"
```

deltamarnix (changes requested):
I have a request for another test. If you can find the corresponding issue, please link it here.

savente93:
I've thought about it a bit more and I actually also would like to see a test in the actual config component with a default template. i.e. create a config component with a default `toml` template that has things like comments in there, modify some settings, write it out and then verify that the comments are still there

Tjalling-dejong:
@savente93  I added a test using the config component and setting a value to that component. The test showed that the set method breaks the ordering of a toml document because it breaks the internal referencing. I rewrote the set method so that it recursively sets a variable and this fixed the issue.
I did encouter that comments belonging to a toml section will dissapear when you update a value within the section.

hydromt/model/components/config.py
```diff
@@ -312,6 +303,15 @@ def update(self, data: Dict[str, Any]):
...
             self.set(k, v)
 
+    def _set(self, key: str, value: Any, data: dict):
+        if "." in key:
+            parent_key, child_key = key.split(".", maxsplit=1)
+            if parent_key not in data:
+                data[parent_key] = {}
+            self._set(child_key, value, data[parent_key])
+        else:
+            data[key] = value
```
deltamarnix, line 313: I asked the copilot reviewer for some help because I head the feeling that something like this alreaady exists. You can use `functools.reduce` to achieve the same with only 2 lines:

```py
from functools import reduce

def set_nested_dict(key: str, value: Any, data: dict):
    keys = key.split(".")
    reduce(lambda d, k: d.setdefault(k, {}), keys[:-1], data)[keys[-1]] = value
```

savente93:
oh @Tjalling-dejong I also see you added a wflow toml as a test case, this is fine, but please add a comment that this is not intended to be a working wflow example, and is just for the toml testing.

hydromt/model/components/config.py
Tjalling-dejong: Thats neat, I'll implement your solution.

Tjalling-dejong:
> I think these are okay, but please document these somewhere, either in the docs if you can find a good place for it like the config component, or in the doc string. Either is fine
>

I added documentation to the config component docstring. It looks like that location is the most appropriate to document it.

savente93 (approved):
I think this looks good for the core part. Nice work Tjalling.

savente93:
Merging this with an override as Marnix is on holiday until next week and we don't want to keep this open in the meantime.

Issue #1195, opened 2025-05-21
Warning on providing a mask


dalmijn:
### HydroMT version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

Request geospatial data from the data catalog using a mask, either bbox or geom.
Make sure a logger is present.

### Current behaviour

Gives an ungodly warning with the geometries entirely in the message.
This happend in the `resolve` method of the `ConventionResolver` class.

### Desired behaviour

No warning.

### Additional context

No.

PR #1215, opened 2025-07-15
Re-enable chunking on read

### LINKED ISSUE #1193: Chunking not applied in the raster drivers
### HydroMT version checks

### Reproducible Example

The following code will take a long time to compute and then give an memory error
`memoryError: Unable to allocate 280. GiB for an array with shape (1, 174000, 432000) and data type uint32`

```
import hydromt 
from hydromt.model.processes.basin_mask import get_basin_geometry

# Bounding box encompassing the Rhine basin
bbox = [ 3.56583333, 46.32583333, 11.87583333, 52.00833333]

data_catalog = hydromt.DataCatalog(data_libs="deltares_data")
basin_index = data_catalog.get_source("merit_hydro_index")

# This will take forever, if it even loads
ds = ds = data_catalog.get_rasterdataset("merit_hydro", bbox=bbox)

# When I for example whan to generate the Rhine basin it will give a memory error
gdf_bas, _ =  get_basin_geometry(ds=ds, basin_index=basin_index, xy=[5.076212, 51.826497])

```

The same code for v0.10 works fine: 

```
import hydromt 
from hydromt.workflows import get_basin_geometry

# Bounding box encompassing the Rhine basin
bbox = [ 3.56583333, 46.32583333, 11.87583333, 52.00833333]

data_catalog = hydromt.DataCatalog(data_libs="deltares_data")
basin_index = data_catalog.get_source("merit_hydro_index")


ds = ds = data_catalog.get_rasterdataset("merit_hydro", bbox=bbox)

gdf_bas, _ =  get_basin_geometry(ds=ds, basin_index=basin_index, xy=[5.076212, 51.826497])

```

### Current behaviour

I think the issue is related to dask. In any case the chunking is not working

### Desired behaviour

The v1 code should have the same behaviour as  v0.10

### Additional context

_No response_

dalmijn:
## Issue addressed

Fixes #1193 #1195 

## Explanation

Added chunking as an argument in the signature of the `get_rasterdataset` method of `DataCatalog`. This arguments trickles all the way down to the drivers. In the drivers (i.e. `RasterDatasetXarrayDriver` & `RasterioDriver`) a check is done on the signature and the chunking present under driver options in the data catalog yaml file. The signature is prioritised over the the definition of the chunking in the data catalog. Mosaic raster's (not vrt's) cause small issues in the chunking, which resulted in resetting the chunking after reading.
Added a flag only available in the data catalog whether or not to enable local caching of vrt's after chatting a bit with @roeldegoede. This defaults to `False` but can be enabled by settings 'cache' under driver options.
The caching of vrt files was broken in general, so a fix based on gdal's `gdalbuildvrt` has been implemented to combat that. Now everytime new files get cached of a specific dataset, a new vrt is created based on all cached tiles.

EDIT:
Removed support for Python 3.9 as import issues arose after updating the enviroments due to the addition of gdal. As 3.10 is also already discouraged from using according to e.g. [this site](https://scientific-python.org/specs/spec-0000/) it has been removed as well.
Pinned shapely to >= 2.1 as a change was made in how the function `oriented_envelope` return the coordinates.
Removed settings the 'cache' flag from the data catalog yaml file as the `DataCatalog` already had a flag for caching in it's constructor. 
Added testing for vrt caching, as it didnt already exist explicitly.

EDIT EDIT:
Made gdal an optional dependency, which means that caching of the vrt's is not available until gdal and it's bindings are installed.

## General Checklist

- [x] Updated documentation
- [x] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/_utils/caching.py
+++ b/hydromt/_utils/caching.py
@@ -12,15 +12,19 @@ import geopandas as gpd
 import numpy as np
 from affine import Affine
 from fsspec import AbstractFileSystem, url_to_fs
-from osgeo import gdal
 from pyproj import CRS
 
+from hydromt._compat import HAS_GDAL
 from hydromt._typing.type_def import StrPath
 from hydromt._utils.uris import _strip_scheme, _strip_vsi
 from hydromt.config import SETTINGS
 
-logger = logging.getLogger(__name__)
+if HAS_GDAL:
+    from osgeo import gdal
 
+    gdal.UseExceptions()
+
+logger = logging.getLogger(__name__)
 
 __all__ = ["_copy_to_local", "_cache_vrt_tiles"]
 
@@ -109,6 +113,10 @@ def _cache_vrt_tiles(
     vrt_destination_path : Path
         Path to cached vrt
     """
+    if not HAS_GDAL:
+        raise ImportError("Can't cache vrt's without GDAL installed.")
+
+    # Get the filesystem type
     if fs is None:
         fs: AbstractFileSystem = url_to_fs(vrt_uri)[0]
 
--- a/hydromt/data_catalog/data_catalog.py
+++ b/hydromt/data_catalog/data_catalog.py
@@ -106,7 +106,9 @@ class DataCatalog(object):
             If None, no default data catalog is used.
         cache: bool, optional
             Set to true to cache data locally before reading.
-            Currently only implemented for tiled rasterdatasets, by default False.
+            Currently only implemented for tiled rasterdatasets.
+            Requires the installation of GDAL and it's python bindings.
+            By default False.
         cache_dir: str, Path, optional
             Directory root path to cache data to, by default ~/.hydromt
         logger : logger object, optional
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -15,7 +15,6 @@ dependencies = [
   "click",                  # CLI configuration
   "dask",                   # lazy computing
   "fsspec[http]",           # general file systems utilities
-  "gdal>=3.5.0",            # Our favourite geospatial library
   "geopandas[all]>=0.10",   # pandas but geo, wraps fiona and shapely
   "importlib_metadata",     # entrypoints backport
   "mercantile",             # tile handling
@@ -65,8 +64,9 @@ dynamic = ['version', 'description']
 [project.optional-dependencies]
 io = [
   "gcsfs>=2023.12.1", # google cloud file system
-  "openpyxl",         # excel IO
+  "gdal>=3.5.0",      # Our favourite geospatial library
   "fastparquet",      # parquet IO
+  "openpyxl",         # excel IO
   "pillow",           # image IO
   "s3fs",             # S3 file system
 ]
@@ -205,15 +205,15 @@ python = "3.12.*"
 [tool.pixi.feature.py313.dependencies]
 python = "3.13.*"
 
-[tool.pixi.dependencies]
-gdal = ">=3.5.0"
-pyflwdir = ">=0.5.4"
-
 [tool.pixi.pypi-dependencies]
 hydromt = { path = ".", editable = true }
 
 [tool.pixi.feature.doc.dependencies]
 pandoc = "*"
+
+[tool.pixi.feature.io.dependencies]
+gdal = ">=3.5.0"
+pyflwdir = ">=0.5.4"
 
 [tool.pixi.feature.doc.tasks]
 doctest = { cmd = [
@@ -366,6 +366,7 @@ full-latest = { features = [
   "doc",
   "examples",
 ], solve-group = "py313" }
+min-latest = { features = ["py313", "test"], solve-group = "py313" }
 slim-latest = { features = [
   "py313",
   "io",
--- a/hydromt/_compat.py
+++ b/hydromt/_compat.py
@@ -5,10 +5,25 @@ from packaging.version import Version
 
 __all__: List[str] = []
 
-HAS_PYET = False
 HAS_GCSFS = False
-HAS_S3FS = False
+HAS_GDAL = False
 HAS_OPENPYXL = False
+HAS_PYET = False
+HAS_S3FS = False
+
+try:
+    import gcsfs
+
+    HAS_GCSFS = True
+except ImportError:
+    pass
+
+try:
+    from osgeo import gdal
+
+    HAS_GDAL = True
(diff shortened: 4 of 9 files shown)


savente93 (approved):
I'm not too familiar with how VRT's work so I'm not sure I can really comment on the semantics of the code, but I see nothing wrong and it's well tested, so I'm happy to accept this.

pyproject.toml
```diff
@@ -15,6 +15,7 @@ dependencies = [
   "click",                  # CLI configuration
   "dask",                   # lazy computing
   "fsspec[http]",           # general file systems utilities
+  "gdal>=3.5.0",            # Our favourite geospatial library
```
deltamarnix: Dirk mentioned that adding gdal as a dependency is going to have issues with pip installations. It would make sense if we add vrt under a separate feature flag, so that users can install it if they know what they are doing. And I would add it to the conda recipe to always add it.

```toml
vrt = ["gdal>=3.5.0"]
```

And also explain in the documentation what adding this extra optional feature would mean for the user.

pyproject.toml
```diff
@@ -214,6 +205,10 @@ python = "3.12.*"
 [tool.pixi.feature.py313.dependencies]
 python = "3.13.*"
 
+[tool.pixi.dependencies]
+gdal = ">=3.5.0"
+pyflwdir = ">=0.5.4"
```
deltamarnix: Is pyflwdir not available under pypi?

pyproject.toml
```diff
@@ -109,9 +108,6 @@ examples = [
   "notebook",   # jupyter integration
 ]
 
-full = ["hydromt[io,extra,dev,test,doc,examples]"]
-slim = ["hydromt[io,extra,examples]"]
```
deltamarnix, line 113: Why did you decide to remove these feature sets? Is it never used by users?

hydromt/_utils/caching.py
```diff
@@ -199,10 +196,44 @@ def _cache_vrt_tiles(
...
+
+    # Build the vrt with gdal
+    out_ds = gdal.BuildVRT(
+        destName=vrt_destination_path.as_posix(),
+        srcDSOrSrcDSTab=new,
+    )
+
+    # Close and dereference the gdal dataset
+    out_ds.Close()
+    out_ds = None
```
deltamarnix, line 245:
```suggestion
    del out_ds
```

Although I'm not sure if that would be needed. `out_ds` goes out of scope anyways when the method is finished. So it looks superfluous to me.

hydromt/data_catalog/data_catalog.py
```diff
@@ -1330,11 +1336,16 @@ def get_rasterdataset(
         else:
             raise ValueError(f'Unknown raster data type "{type(data_like).__name__}"')
 
+        # This isnt briliant but works the best at this stage
+        if isinstance(source.driver, RasterioDriver):
+            source.driver.options.update({"cache": self.cache})
```
deltamarnix, line 1343: Is this something you could add to every driver, instead of checking? Then let the driver decide if it will do anything with it. That would remove the dependency.

hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -130,6 +134,12 @@ def _open() -> Union[xr.DataArray, xr.Dataset]:
         else:
             ds = _open()
 
+        # Mosiac's can mess up the chunking, which can error during writing
+        # Or maybe setting
+        chunks = options.get("chunks")
```
deltamarnix, line 139: Changing everything to `pop`, except this one? ;)

hydromt/data_catalog/sources/rasterdataset.py
```diff
@@ -50,6 +50,7 @@ def read_data(
         variables: Optional[List[str]] = None,
         time_range: Optional[TimeRange] = None,
         zoom: Optional[Zoom] = None,
+        chunks: Optional[dict] = None,
```
deltamarnix, line 53: From the documentation in xarray: 

> chunks ([int](https://docs.python.org/3/library/functions.html#int), [tuple](https://docs.python.org/3/library/stdtypes.html#tuple) of [int](https://docs.python.org/3/library/functions.html#int), "auto" or [mapping](https://docs.python.org/3/glossary.html#term-mapping) of [hashable](https://docs.xarray.dev/en/latest/user-guide/terminology.html#term-name) to [int](https://docs.python.org/3/library/functions.html#int) or a TimeResampler, optional)

So maybe it should just be `Any`?

open-vscode.bat
deltamarnix, line 1: Why did you remove this file? I love opening my vs code with this bat file. What's the alternative?

scripts/migrate_catalog_to_v1.py
```diff
@@ -43,6 +43,17 @@
 }
 
 
+class CatalogDumper(SafeDumper):
```
deltamarnix, line 46: Not sure why this is part of your PR.

hydromt/data_catalog/data_catalog.py
dalmijn: Not really. Other drivers have their own way of resolving the driver kwargs

hydromt/_utils/caching.py
dalmijn: Considered as good practice when working with gdal shadow objects

open-vscode.bat
dalmijn: Vscode supports pixi envs

hydromt/data_catalog/drivers/raster/rasterio_driver.py
dalmijn: Yep.

pyproject.toml
dalmijn: Yes, but installing hydromt via conda causes issues between numba and pyflwdir, so conda specified. I dunno maybe some discrepancy between the pypi recipe and conda recipe of pyflwidir

hydromt/data_catalog/sources/rasterdataset.py
dalmijn: No, in hydromt it's passed around internally as a dict.

savente93:
Marnix has gone on vacation so in the interest of being able to move on, I discussed the issues with Brendan and overruled the requested changes. We can address things if they are still relevant when he returns.

PR #1228, opened 2025-07-25
Clarify and update write and read nc

dalmijn:
## Issue addressed

Fixes #<issue number>

## Explanation

Added compression and CF compliant dimensions encoding to `_write_nc`. Also added the option do to delay the compute of the datasets and use a dask progressbar instead. Also made `_write_nc` write a single file a simple loop over the data of the components where necessary. This in order to take away the footfun that was the encoding in the old `_write_nc`. Separated `_read_nc` into  `_read_nc` for a single file and `_read_ncs` for the read of multiple files using a pattern.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/_io/writers.py
+++ b/hydromt/_io/writers.py
@@ -163,7 +163,7 @@ def _write_nc(
     force_overwrite: bool = False,
     **kwargs,
 ) -> Optional[DeferedFileClose]:
-    """Write dictionnary of xarray.Dataset and/or xarray.DataArray to netcdf files.
+    """Write xarray.Dataset and/or xarray.DataArray to netcdf file.
 
     Possibility to update the xarray objects attributes to get GDAL compliant NetCDF
     files, using :py:meth:`~hydromt.raster.gdal_compliant`.
@@ -177,7 +177,7 @@ def _write_nc(
     Parameters
     ----------
     ds : xr.DataArray | xr.Dataset
-        Data to be written to the drive
+        Dataset to be written to the drive
     filepath : Path | str
         Full path to the outgoing file
     compress : bool, optional
--- a/hydromt/_io/writers.py
+++ b/hydromt/_io/writers.py
@@ -3,7 +3,7 @@
 import os
 from logging import Logger, getLogger
 from os import makedirs
-from os.path import dirname, exists, isdir, isfile, join
+from os.path import dirname, exists, isdir, join
 from pathlib import Path
 from tempfile import TemporaryDirectory
 from typing import Any, Dict, List, Optional, Union, cast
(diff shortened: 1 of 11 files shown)


hydromt/_io/writers.py
```diff
@@ -119,18 +120,50 @@ def _zarr_writer(
...
+    obj = ds.to_netcdf(
+        filepath,
+        compute=compute,
+        **kwargs,
+    )
+    if compute:
+        return
+
+    with ProgressBar():
+        obj.compute()
```
aslowwriter, line 151: I'm not a fan of having the progress bar in there, since it's messed with the output of documentation and examples before. is this really necessary?

hydromt/_io/writers.py
```diff
@@ -143,54 +176,82 @@ def _write_nc(
...
+    logger.debug(f"Writing file {filepath.as_posix()}")
+    # Check whether the file already exists
+    if filepath.is_file() and not force_overwrite:
+        raise IOError(f"File {filepath.as_posix()} already exists")
+    if not filepath.parent.is_dir():
+        filepath.parent.mkdir(parents=True)
+
+    # Focus on the encoding and set these for all dims, coords and data vars
+    encoding = kwargs.pop("encoding", {})
+    for var in set(list(ds.coords) + list(ds.data_vars)):
```
aslowwriter:
```suggestion
    for var in set(ds.coords) | set(ds.data_vars):
```

savente93 (changes requested):
Looks good to me aside from the progress bar issue. After we resolve that we can merge

hydromt/_io/writers.py
dalmijn: It's disabled by default, when writing e.g. the forcing it is quite handy and then it can be enabled, it only is enabled when compute is set to false

hydromt/_io/writers.py
dalmijn: Sets cannot be appended.

aslowwriter: my bad, this should work tho

hydromt/_io/writers.py
aslowwriter: okay, it's not a huge change so we can try it

savente93 (approved):
Updated my suggestion, otherwise I think this is fine. Thanks

hydromt/_io/writers.py
dalmijn: Good one, will do

hydromt/_io/writers.py
dalmijn: To add to it, it is controlled, essentially, by the components. Where doing nothing results in no progressbar

PR #1229, opened 2025-07-25
feat: separate core functionality from nice-to-have utils in Grid-&Mesh Components

### LINKED ISSUE #1200: Upgrade the GridComponent
### Kind of request

Adding new functionality

### Enhancement Description

The ``GridComponent`` does a little too many things at the moment which makes it sometimes tough to re-use. Especially the ``create_grid`` and ``add_data_to_grid`` methods may not always be good to expose by default.

There would be two ways to get around it:

- have a "lean" version of ``GridComponent`` where we do not have the create or add methods. Names would then be ``GridComponent`` (lean version) and ``GridExtraComponent`` (with create and add methods)
- Remove those functions and make sure they are fully available in worflows/processes.
- Hybrid (if possible): make these functions optional

Pros and cons of each solution:

#### Creating a new component

- The extra methods are still directly available for the plugin if they want without any coding/maintenance. But they are all available.
- The extra methods are also available for core users if they do not always want to develop a plugin. They then only may need to convert from netcdf/geojson/toml/yaml to their specific format if required. This is also useful for hydromt core documentation/examples.
- Drawback: extra maintenance? (add_data and create in both component and processes but identical... GridExtraComponent would fully inherit from GridComponent with the extra methods create and add_data only)

#### Moving to workflows

- Plugins can choose which of the methods they want to allow but they need to add the method call and documentation in their own components. They will not be able to copy-paste from core unless we do store the examples to be copy-pasted somewhere  (leading to maintenance on both core and multiple plugins for the same function).
- Plugins or (extended) python code do become necessary even for simple models.
- Less maintenance?

#### Hybrid

- An extra argument when you instantiate the components to choose which methods to include. It's a little more work on core config and relatively small on plugin side as they only need to list the methods they would like to allow/expose.
- No need to create a new component but need to check if it is feasable without too much extra maintenance.
- Instantiation might be harder to test and no tab completion help.

Some example of what it could look like (check component inheritance):

``` python

@hydromt_step # check if handled here or in component __init__
def create_grid(self):
    da = processes.create_grid
    self.set(da)

@hydromt_step
def add_grid_from_rasterdataset(self, raster_fn):
   da_in = self.data_catalog.get_rasterdataset(raster_fn)
   da_out = processes.grid_from_raster(da_in)
   self.set(da_out)

GRID_METHODS = {
     "create_grid": create_grid,
     "add_grid_from_rasterdataset": add_grid_from_rasterdataset,
}

class GridComponent:
    def __init__(self, additional_methods=[]):
        for name in additional_methods:
             if name in GRID_METHODS:
                  setattr(self, name, GRID_METHODS[name].__get__(self))

staticmaps = GridComponent(additional_methods =["create_grid", "add_data_from_rasterdataset"])
```

### Use case

The lean version would be very useful for example states of forcing objects.

### Additional Context

Also some things to fix in the ``GridComponent``:

- set: add masking and fix issues with np.ndarray (elif to if and check if _data is already not empty)
- region_component_file: make it optional to have a geojson written (especially if region_component points to something else)

The decision here also applies to ``MeshComponent``.

LuukBlom:
## Issue addressed

Fixes #1200 

## Explanation

See my comment under the issue, code should be self-explanatory.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/model/components/grid.py
+++ b/hydromt/model/components/grid.py
@@ -30,7 +30,7 @@ from hydromt.model.steps import hydromt_step
 if TYPE_CHECKING:
     from hydromt.model.model import Model
 
-__all__ = ["GridComponent"]
+__all__ = ["GridComponent", "GridExtraComponent"]
 
 logger: Logger = getLogger(__name__)
 
@@ -49,7 +49,7 @@ class GridComponent(SpatialModelComponent):
         *,
         filename: str = "grid/grid.nc",
         region_component: Optional[str] = None,
-        region_filename: str = "grid/grid_region.geojson",
+        region_filename: Optional[str] = "grid/grid_region.geojson",
     ):
         """
         Initialize a GridComponent.
@@ -63,13 +63,14 @@ class GridComponent(SpatialModelComponent):
             By default "grid/grid.nc".
         region_component: str, optional
             The name of the region component to use as reference for this component's region.
+            If provided, the region is not written to disk.
             If None, the region will be set to the grid extent. Note that the create
             method only works if the region_component is None. For add_data_from_*
             methods, the other region_component should be a reference to another
             grid component for correct reprojection.
-        region_filename: str
-            The path to use for reading and writing of the region data by default.
-            By default "grid/grid_region.geojson".
+        region_filename: Optional[str] = "grid/grid_region.geojson",
+            The path to use for writing the region data to a file.
+            By default "grid/grid_region.geojson". If None, the region is not written to disk.
         """
         # region_component referencing is not possible for grids. The region should be passed via create().
         super().__init__(
@@ -108,12 +109,17 @@ class GridComponent(SpatialModelComponent):
         if isinstance(data, np.ndarray):
             if data.shape != self._data.raster.shape:
                 raise ValueError("Shape of data and grid maps do not match")
+            if data.size == 0:
+                raise ValueError("Cannot set grid data with empty array")
             data = xr.DataArray(dims=self._data.raster.dims, data=data, name=name)
-        elif isinstance(data, xr.DataArray):
+            data = data.where(self._data.raster.mask)
+
+        if isinstance(data, xr.DataArray):
             if name is not None:
                 data.name = name
             data = data.to_dataset()
-        elif not isinstance(data, xr.Dataset):
+
+        if not isinstance(data, xr.Dataset):
             raise ValueError(f"cannot set data of type {type(data).__name__}")
 
         if len(self._data) == 0:  # empty grid
@@ -161,7 +167,18 @@ class GridComponent(SpatialModelComponent):
         """
         self.root._assert_write_mode()
         region_options = region_options or {}
-        self.write_region(**region_options)
+
+        if self._region_component is None:
+            if self._region_filename:
+                self.write_region(**region_options)
+            else:
+                logger.debug(
+                    "Skip writing region to disk: no region component or filename provided to GridComponent."
+                )
+        else:
+            logger.debug(
+                f"Skip writing region to disk: region component `{self._region_component}` was provided to GridComponent --- it should handle writing."
+            )
 
         if len(self.data) == 0:
             exec_nodata_strat(
@@ -169,7 +186,7 @@ class GridComponent(SpatialModelComponent):
                 strategy=NoDataStrategy.WARN,
             )
             return None
-        self.write_region()
+
         # write_nc requires dict - use dummy 'grid' key
         return _write_nc(
             {"grid": self.data},
@@ -221,6 +238,120 @@ class GridComponent(SpatialModelComponent):
         )
         for ds in loaded_nc_files.values():
             self.set(ds)
+
+    @property
+    def res(self) -> Optional[Tuple[float, float]]:
+        """Returns the resolution of the model grid."""
+        if len(self.data) > 0:
+            return self.data.raster.res
+        exec_nodata_strat(
+            msg="No grid data found for deriving resolution",
+            strategy=NoDataStrategy.WARN,
+        )
+        return None
+
+    @property
+    def transform(self) -> Optional[Affine]:
+        """Returns spatial transform of the model grid."""
+        if len(self.data) > 0:
+            return self.data.raster.transform
+        exec_nodata_strat(
+            msg="No grid data found for deriving transform",
+            strategy=NoDataStrategy.WARN,
+        )
+        return None
+
+    @property
+    def crs(self) -> Optional[CRS]:
+        """Returns coordinate reference system embedded in the model grid."""
+        if self.data.raster is None:
+            exec_nodata_strat(
+                msg="No grid data found for deriving crs",
+                strategy=NoDataStrategy.WARN,
+            )
+            return None
+        if self.data.raster.crs is None:
+            exec_nodata_strat(
+                msg="No crs found in grid data",
+                strategy=NoDataStrategy.WARN,
+            )
+            return None
+        return CRS(self.data.raster.crs)
+
+    @property
+    def bounds(self) -> Optional[Tuple[float, float, float, float]]:
+        """Returns the bounding box of the model grid."""
+        if len(self.data) > 0:
+            return self.data.raster.bounds
+        exec_nodata_strat(
+            msg="No grid data found for deriving bounds",
+            strategy=NoDataStrategy.WARN,
+        )
+        return None
+
+    @property
+    def _region_data(self) -> Optional[gpd.GeoDataFrame]:
+        """Returns the geometry of the model area of interest."""
+        if len(self.data) > 0:
+            crs: Optional[Union[int, CRS]] = self.crs
+            if crs is not None and hasattr(crs, "to_epsg"):
+                crs = crs.to_epsg()  # not all CRS have an EPSG code
+            return gpd.GeoDataFrame(geometry=[box(*self.bounds)], crs=crs)
+        exec_nodata_strat(
+            msg="No grid data found for deriving region", strategy=NoDataStrategy.WARN
+        )
+        return None
+
+    @property
+    def data(self) -> xr.Dataset:
+        """Model static gridded data as xarray.Dataset."""
+        if self._data is None:
+            self._initialize_grid()
+        assert self._data is not None
+        return self._data
+
+    def _initialize_grid(self, skip_read: bool = False) -> None:
+        """Initialize grid object."""
+        if self._data is None:
+            self._data = xr.Dataset()
+            if self.root.is_reading_mode() and not skip_read:
+                self.read()
+
+    def test_equal(self, other: ModelComponent) -> Tuple[bool, Dict[str, str]]:
+        """Test if two components are equal.
+
+        Parameters
+        ----------
+        other : ModelComponent
+            The component to compare against.
+
+        Returns
+        -------
+        Tuple[bool, Dict[str, str]]
+            True if the components are equal, and a dict with the associated errors per property checked.
+        """
+        eq, errors = super().test_equal(other)
+        if not eq:
+            return eq, errors
+        other_grid = cast(GridComponent, other)
+        try:
+            xr.testing.assert_allclose(self.data, other_grid.data)
+        except AssertionError as e:
+            errors["data"] = str(e)
+
+        return len(errors) == 0, errors
+
+
+class GridExtraComponent(GridComponent):
+    """ModelComponent class for grid components with extra functions for creating and adding data.
+
+    This class is used for setting, creating, writing, and reading regular grid data for a
+    HydroMT model. The grid component data stored in the ``data`` property of this class is of the
+    hydromt.gis.raster.RasterDataset type which is an extension of xarray.Dataset for regular grid.
+
+    Compared to the GridComponent, this class contains additional methods for creating
+    grid data from a region, adding data from constants, raster datasets, and geodataframes.
+    """
 
     @hydromt_step
     def create_from_region(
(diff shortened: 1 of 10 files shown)


docs/guides/plugin_dev/migrating_to_v1.rst
```diff
@@ -374,8 +374,9 @@ The `GridMixin` and `GridModel` have been restructured into one `GridComponent`
 a weak reference to one general `Model` instance. The `set_grid`, `write_grid`,
 `read_grid`, and `setup_grid` have been changed to the more generically named `set`,
 `write`, `read`, and `create` methods respectively. Also, the `setup_grid_from_*`
-methods have been changed to `add_data_from_*`. The functionality of the GridComponent
-has not been changed compared to the GridModel.
+methods have been removed with the goal of simplifying the API. Removed functionality
+was moved to the `hydromt.model.processes.grid` submodule containing the workflow functions.
+The functionality of the GridComponent has not been changed compared to the GridModel.
```
hboisgon: I guess that then it did

```suggestion
```

docs/guides/plugin_dev/workflows.rst
```diff
@@ -110,11 +56,7 @@ Below is an example of the YAML format used in HydroMT:
...
-- **`grid.create_from_region`**: Creates a grid based on a specified bounding box (bbox), resolution, and coordinate reference system (CRS). Additional options include setting basin and hydrography paths.
-- **`grid.add_data_from_constant`**: Adds a constant value to the grid. Parameters like `name`, `dtype`, and `nodata` specify how the constant data is handled.
-- **`grid.add_data_from_rasterdataset`**: Adds data from a raster dataset. It includes options to specify variables, reprojection methods, and renaming rules for variables.
-- **`grid.add_data_from_raster_reclass`**: Reclassifies raster data based on a specified reclassification table and applies transformations to the grid.
-- **`grid.add_data_from_geodataframe`**: Adds vector data to the grid, rasterizing specific attributes, handling nodata values, and renaming variables.
+- **`grid.read`**: Reads the grid data from a specified file (e.g., `grid.nc`). You can generate and customize this file by using the workflow functions in the submodule `hydromt.model.processes.grid.py`.
```
hboisgon: Not true anymore. People have to create their own plugin with the implementation of the functions in processes.grid in order to use them in a hydromt build configuration YAML file.

```suggestion
- **`grid.read`**: Reads the grid data from a specified file (e.g., `grid.nc`). 
```

hydromt/model/components/grid.py
```diff
@@ -106,14 +99,19 @@ def set(
         if name is None and name_required:
             raise ValueError(f"Unable to set {type(data).__name__} data without a name")
         if isinstance(data, np.ndarray):
+            if data.size == 0:
+                raise ValueError("Cannot set grid data with empty array")
             if data.shape != self._data.raster.shape:
                 raise ValueError("Shape of data and grid maps do not match")
             data = xr.DataArray(dims=self._data.raster.dims, data=data, name=name)
-        elif isinstance(data, xr.DataArray):
+            data = data.where(self._data.raster.mask)
```
hboisgon: Why add this line to np.ndarray? If we start to allow masking here, should be done for all types of data and not just np.ndarray (also xr objects). If you implement masking I would add a flag and copy from how we do it in WflowModel.staticmaps.set

hydromt/model/components/grid.py
```diff
@@ -161,15 +159,26 @@ def write(
...
         self.root._assert_write_mode()
         region_options = region_options or {}
-        self.write_region(**region_options)
+
+        if self._region_component is None:
+            if self._region_filename:
+                self.write_region(**region_options)
+            else:
+                logger.debug(
+                    "Skip writing region to disk: no region component or filename provided to GridComponent."
```
hboisgon: In case no region component the other message below should appear
```suggestion
                    "Skip writing region to disk: no region component filename provided to GridComponent."
```

hydromt/model/components/grid.py
```diff
@@ -723,6 +346,6 @@ def _get_grid_data(self) -> Union[xr.DataArray, xr.Dataset]:
                 )
             return reference_component.data
 
-        if self.data is None:
+        if self._data is None:
```
hboisgon: It was good before. Compared to _data it will trigger read if allowed and not already done (eg important in update mode)
```suggestion
        if self.data is None:
```

hydromt/model/components/grid.py
```diff
@@ -723,6 +346,6 @@ def _get_grid_data(self) -> Union[xr.DataArray, xr.Dataset]:
...
             raise ValueError("Unable to get grid data from this component.")
-        return self.data
+        return self._data
```
hboisgon: It was good before. Compared to _data it will trigger read if allowed and not already done (eg important in update mode)
```suggestion
        return self.data
```

hydromt/model/components/mesh.py
```diff
@@ -715,61 +451,21 @@ def _grid_is_equal(self, grid_name: str, data: xu.UgridDataset) -> bool:
...
-                f"Referenced region component is not a MeshComponent: '{self._region_component}'."
-            )
-        if reference_component.data is None:
-            raise ValueError(
-                f"Unable to get mesh data from the referenced region component: '{self._region_component}'"
-            )
-        if grid_name not in reference_component.mesh_names:
-            raise ValueError(
-                f"Grid '{grid_name}' not found in mesh of '{self._region_component}'."
-            )
```
hboisgon: Why remove these methods? They are not of type add_data (ie they do not call self.set in the end) but classic mesh manipulation (like in geoms you added remove  or other equivalent methods)

hydromt/model/processes/mesh.py
```diff
@@ -171,35 +171,47 @@ def create_mesh2d_from_mesh(
 
     # Check crs and reproject to model crs
     grid_crs = grid.crs
+    idx = [g.name for g in mesh2d.ugrid.grids].index(grid.name)
+
```
hboisgon: Why did you add this idx? The code above should make sure we have a single grid selected based on ``grid_name`` argument.

tests/cli/test_cli.py
hboisgon, line 1: Do we need to replace these tests somehow? @savente93

hydromt/model/components/grid.py
LuukBlom: Ah I misread the issue then, I assumed the masking only had to do with the np.ndarray.

> Additional Context
> Also some things to fix in the GridComponent:
> 
> set: add masking and fix issues with np.ndarray (elif to if and check if _data is already not empty)

I'll update

hydromt/model/components/grid.py
LuukBlom: Brendan asked me to add this since he needed it for the ForcingComponent to get out of an infinite loop. 
please see: https://github.com/Deltares/hydromt_wflow/pull/513/files#r2250647838

hydromt/model/components/grid.py
LuukBlom: Brendan asked me to add this since he needed it for the ForcingComponent to get out of an infinite loop.
please see: https://github.com/Deltares/hydromt_wflow/pull/513/files#r2250647838

hydromt/model/components/mesh.py
LuukBlom: These werent used anywhere else, so I thought removing would be best. I'll add them back in!

hydromt/model/processes/mesh.py
LuukBlom: While rewriting the tests, the `ugrid.grid` accessor produced an error saying multiple grids were defined (and there were) and that you should use `.grids[index]`. 

I agree I dont love the change though

tests/cli/test_cli.py
aslowwriter: I think the code involved is tested sufficiently by our other tests that we can leave these. They should be tested sufficiently by the testing of the internal methods I think

LuukBlom:
> Don't forget to update the changelog ;) 

Ah good one, thanks! 

> I guess the examples notebooks did not need updating ?

Correct! None of the removed functions were used.

hydromt/model/processes/mesh.py
LuukBlom: I found my mistake and reverted the changes!

docs/changelog.rst
```diff
@@ -17,6 +17,7 @@ New
 
 Changed
 -------
+When setting data in the GridComponent, mask the new data using the existing data as a mask. (#1229)
```
hboisgon:
```suggestion
Allow masking in `GridComponent.set`. (#1229)
```

docs/changelog.rst
```diff
@@ -25,6 +26,11 @@ Fixed
 Deprecated
 ----------
 
+Removed
+-------
+- From the `GridComponent`, the functions `create_from_region`, `add_data_from_constant`, `add_data_from_rasterdataset`, `add_data_from_raster_reclass`, `add_data_from_geodataframe`. (#1229)
+- From the `MeshComponent`, the functions `create_2d_from_region`, `add_2d_data_from_rasterdataset`, `add_2d_data_from_raster_reclass`. (#1229)
```
hboisgon:
```suggestion
- From the `GridComponent`, the functions `create_from_region`, `add_data_from_constant`, `add_data_from_rasterdataset`, `add_data_from_raster_reclass`, `add_data_from_geodataframe`. (#1229) Equivalent workflows are still available.
- From the `MeshComponent`, the functions `create_2d_from_region`, `add_2d_data_from_rasterdataset`, `add_2d_data_from_raster_reclass`. (#1229) Equivalent workflows are still available.
```

hydromt/model/components/grid.py
```diff
@@ -106,22 +104,38 @@ def set(
...
                 data.name = name
             data = data.to_dataset()
-        elif not isinstance(data, xr.Dataset):
+
+        if not isinstance(data, xr.Dataset):
             raise ValueError(f"cannot set data of type {type(data).__name__}")
 
+        # Check if noth is really up and south therefore is down
+        if data.raster.res[1] > 0:
+            data = data.raster.flipud()
```
hboisgon: This is Wflow specific, e.g. SFINCS does not want this. Either remove or add an argument to the function for NS or SN orientation.

hydromt/model/components/grid.py
```diff
@@ -96,6 +90,10 @@ def set(
         name: str, optional
             Name of new map layer, this is used to overwrite the name of a DataArray
             and ignored if data is a Dataset
+        mask: xr.DataArray, optional
+            Name of the mask layer in the grid data, or directly the mask layer to use.
```
hboisgon:
```suggestion
            Name of the mask layer in the grid (self) or data, or directly the mask layer to use.
```

hydromt/model/components/grid.py
```diff
@@ -96,6 +90,10 @@ def set(
...
+            Can be a DataArray where `.raster.nodata` is used to define the mask.
```
hboisgon:
```suggestion
            Should be a DataArray where `.raster.nodata` is used to define the mask.
```

hydromt/model/components/grid.py
hboisgon: I still think it should be ``self.data``. I see the issue is that Brendan is adding checks in ``ForcingComponent.data``. It should not be. Any checks is done in ``set`` which controls already anything that will be ``data`` so not need to double-check and then enter in a infinite loop. See where the extra not needed checks are located causing the infinite loop: https://github.com/Deltares/hydromt_wflow/blob/e0c3408ef4d0e78e0792d0d2c04b4cd723af157b/hydromt_wflow/components/forcing.py#L102-L105

hydromt/model/processes/mesh.py
```diff
@@ -166,11 +166,23 @@ def create_mesh2d_from_mesh(
...
+    if "name" not in mesh2d.ugrid.grid.attrs:
+        mesh2d.ugrid.grid.attrs["name"] = grid.name
+    else:
+        u_name = mesh2d.ugrid.grid.attrs["name"]
+        if grid.name != u_name:
+            logger.warning(
+                f"Grid name `{grid.name}` does not match mesh2d name `{u_name}`. "
+                "Setting mesh2d name to grid name."
+            )
+            mesh2d.ugrid.grid.attrs["name"] = grid.name
```
hboisgon: Is this all necessary? mesh2d is created directly from grid.to_dataset so would name differ? Especially since it seems grid.name always exist?

hboisgon (changes requested):
Almost there indeed :) I think the infinite loop issue is caused by hydromt_wflow and not hydromt core so I would keep calling self.data directly. The other thing is forcing South-North orientation, some plugins do not want that so it should not automatically be done but can be added with a boolean option for plugins that do want it like Wflow.

docs/changelog.rst
```diff
@@ -17,6 +17,7 @@ New
 
 Changed
 -------
+Allow masking in `GridComponent.set`. (#1229)
```
hboisgon: Nitpicky but could count as Added as this is new functionnality from that function.

hydromt/model/components/grid.py
```diff
@@ -84,6 +77,8 @@ def set(
         self,
         data: Union[xr.DataArray, xr.Dataset, np.ndarray],
         name: Optional[str] = None,
+        mask: Optional[Union[str, xr.DataArray]] = None,
+        north_is_up: bool = True,
```
hboisgon: SHould have mentioned before but in [_write_nc](https://github.com/Deltares/hydromt/blob/e2534f806e4de5b9c93f0cf3b291c85f34442e98/hydromt/_io/writers.py#L155) we use ``force_sn``. Good to harmonise :)

hydromt/model/components/grid.py
```diff
@@ -106,22 +108,45 @@ def set(
...
+        if data.raster.res[1] > 0:  # increasing y = South to North
+            if not north_is_up:
+                data = data.raster.flipud()
+        elif data.raster.res[1] < 0:  # increasing y = North to South
+            if north_is_up:
+                data = data.raster.flipud()
+        else:
+            logger.warning(
+                "Grid data resolution is zero in y direction, cannot determine north-south orientation."
+            )
```
hboisgon: I don't think this is needed. Error will appear from raster.res > raster.ycoords if something is amiss

hboisgon (approved):
Just a few minor comments so I approve already :)

hydromt/model/components/grid.py
LuukBlom: Hmm, now that Im reading it again. What do we want to do here @hboisgon:

```python
    ("lats", "force_sn", "should_flip"),
    [
        ([1.0, 2.0], True, False),  # data=SN, force_sn=True => no flip
        ([2.0, 1.0], True, True),  # data=NS, force_sn=True => flip
        ([1.0, 2.0], False, True),  # data=SN, force_sn=False => flip --- should this one be flipped?
        ([2.0, 1.0], False, False),  # data=NS, force_sn=False => no flip
    ],
```

`force_sn=False` to me can mean 2 things:
1. we do not change incoming data
2. we always force incoming to NS.

PR #1244, opened 2025-08-20
Fix update CLI

### LINKED ISSUE #1242: Cannot run update from CLI
### HydroMT version checks

### Reproducible Example

hydromt update wflow ...

### Current behaviour

Errors because opt is not a valid argument:

```
hydromt.cli.main - main - ERROR - Model.update() got an unexpected keyword argument 'opt'
```

### Desired behaviour

Pass setps=opt["steps"] instead of opt (see build example)

### Additional context

_No response_

hboisgon:
## Issue addressed

Fixes #1242 

## Explanation

Because the config needs to have "steps" as a list, the update CLI command was broken.
I also fixed the tests. Some of them should have been broken already before but were not designed well enough to catch errors.

While fixing these, I also realised the --opt and -c options in the build/update CLI were broken also because they were not converted to the "steps" list. As we removed the way more used --region argument, I decided to stop support and just make the config file mandatory. I still like that you can supply it with a -i or --config option rather than having to follow a precise order so I prefered to add an ValueError message when parsing. 

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/cli/_utils.py
+++ b/hydromt/cli/_utils.py
@@ -75,24 +75,14 @@ def parse_json(_ctx: click.Context, _param, value: str) -> Dict[str, Any]:
 
 ### general parsing methods ##
 def parse_config(
-    path: Optional[Union[Path, str]] = None, opt_cli: Optional[Dict[str, Any]] = None
+    path: Optional[Union[Path, str]] = None,
 ) -> Dict[str, Any]:
-    """Parse config from `path` and combine with command line options `opt_cli`."""
+    """Parse config from `path`."""
     opt = {}
     if path is not None and isfile(path):
         opt = _config_read(path, abs_path=True, skip_abspath_sections=["setup_config"])
     elif path is not None:
         raise IOError(f"Config not found at {path}")
-    if opt_cli is not None:
-        for section in opt_cli:
-            if not isinstance(opt_cli[section], dict):
-                raise ValueError(
-                    "No section found in --opt values: "
-                    "use <section>.<option>=<value> notation."
-                )
-            if section not in opt:
-                opt[section] = opt_cli[section]
-                continue
-            for option, value in opt_cli[section].items():
-                opt[section].update({option: value})
+    else:
+        raise ValueError("Config path is required. Use -i or --config <path>")
     return opt
--- a/hydromt/cli/main.py
+++ b/hydromt/cli/main.py
@@ -112,14 +112,6 @@ verbose_opt = click.option("--verbose", "-v", count=True, help="Increase verbosi
 
 quiet_opt = click.option("--quiet", "-q", count=True, help="Decrease verbosity.")
 
-opt_cli = click.option(
-    "--opt",
-    multiple=True,
-    callback=_utils.parse_opt,
-    help="Method specific keyword arguments, see the method documentation "
-    "of the specific model for more information about the arguments.",
-)
-
 data_opt = click.option(
     "-d",
     "--data",
@@ -199,7 +191,6 @@ def main(ctx, models, components, plugins):
     type=str,
 )
 @arg_root
-@opt_cli
 @opt_config
 @data_opt
 @deltares_data_opt
@@ -212,7 +203,6 @@ def build(
     _ctx: click.Context,
     model,
     model_root,
-    opt,
     config,
     data,
     dd,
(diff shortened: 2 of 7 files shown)


hydromt/model/model.py
```diff
@@ -141,7 +141,7 @@ def _determine_region_component(self, region_component: Optional[str]) -> str:
                 )
             if len(has_region_components) == 0:
                 logger.warning("No region component found in components.")
-                return ""
+                return None
```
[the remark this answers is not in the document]
hboisgon: I checked all calls to this function / property. So the change should be fine and more clear to use None than "" for checks in further functions

hydromt/model/model.py
```diff
@@ -402,9 +404,7 @@ def read(self, components: Optional[List[str]] = None) -> None:
 
     @staticmethod
     def _options_contain_write(steps: List[Dict[str, Dict[str, Any]]]) -> bool:
-        return any(
-            next(iter(step_dict)).split(".")[-1] == "write" for step_dict in steps
-        )
+        return any("write" in next(iter(step_dict)) for step_dict in steps)
```
LuukBlom, line 407: Im not sure this actually works as intended, but correct me if Im wrong.

This only checks the first key in each `step_dict`.

Seems to me we should replace this by something that will loop over all keys across all dictionaries and return True if any contain `"write"`:
```python
def _options_contain_write(steps: List[Dict[str, Dict[str, Any]]]) -> bool:
    for step_dict in steps:
        if "write" in step_dict:
            return True
    return False
```
or 

```python
def _options_contain_write(steps: List[Dict[str, Dict[str, Any]]]) -> bool:
    return any("write" in key for step_dict in steps for key in step_dict)
```

hydromt/cli/main.py
LuukBlom, line 1: I would like type hints and parameter descriptions in the docstrings for the cli functions, so its easy to see what for example `config` or `data` should be.

hydromt/cli/_utils.py
```diff
@@ -75,24 +35,14 @@ def parse_json(_ctx: click.Context, _param, value: str) -> Dict[str, Any]:
...
-                    "No section found in --opt values: "
-                    "use <section>.<option>=<value> notation."
-                )
-            if section not in opt:
-                opt[section] = opt_cli[section]
-                continue
-            for option, value in opt_cli[section].items():
-                opt[section].update({option: value})
+    else:
+        raise ValueError("Config path is required. Use -i or --config <path>")
```
LuukBlom: Is there a reason we allow for path to be `None`? Seems like that should actually never occur

hydromt/model/model.py
hboisgon: "write" just needs to be in the name of the hydromt_steps method (and steps is a list of dict with one dict per method). This implementation worked for the forcing example in hydromt wflow (wflow_update_forcing.ipynb) so I think it should do the job?

hydromt/cli/_utils.py
hboisgon: I added some comments in the main message of this PR. But I think allowing for None would give a clearer error message to the user? Also I prefer to still use a -i or --config option rather than make this mandatory and users need to be careful about order or arguments in their hydromt CLI. But maybe there is a better way to do this directly in the CLI cick options?

LuukBlom: I understand your comments and there is indeed a better way to do this using the click library:

In the functions (`check` & `export`) where its not required (because we check `if config: do_something()`), we can use this `click.option` already defined. (options can be None)
```python
opt_config = click.option(
    "-i",
    "--config",
    type=click.Path(resolve_path=True, exists=True, dir_okay=False, file_okay=True),
    help="Path to hydroMT configuration file, for the model specific implementation.",
)
```

In the functions where path is required (`build` and `update`) using `click.argument` (arguments are always required, but also positional so `-i` wont work anymore.)

```python
arg_config = click.argument(
    "config",
    type=click.Path(resolve_path=True, exists=True, dir_okay=False, file_okay=True),
    help="Path to hydroMT configuration file, for the model specific implementation.",
)
```
Alternatively, we can re-use the option and add to top of the function
```python
def ...:
    if not config:
        raise click.UsageError("Config path is required. Use -i or --config <path>")
...
```

parse config would then become:
```python
def parse_config(path: Union[Path, str]) -> Dict[str, Any]:
    """Parse config from `path`. Raises if file is missing."""
    if not isfile(path):
        raise IOError(f"Config not found at {path}")
    return _config_read(path, abs_path=True, skip_abspath_sections=["setup_config"])
```

But also, this doesnt really matter much imo, so up to you if you want to change it.

deltamarnix: I have made the Option required for build and update.

hydromt/cli/main.py
deltamarnix: Not changing to keep the PR small for now

LuukBlom (approved):

PR #1245, opened 2025-08-22
Revert changed behaviour of "buffer" argument in get_rasterdataset

### LINKED ISSUE #1226: Changed behaviour of "buffer" argument in get_rasterdataset
After migrating to v1, the buffer argument in the get_rasterdataset argument works differently, which has very big consequences for the clipping/sliicing of the data. It seems that the buffer argument should be provided in meters, which in essence is more aligned with for exampe get_geodataframe, and not any longer in pixels. Nonetheless, providing "pixels" seems to be more user friendly. 

For example, when you are using ERA5 data with 0.25 degrees resolution for a small SFINCS domain, the get_rasterdataset now only returns a dataset with length x=1, y=1, which is therefore no 2D raster anymore. Solution for this is to provide a buffer of 25km or so, to still get some surrounding points as well. In the old HydroMT, where we were using the "pixels" instead of "meters", a buffer of 2 pixels was always sufficient, independent on the data resolution.

I don't think there is a right or wrong here, but I was wondering whether this was designed as such on purpose or by accident.

LuukBlom:
## Issue addressed

Fixes #1226 

## Explanation
Introduce the `buffer` arg to various functions in the `RasterDatasetAdapter` to make sure it gets passed to the `ds.raster.clip_bbox` function in `RasterDatasetAdapter._slice_spatial_dimensions`.

`ds.raster.clip_bbox()` still accepts buffer in pixels!

Added tests that assert on the bounds of the returned datasets for 3 cases in `RasterDatasetAdapter`: 

- no mask, no buffer
- yes mask, no buffer
- yes mask, yes buffer

and added 2 tests that test this behaviour from `DataCatalog.get_rasterdataset()`

## General Checklist

- [x] Updated documentation
- [x] Updated changelog.rst

## Data/Catalog checklist


hboisgon (changes requested):
Nice! Thanks for fixing it @LuukBlom! 

All looks good but I would still remove or modify the calls to the function ``_parse_geom_bbox_buffer(geom, bbox, buffer)`` in the different functions because now the buffer argument is used twice with once a value in meters and then the same value in pixel. I think you can still call this function but then buffer can be 0.

PR #1251, opened 2025-08-27
crs from geodataframe has priority over catalog crs

### LINKED ISSUE #1238: Use of crs argument in the DataCatalog
### HydroMT version checks

### Reproducible Example

data_catalog.get_geodataframe with a source that contains crs argument (eg hydro_reservoirs)

### Current behaviour

From the log, it seems the CRS is overwritten. In the previous version, the CRS argument from the data catalog would only be used if it could not be derived from the file (ie geom.crs is None). If CRS can be derived from the file then the data catalog CRS is ignored and fullfills its role as metadata.

The message from the model build log:

```
2025-08-15 13:28:09,053 - hydromt.data_catalog.drivers.geodataframe.pyogrio_driver - pyogrio_driver - WARNING - Overwriting crs of GeoDataFrame to EPSG:4326
```

### Desired behaviour

CRS is not overwritten

### Additional context

_No response_

hboisgon:
## Issue addressed

Fixes #1238 

## Explanation

Data catalog crs should only be used if not defined within the file attributes itself.
While checking other drivers, crs conversion is actually (properly) done in self.transform so in the data adapters rather than in the drivers. There the behavior is as expected.
So I removed this incorrect part form geodataframe driver.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/data_catalog/drivers/geodataframe/pyogrio_driver.py
+++ b/hydromt/data_catalog/drivers/geodataframe/pyogrio_driver.py
@@ -66,9 +66,18 @@ class PyogrioDriver(GeoDataFrameDriver):
         if metadata is not None:
             crs = metadata.crs
         if crs is not None:
-            if gdf.crs is not None:
-                logger.warning(f"Overwriting crs of GeoDataFrame to {crs}")
-            gdf.set_crs(crs)
+            if gdf.crs is None:
+                logger.info(f"Setting crs of GeoDataFrame from catalog crs: {crs}")
+                gdf.set_crs(crs)
+            elif gdf.crs != crs:
+                logger.warning(
+                    f"CRS of GeoDataFrame {gdf.crs} does not match catalog metadata "
+                    f"CRS: {crs}. Keeping the original CRS."
+                )
+        elif gdf.crs is None:
+            raise ValueError(
+                "The GeoDataFrame has no CRS. Set one using the crs option."
+            )
 
         if gdf.index.size == 0:
             exec_nodata_strat(f"No data from driver {self}'.", strategy=handle_nodata)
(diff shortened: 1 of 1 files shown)


hydromt/data_catalog/drivers/geodataframe/pyogrio_driver.py
```diff
@@ -62,14 +67,6 @@ def read(
...
             raise IOError(f"DataFrame from uri: '{_uri}' contains no geometry column.")
 
-        crs = None
-        if metadata is not None:
-            crs = metadata.crs
-        if crs is not None:
-            if gdf.crs is not None:
-                logger.warning(f"Overwriting crs of GeoDataFrame to {crs}")
-            gdf.set_crs(crs)
-
```
LuukBlom, line 72: I dont know if I agree with the issue

The issue says 
> In the previous version, the CRS argument from the data catalog would only be used if it could not be derived from the file (ie geom.crs is None). If CRS can be derived from the file then the data catalog CRS is ignored and fullfills its role as metadata.

But to me it seems that a `crs` argument to `read()`, means that whatever is returned will have that crs. So we either set the crs if the gdf doesnt have it, or we reproject it to convert the crs.

So this code could be:
```python
crs = None
if metadata is not None:
    crs = metadata.crs
if crs is not None:
    if gdf.crs is None:
        logger.warning(f"No crs defined for GeoDataFrame, setting crs to {crs}")
        gdf.set_crs(crs)
    else:
        logger.warning(f"Reprojecting GeoDataFrame from {gdf.crs} to {crs}")
        gdf.to_crs(crs)
```

PR #1254, opened 2025-08-30
Close xarray before finishing DeferredFileClose

### LINKED ISSUE #778: Model.close() and/or Model as context manager

deltamarnix:
## Issue addressed

Fixes #778 

## Explanation

Xarray keeps Datasets open and does not support concurrent reading and writing to the same file. This gives conflicts in `Model.update()`. Therefore, all open Datasets need to be tracked within each component. Writing is first done to a temporary file next to the target. Reading can proceed as normal. After that, all open datasets are closed with `ModelComponent.cleanup()`. And the model can finish writing by calling `ModelComponent.finish_write()` right after to clean up all DeferredFileClose objects.

This solves the problem with open xarray Datasets. Bookkeeping of these Datasets is hard and also requires the caller of the code to be careful with keeping datasets open. If the user calls python functions like `ModelComponent.set`, hydromt doesn't know if it is a netcdf file underneath and if it should close it. Therefore, the end user is responsible in that case and we will log a PermissionError to the user during build or update.

This functionality does not yet change the way that the DataCatalog opens xarray objects and the bookkeeping there is not yet aligned. There will be similar cases where the original Dataset should actually be closed.

It is important to notice that retrieving a DataArray from a Dataset does not copy over the close handler (`Dataset._close`). Therefore, the original Dataset always needs to be in the bookkeeping somewhere. For the components it is possible to do this by adding them to `ModelComponent._open_datasets`.

## General Checklist

- [ ] Updated documentation
- [x] Updated changelog.rst

## Data/Catalog checklist


tests/model/test_model.py
LuukBlom, line 1: Can we add an integration test that checks whether the model uses this new defered fileclosing correctly when doing a build/update?

If you wrap your test in this:
```python
with tempfile.TemporaryDirectory() as tmpdir:
    ...
```

you get the same pytest tmpdir functionality, but it automatically deletes it at the end of the scope.

hydromt/_io/readers.py
```diff
@@ -868,116 +868,66 @@ def _parse_values(
...
 
-def _read_nc(
-    filepath: Path | str,
-    *,
-    mask_and_scale: bool = False,
-    single_var_as_array: bool = True,
-    load: bool = False,
-    **kwargs,
-) -> xr.Dataset:
+def read_nc(filepath: Path | str, **kwargs) -> xr.Dataset:
```
LuukBlom: Now that the `load` argument is no longer here, there is no reading happening here anymore. Should we rename this function to `open_nc` to communicate better that you have to close what it returns?

hydromt/_io/readers.py
```diff
@@ -868,116 +868,66 @@ def _parse_values(
...
-    if single_var_as_array and len(ds.data_vars) == 1:
-        (ds,) = ds.data_vars.values()
+        ds.set_close(org_close)
 
     # Return the dataset
     return ds
 
 
-def _read_ncs(
+def read_ncs(
```
LuukBlom: same comment: -> `open_ncs`

tests/model/test_model.py
deltamarnix: I have added some assertions to check whether the deferred closing was triggered.

hydromt/model/components/datasets.py
```diff
@@ -135,20 +120,20 @@ def read(
...
-            root=self.root.path,
-            single_var_as_array=single_var_as_array,
-            **kwargs,
-        )
+        ncs = open_ncs(filename_template, root=self.root.path, **kwargs)
         for name, ds in ncs.items():
-            self.set(data=ds, name=name)
+            self._open_datasets.append(ds)
+            if len(ds.data_vars) == 1:
+                (da,) = ds.data_vars.values()
```
aslowwriter, line 133: why the tuple?  is this intentional?

hydromt/model/components/spatialdatasets.py
```diff
@@ -151,20 +147,20 @@ def read(
...
-            root=self.root.path,
-            single_var_as_array=single_var_as_array,
-            **kwargs,
-        )
+        ncs = open_ncs(filename_template, root=self.root.path, **kwargs)
         for name, ds in ncs.items():
-            self.set(data=ds, name=name)
+            self._open_datasets.append(ds)
+            if len(ds.data_vars) == 1:
+                (da,) = ds.data_vars.values()
```
aslowwriter, line 160: Why this tuple?

tests/components/test_spatialdatasets_component.py
```diff
@@ -112,3 +113,23 @@ def test_add_raster_data_from_raster_reclass(tmpdir, demda, lulcda):
     )
 
     assert "roughness_manning" in model.maps.data
+
+
+def test_spatialdataset_updates_netcdf(raster_ds, tmpdir: Path, caplog):
+    write_path = Path(tmpdir) / "spatial_datasets/forcing.nc"
```
aslowwriter:
```suggestion
    write_path = Path(tmpdir) / "spatial_datasets" / "forcing.nc"
```
otherwise you end up with mixed file seperators on windows

tests/model/test_model.py
```diff
@@ -362,6 +361,15 @@ def test_model_build_update_with_data(tmpdir, demda, obsda, monkeypatch, caplog)
     assert any(
         log_record.message == "maps.set.name=elevtn2" for log_record in caplog.records
     )
+    assert any(
+        log_record.message
+        == f"Could not write to file {Path(tmpdir / 'spatial_datasets' / 'precip.nc').as_posix()}, deferring write"
```
aslowwriter: if `tmpdir` is already a path then is the cast necessary? I suppose it has to be otherwise this should fail, in which case I think you can remove it.

hydromt/model/components/datasets.py
deltamarnix: This is a simple way to pick out the first item of a list. Should I change it to da = next(iter(ds.data_vars.values()))?

hydromt/model/components/spatialdatasets.py
deltamarnix: This code actually came from read_nc, I just moved it to here.

hydromt/model/components/datasets.py
aslowwriter: no, I didn't know that. that's nice. you can leave it like that

tests/model/test_model.py
deltamarnix: No, tmpdir is not a path, it's a py.path.local. I'm changing the type hints everywhere in the code.

Issue #1243, opened 2025-08-19
Cannot open csv file from path anymore


hboisgon:
### HydroMT version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

Create a stations.csv file with this content

```
ID,Name,x,y,uparea
1001,Gauge1,11.9594,45.8925,3642
1002,Gauge2,12.3395,46.1492,2
1003,Gauge3,12.0785,46.1122,837
```

### Current behaviour

With geodataframe:

``` python
data_catalog.get_geodataframe("stations.csv")
```
returns
```
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    data_catalog.get_geodataframe("stations.csv")
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "D:\Repos\hydromt\hydromt_wflow\.pixi\envs\default\Lib\site-packages\hydromt\data_catalog\data_catalog.py", line 1465, in get_geodataframe
    gdf = source.read_data(
        mask=mask,
    ...<2 lines>...
        variables=variables,
    )
  File "D:\Repos\hydromt\hydromt_wflow\.pixi\envs\default\Lib\site-packages\hydromt\data_catalog\sources\geodataframe.py", line 72, in read_data
    gdf: gpd.GeoDataFrame = self.driver.read(
                            ~~~~~~~~~~~~~~~~^
        uris,
        ^^^^^
    ...<4 lines>...
        handle_nodata=handle_nodata,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "D:\Repos\hydromt\hydromt_wflow\.pixi\envs\default\Lib\site-packages\hydromt\data_catalog\drivers\geodataframe\pyogrio_driver.py", line 63, in read
    raise IOError(f"DataFrame from uri: '{_uri}' contains no geometry column.")
OSError: DataFrame from uri: 'D:/Repos/hydromt/hydromt_wflow/examples/test_stations.csv' contains no geometry column.
```

With dataframe
``` python
data_catalog.get_dataframe("stations.csv")
```
returns
```
Traceback (most recent call last):
  File "<python-input-5>", line 1, in <module>
    cat.get_dataframe("test_stations.csv")
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1262, opened 2025-09-04
added check for setting driver when using get_geodataframe with file path

Tjalling-dejong:
## Issue addressed
https://github.com/Deltares/hydromt/issues/1243

## Explanation

I added a check in the get_geodataframe method DataCatalog that checks if a driver argument is given and if not, tries to infer if the file should be read with a GeoDataFrameTable driver. If none of the conditions are met then the driver uses the fallback driver, that pyogrio.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/_abstract_base.py
+++ b/hydromt/_abstract_base.py
@@ -3,7 +3,8 @@
 from __future__ import annotations
 
 from abc import ABC, abstractmethod
-from typing import Any, Callable, ClassVar, Dict, Generator, List
+from collections.abc import Iterator
+from typing import Any, Callable, ClassVar, Self
 
 from pydantic import (
     BaseModel,
@@ -52,8 +53,8 @@ class AbstractBaseModel(BaseModel, ABC):
             cls._load_plugins()
 
             # Find which subclass to instantiate.
-            possible_subclasses: List[AbstractBaseModel] = list(
-                filter(lambda dr: dr.name == name, cls._find_all_possible_types())
+            possible_subclasses: list[AbstractBaseModel] = list(
+                filter(lambda dr: dr.name == name, cls.find_all_possible_types())
             )
             if len(possible_subclasses) == 0:
                 raise ValueError(f"Unknown 'name': '{name}'")
@@ -70,9 +71,9 @@ class AbstractBaseModel(BaseModel, ABC):
     def _serialize(
         self,
         nxt: SerializerFunctionWrapHandler,
-    ) -> Dict[str, Any]:
+    ) -> dict[str, Any]:
         """Add name to serialized result."""
-        serialized: Dict[str, Any] = nxt(self)
+        serialized: dict[str, Any] = nxt(self)
         serialized["name"] = self.name
         return serialized
 
@@ -83,16 +84,15 @@ class AbstractBaseModel(BaseModel, ABC):
         ...
 
     @classmethod
-    def _find_all_possible_types(cls) -> Generator[None, None, AbstractBaseModel]:
-        """Recursively generate all possible types for this object.
-
-        Logic relies on __bases__ and __subclass__() of the BaseDriver class,
-        which means that all drivers and plugins should be loaded in before.
-        """
-        # any concrete class is a possible type
+    def _iter_all_possible_types(cls) -> Iterator[type[Self]]:
+        """Recursively yield all possible sub types for this object."""
         if ABC not in cls.__bases__:
             yield cls
 
-        # continue looking for possible types in subclasses
         for subclass in cls.__subclasses__():
-            yield from subclass._find_all_possible_types()
+            yield from subclass._iter_all_possible_types()
+
+    @classmethod
+    def find_all_possible_types(cls) -> set[type[Self]]:
+        """Return all possible sub types for this object as a set."""
+        return set(cls._iter_all_possible_types())
--- a/hydromt/_abstract_base.py
+++ b/hydromt/_abstract_base.py
@@ -93,6 +93,6 @@ class AbstractBaseModel(BaseModel, ABC):
             yield from subclass._iter_all_possible_types()
 
     @classmethod
-    def find_all_possible_types(cls) -> set[type[Self]]:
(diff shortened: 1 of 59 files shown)


hydromt/data_catalog/data_catalog.py
```diff
@@ -1439,9 +1439,16 @@ def get_geodataframe(
             else:
                 if "provider" not in kwargs:
                     kwargs.update({"provider": "user"})
-                driver: str = kwargs.pop(
-                    "driver", GeoDataFrameSource._fallback_driver_read
-                )
+                if "driver" in kwargs:
+                    driver = kwargs.pop("driver")
+                # Check if file extension corresponds to the geodataframe_table driver
+                elif (
```
deltamarnix: I propose a bit more general purpose solution by not adding GeoDataFrameTableDriver to the data catalog. Instead, ask `GeoDataFrameSource` for `get_default_driver_read(filepath)` and let the data source return the driver name based on the extension.

hydromt/data_catalog/drivers/geodataframe/table_driver.py
```diff
@@ -26,6 +26,7 @@ class GeoDataFrameTableDriver(GeoDataFrameDriver):
     """Driver for reading in GeoDataFrames from tabular formats."""
 
     name: ClassVar[str] = "geodataframe_table"
+    _supported_extensions: ClassVar[Set[str]] = {".csv", ".xlsx", ".xls", ".parquet"}
```
deltamarnix: I like that this info is here. You can make it public, because you're going to need it in the data source. And since it's static data I would propose `SUPPORTED_EXTENSIONS`

hydromt/data_catalog/sources/geodataframe.py
```diff
@@ -266,3 +266,11 @@ def to_stac_catalog(
 
             stac_catalog.add_item(stac_item)
             return stac_catalog
+
+    @classmethod
+    def _infer_default_driver(cls, uri: str) -> GeoDataFrameDriver:
```
deltamarnix: Actually returns a `str`


```suggestion
    def _infer_default_driver(cls, uri: str) -> str:
```

hydromt/data_catalog/data_catalog.py
```diff
@@ -1439,9 +1439,7 @@ def get_geodataframe(
             else:
                 if "provider" not in kwargs:
                     kwargs.update({"provider": "user"})
-                driver: str = kwargs.pop(
-                    "driver", GeoDataFrameSource._fallback_driver_read
-                )
+                driver = kwargs.pop("driver", None)
                 name = basename(data_like)
                 source = GeoDataFrameSource(
                     name=name, uri=str(data_like), driver=driver, **kwargs
```
deltamarnix: You can remove the popping of the driver, because you don't do anything with it anymore. Doing it like this will always add a driver=None if the driver is not set. Let the kwargs handle it, I would say.

```suggestion
name = basename(data_like)
source = GeoDataFrameSource(
name=name, uri=str(data_like), **kwargs
```

hydromt/data_catalog/sources/data_source.py
```diff
@@ -89,8 +89,14 @@ def _validate_data_type(cls, data: Any) -> Any:
             if data_type := copy_data.pop("data_type", None):
                 if data_type != cls.data_type:
                     raise ValueError(f"'data_type' must be '{cls.data_type}'.")
+            if not copy_data.get("driver"):
+                copy_data["driver"] = cls._infer_default_driver(copy_data["uri"])
```
deltamarnix: According to GeoDataFrameSource, uri could be None? Consider using `copy_data.get("uri")` to be on the safe side.

hydromt/_abstract_base.py
```diff
@@ -83,16 +84,15 @@ def _load_plugins(cls):
...
 
-        # continue looking for possible types in subclasses
         for subclass in cls.__subclasses__():
-            yield from subclass._find_all_possible_types()
+            yield from subclass._iter_all_possible_types()
+
+    @classmethod
+    def find_all_possible_types(cls) -> set[type[Self]]:
+        """Return all possible sub types for this object as a set."""
+        return set(cls._iter_all_possible_types())
```
deltamarnix: What is wrong with returning an iterator? You only use it in for loops anyways.

hydromt/data_catalog/sources/geodataframe.py
```diff
@@ -266,3 +265,13 @@ def to_stac_catalog(
...
+
+    @classmethod
+    def _infer_default_driver(cls, uri: str | None = None) -> str:
+        if uri is None:
+            return cls._fallback_driver_read
+        _, extension = splitext(uri)
+        for driver in GeoDataFrameDriver.find_all_possible_types():
+            if extension in driver.SUPPORTED_EXTENSIONS:
+                return driver.name
+        return cls._fallback_driver_read
```
deltamarnix: I don't know if you like it, but it is possible to write this as a one-liner with [`next()`](https://docs.python.org/3/library/functions.html#next).


```suggestion
        return next(driver.name for driver in GeoDataFrameDriver.find_all_possible_types() if extension in driver.SUPPORTED_EXTENSIONS, cls._fallback_driver_read)
```

Tjalling-dejong: Good suggestion. I'll change it

hydromt/data_catalog/sources/data_source.py
Tjalling-dejong: Good point

hydromt/_abstract_base.py
Tjalling-dejong: I dont know. @LuukBlom  made the change

Tjalling-dejong:
> That's a nice addition! For now it is only implemented for GeoDataFrameDriver. Are there already other drivers where you would like to see this change applied?

Not for now. If it is requested by our users we could easily implement it.

PR #1272, opened 2025-09-16
Update logging

### LINKED ISSUE #778: Model.close() and/or Model as context manager

LuukBlom:
## Issue addressed

Fixes #778 

# Notes for Reviewer

The number of files changed is huge due to the addition of `get_hydromt_logger()` and the updates it causes throughout the codebase. 
EDIT: no longer the case since I removed `get_hydromt_logger`, but i did do global replaces so still quite some files changed.
Also, ruff automatically makes all imports absolute.

The relevant files to review for this PR are: `log.py`, `model.py`, `root.py` and `cli/main.py`

## Explanation

While reviewing [this PR](https://github.com/Deltares/hydromt/pull/1254), I came across the problem that log files were open, while all the other input/output files were nicely closed. 

After some investigation, I found the reason: `Modelroot.set` calls `_update_logger_filehandler`, which opens a logfile at `<model_root>/"hydromt.log"`.
This happened even if `_setuplog` had not been called, meaning there shouldn't be a log file and the only log handler is a `StreamHandler` (e.g. console). 

Non-trivial files changed are:
- `log.py`: Added `initialize_logging`, `add_filehandler`, `remove_filehandler`, and a context manager `to_file`.
- `__init__.py`: I added an `initialize_logging` call in the root level `__init__.py` so the console handler for hydromt will always be configured. (code here runs only once per interpreter session, even if it is imported many times.)
- `model.py`: Added `Model` as context manager, updated `build` and `update` to handle creating/managing the log files. 
- `model/root.py`: Removed log file handling, now only handles ModelMode & path. The copying over of the previous logfile is now the responsibility of the function that adds the filehandler. (e.g. update, build or cli)
- `cli/main.py`: cli functions now only set the log level and let the model create log files if possible. for `check` and `export` we still add a filehandler here since they dont call any Model function

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,7 @@ dependencies = [
 	"xarray",                 # ndim data arrays
 	"xmltodict",              # xml parser also used to read VRT
 	"xugrid>=0.9.0",          # xarray wrapper for mesh processing
-	"zarr>=2,<3",             # zarr
+	"zarr>=2.17,<4",          # zarr
 ]
 
 requires-python = ">=3.11"
--- a/pixi.lock
+++ b/pixi.lock
@@ -7483,7 +7483,7 @@ packages:
 - pypi: ./
   name: hydromt
   version: 1.2.0.dev0
-  sha256: fc2a631ae05c74ac7c206792ecf0bcd09265b5244a208f8727e520226b07f0f4
+  sha256: e5318e0c7f3a19d1427228c5a2c10f725af3a3b0998dd0b5ef8c8bc15ecfae77
   requires_dist:
   - affine
   - bottleneck
@@ -7518,7 +7518,7 @@ packages:
   - xarray
   - xmltodict
   - xugrid>=0.9.0
-  - zarr>=2,<3
+  - zarr>=2.17,<4
   - flit ; extra == 'dev'
   - mypy ; extra == 'dev'
   - pandas-stubs ; extra == 'dev'


data/catalogs/test_data_catalog.py
```diff
@@ -22,18 +22,18 @@
 
 import argparse
 import json
-import logging
 from os.path import exists
 
 from pydantic_core import ValidationError
 
 from hydromt import DataCatalog
+from hydromt._utils.log import get_hydromt_logger
```
deltamarnix: It would make sense to just move this to hydromt.log, so that plugin developers can also access it.

hydromt/_utils/caching.py
```diff
@@ -91,7 +90,7 @@ def _cache_vrt_tiles(
     vrt_uri: str,
     fs: Optional[AbstractFileSystem] = None,
     geom: Optional[gpd.GeoSeries] = None,
-    cache_dir: StrPath = SETTINGS.cache_root,
```
deltamarnix, line 94: If you don't like StrPath, then you should also remove it from the types 😸

hydromt/_utils/log.py
```diff
@@ -1,48 +1,87 @@
...
-            handler.close()
+def get_hydromt_logger(name: str | None = None) -> logging.Logger:
+    """Get a logger with the given name, or the root hydromt logger if name is None."""
+    if name is None:
+        return logging.getLogger("hydromt")
+    else:
+        return logging.getLogger(f"hydromt.{name}")
+
+
+ROOT_LOGGER = get_hydromt_logger()
```
deltamarnix: What would be the different between using `get_hydromt_logger()` and getting `ROOT_LOGGER`? Should ROOT_LOGGER be private so that other modules cannot simply access it?

hydromt/_utils/log.py
```diff
@@ -1,48 +1,87 @@
...
+    """
+    if path_or_filename is None:
+        match_str = None
+    elif isinstance(path_or_filename, str):
+        match_str = path_or_filename
+    else:
+        match_str = str(path_or_filename)
+
+    def _remove_matching(logger: logging.Logger):
+        for handler in list(logger.handlers):
```
deltamarnix: Are you `list()`-ing these because you can `removeHandler()` inside the function?

hydromt/_utils/log.py
```diff
@@ -1,48 +1,87 @@
...
+                    logger.removeHandler(handler)
+                    handler.flush()
+                    handler.close()
+
+    # main hydromt logger
+    _remove_matching(ROOT_LOGGER)
+
+    if include_children:
+        for name, lg in logging.Logger.manager.loggerDict.items():
+            if isinstance(lg, logging.Logger) and name.startswith("hydromt."):
```
deltamarnix: "hydromt" is used multiple times in this file. Is there a way to get the root name of the module and use that instead? To remove magic variables.

hydromt/_utils/log.py
```diff
@@ -51,31 +90,41 @@ def _setuplog(
...
+    logger: logging.Logger = ROOT_LOGGER,
+):
     """Add file handler to logger."""
-    if not os.path.isdir(os.path.dirname(path)):
-        os.makedirs(os.path.dirname(path))
-    isfile = os.path.isfile(path)
-    ch = FileHandler(path)
-    ch.setFormatter(Formatter(fmt))
+    path.parent.mkdir(parents=True, exist_ok=True)
+    ch = logging.FileHandler(path)
```
deltamarnix: What's `ch`? Consider a more meaningful variable name.

hydromt/cli/main.py
```diff
@@ -238,7 +236,9 @@ def build(
     -d /path/to/data_catalog.yml -v
     """  # noqa: E501
     log_level = max(10, 30 - 10 * (verbose - quiet))
-    log._setuplog(join(model_root, HYDROMT_LOG_PATH), log_level=log_level, append=False)
+    log._setuplog(
```
deltamarnix: _setuplog seems to be a non-private function.

hydromt/cli/main.py
```diff
@@ -332,7 +334,7 @@ def update(
     # logger
     mode = "r+" if model_root == model_out else "r"
     log_level = max(10, 30 - 10 * (verbose - quiet))
-    log._setuplog(join(model_out, HYDROMT_LOG_PATH), log_level=log_level)
+    log._setuplog(path=Path(model_out) / HYDROMT_LOG_PATH, log_level=log_level)
```
deltamarnix: It doesn't matter much, but using `/` does mean that you construct a Path object twice. Once for `Path()` and then for `/` you get a second. I try to use a single constructor and let pathlib handle the conjunction.

```suggestion
    log._setuplog(path=Path(model_out,  HYDROMT_LOG_PATH), log_level=log_level)
```

Also, `HYDROMT_LOG_PATH` seems to be a `HYDROMT_LOG_FILENAME`.

hydromt/_utils/log.py
```diff
@@ -1,48 +1,87 @@
...
+                    handler_path.name,
+                }:
+                    logger.removeHandler(handler)
+                    handler.flush()
+                    handler.close()
+
+    # main hydromt logger
+    _remove_matching(ROOT_LOGGER)
+
+    if include_children:
```
deltamarnix: Do you expect handlers on the child loggers? The log setup is only done on the root, right? After that there is only one handler that should be taken care of.

hydromt/_utils/log.py
```diff
@@ -51,31 +90,41 @@ def _setuplog(
         Whether to append (True) or overwrite (False) to a logfile at path,
         by default True
     """
-    main_logger: Logger = getLogger("hydromt")
-    _wait_and_remove_file_handlers(main_logger)
-    captureWarnings(True)
-    main_logger.setLevel(log_level)
-    console = StreamHandler(sys.stdout)
+    if path is not None:
+        remove_hydromt_file_handlers(path_or_filename=path)
```
deltamarnix: Is it still needed to close any old handlers? I think this could potentially go...

hydromt/_utils/log.py
```diff
@@ -51,31 +90,41 @@ def _setuplog(
...
+    logging.captureWarnings(True)
+    logger.setLevel(log_level)
+
+    # Remove all stream handlers
+    logger.handlers = [
+        h for h in logger.handlers if not isinstance(h, logging.StreamHandler)
+    ]
```
deltamarnix: Also here I wonder why it is needed to do all this cleaning.

hydromt/cli/main.py
```diff
@@ -420,7 +424,7 @@ def check(
     """  # noqa: E501
     # logger
     log_level = max(10, 30 - 10 * (verbose - quiet))
-    log._setuplog(join(".", HYDROMT_LOG_PATH), log_level=log_level)
+    log._setuplog(path=Path(".") / HYDROMT_LOG_PATH, log_level=log_level)
```
deltamarnix:
```suggestion
    log._setuplog(path=Path(HYDROMT_LOG_PATH), log_level=log_level)
```

hydromt/_utils/log.py
```diff
@@ -1,81 +1,135 @@
...
-            handler.close()
+def initialize_logging(
+    log_level: int = logging.INFO, formatter: logging.Formatter = _DEFAULT_FORMATTER
+) -> None:
+    """Initialize the hydromt root logger with a console handler, formatter and a log level.
 
+    Example
+    -------
+    ```python
+    from hydromt._utils.log import initialize_logging
```
deltamarnix: I'm not sure if this is the right format. It's rst, not md formatting.

hydromt/_utils/log.py
```diff
@@ -1,81 +1,135 @@
...
-        log message formatter
-    append : bool, optional
-        Whether to append (True) or overwrite (False) to a logfile at path,
-        by default True
+    # Initialize with log level (INFO)
+    initialize_logging(log_level=logging.INFO)
+
+    # Change log level to ERROR
+    logger = logging.getLogger("hydromt")
+    logger.setLevel(logging.ERROR)
```
deltamarnix: Why would someone call these functions directly and not use `set_log_level()`?

hydromt/_utils/log.py
```diff
@@ -1,81 +1,135 @@
...
+    _ROOT_LOGGER.setLevel(log_level)
+    if not _ROOT_LOGGER.hasHandlers():
+        console = logging.StreamHandler(sys.stdout)
+        console.setLevel(log_level)
+        console.setFormatter(formatter)
+        _ROOT_LOGGER.addHandler(console)
+    _ROOT_LOGGER.info(f"HydroMT version: {__version__}")
+
+
+def set_log_level(log_level: int, logger: logging.Logger = _ROOT_LOGGER) -> None:
```
deltamarnix: It feels rather useless to call this on other loggers than the root logger, since someone could just as well have called `logger.setLevel` themselves. This function is only useful when someone wants to change the level of the root logger.

hydromt/_utils/log.py
```diff
@@ -1,81 +1,135 @@
...
+    """Set the log level of the hydromt root logger.
+
+    This also affects all child loggers (e.g., ``hydromt.core``),
+    unless they have their own log level explicitly set.
+
+    Example
+    -------
+    ```python
```
deltamarnix: Also here, formatting is rst, not md.

hydromt/_utils/log.py
```diff
@@ -1,81 +1,135 @@
...
+    # Set log level to ERROR
+    set_log_level(logging.ERROR)
+    ```
+    """
+    logger.setLevel(log_level)
+
+
+def add_filehandler(
+    path: Path,
+    log_level: int | None = None,
```
deltamarnix, line 59: Since there are quite some parameters in this function, I would recommend using named parameters.
```suggestion
    path: Path,
    *,
    log_level: int | None = None,
```

hydromt/_utils/log.py
```diff
@@ -1,81 +1,135 @@
...
+    """Context manager that attaches a file handler for the duration of the block."""
+    path.parent.mkdir(parents=True, exist_ok=True)
+    if not append and path.exists():
+        path.unlink()
+    filehandler = logging.FileHandler(path)
+    if formatter is not None:
+        filehandler.setFormatter(formatter)
+    if log_level is not None:
+        filehandler.setLevel(log_level)
+    logger.addHandler(filehandler)
```
deltamarnix: Why not reuse `add_filehandler()`?

hydromt/_utils/log.py
```diff
@@ -1,81 +1,135 @@
...
+    try:
+        yield
+    finally:
+        logger.removeHandler(filehandler)
+        filehandler.close()
```
deltamarnix: Maybe reuse `remove_file_handler` too?

hydromt/_utils/log.py
```diff
@@ -1,81 +1,135 @@
...
+    """Shut down the logging system by removing all handlers and closing them."""
+    for h in _ROOT_LOGGER.handlers:
+        _ROOT_LOGGER.removeHandler(h)
+        h.close()
+
+    for child in _ROOT_LOGGER.manager.loggerDict.values():
+        if isinstance(child, logging.Logger):
+            for h in child.handlers:
+                child.removeHandler(h)
+                h.close()
```
deltamarnix: What about recursive children?
Or if these are all recursive children, do we still need line 125-127?

hydromt/cli/main.py
```diff
@@ -534,80 +522,78 @@ def export(
...
     log_level = max(10, 30 - 10 * (verbose - quiet))
-    log._setuplog(join(export_dest_path, HYDROMT_LOG_PATH), log_level=log_level)
-    logger.info(f"Output dir: {export_dest_path}")
-
-    if error_on_empty:
-        handle_nodata = NoDataStrategy.RAISE
-    else:
-        handle_nodata = NoDataStrategy.IGNORE
+    log.set_log_level(log_level=log_level)
+    log_path = Path(export_dest_path) / "hydromt_export.log"
```
deltamarnix: Unimportant, but I would still go for:


```suggestion
    log_path = Path(export_dest_path, "hydromt_export.log")
```

hydromt/model/model.py
```diff
@@ -328,49 +378,64 @@ def update(
...
+                    shutil.copyfile(
+                        self.root.path / "hydromt.log", model_out / "hydromt.log"
+                    )
+
+                self.read()
+                mode = "w+" if forceful_overwrite else "w"
+                self.root.set(model_out, mode=mode)
+
+        with log.to_file(
+            self.root.path / "hydromt.log", append=self.root.is_reading_mode()
```
deltamarnix: Maybe nice if append is always true in this case, because we might be updating in the same root.

deltamarnix (commented):
Really nice. Some minor requests for log.py, and it would be nice if we can keep the read logs while reading in `update()`.

hydromt/_utils/caching.py
LuukBlom: True, I dont like it, but I made these changes to fix a circular import error between typing and utils. Would rather not make these changes in this PR tbh

hydromt/_utils/log.py
```diff
@@ -1,81 +1,153 @@
...
-        if isinstance(handler, FileHandler):
-            # remove handler first, as otherwise new calls to the logger may open the
-            # same filename again.
-            logger.removeHandler(handler)
-            # wait for all messages to be processed
-            handler.flush()
-            # then close the handler
-            handler.close()
+def initialize_logging(
+    log_level: int = logging.INFO, formatter: logging.Formatter = _DEFAULT_FORMATTER
```
deltamarnix: I'm wondering if there would be any point anywhere in your code where `initialize_logging` will have other parameters. Since this function is already called on `import` of `hydromt`. So what's the deal with having these optional parameters?

hydromt/_utils/log.py
```diff
@@ -1,81 +1,153 @@
...
-    ch.setLevel(log_level)
-    logger.addHandler(ch)
-    if isfile:
+    logging.captureWarnings(True)
+    _ROOT_LOGGER.setLevel(log_level)
+    if not _ROOT_LOGGER.hasHandlers():
+        console = logging.StreamHandler(sys.stdout)
+        console.setLevel(log_level)
+        console.setFormatter(formatter)
+        _ROOT_LOGGER.addHandler(console)
```
deltamarnix, line 35: I saw that there is a function that might be super useful in this case. `logging.basicConfig()`. This can also set a file and the level in one call.

hydromt/model/model.py
deltamarnix: docs and type don't match for root. Would it be possible to make it `Path | None`, or are there places where an uri has to be set up? Otherwise I would leave this to the caller.

hydromt/model/model.py
```diff
@@ -328,49 +378,64 @@ def update(
...
+                for k, v in merged.items():
+                    logger.info(f"{step}.{k}={v}")
+                method(**merged)
+
+            # If there are any write options included in the steps,
+            # we don't need to write the whole model.
+            if write and not self._options_contain_write(steps):
+                self.write()
+
+            self.close()
```
deltamarnix, line 437: Maybe it's good to mention in the documentation of the Model how we expect people to work with closing the model and context managers.

In the case of build() and update(), the model is automatically closed. So there would be no need for a context manager.

hydromt/model/model.py
LuukBlom: I didnt change these lines? 
Anyways, Im in favour of using Path over str at all times, so I'll update

hydromt/model/model.py
LuukBlom: I dont think it really matters tbh, calling model.close() on a model without any open resources does nothing. 

But, there is a usage example/explanation in the docstring for `Model.__enter__` and `Model.__exit__`.
So I have added those two to the `General methods` section in the api docs.

pyproject.toml
```diff
@@ -43,7 +43,7 @@ dependencies = [
 	"xarray",                 # ndim data arrays
 	"xmltodict",              # xml parser also used to read VRT
 	"xugrid>=0.9.0",          # xarray wrapper for mesh processing
-	"zarr>=2,<3",             # zarr
+	"zarr>=3,<4",             # zarr
```
deltamarnix: I remember that upgrading zarr could cause API challenges. Is it needed for this PR? Also, remove version 2 might cause issues with other applications that are still using it. The last version 2 release was Apr 9, 2025.

hydromt/_utils/log.py
```diff
@@ -1,81 +1,138 @@
...
+    This function should be called once, at the start of the program.
+    For HydroMT, this is achieved by calling it in the `hydromt.__init__.py` file, so it is
+    automatically called when importing the hydromt package.
+    If the root logger has a handler before hydromt is imported, no new handler will be added.
+    """
+    logging.captureWarnings(True)
+    logging.basicConfig(
+        format=_LOG_FORMAT,
+        level=logging.INFO,
+    )
```
deltamarnix: Sorry, I didn't see that basicConfig was a module method instead of an object method. This implementation makes that ALL logging is set to INFO for ALL modules in python. It's likely that we don't want this, because we want to show users only INFO about things that are relevant for them, right?

pyproject.toml
LuukBlom: The reason I changed it was that warnings about zarr (which become errors) were raised in the tests.
So, I looked how much needed changing and its quite easy to upgrade to zarr v3 using the [migration guide](https://zarr.readthedocs.io/en/v3.1.1/user-guide/v3_migration.html) (its very short, not a whole lot changed) 

So that is what I updated in conftest.py.
There is zarr code in the drivers, but that doesn't change between v2 and v3.

Now, searching for zarr in the code, I see that I did miss `hydromt/data/src/era5_download_resample_convert.py::append_zarr()`, which is not called during testing, and which does have v2 code still, so I'll update that as well.

hydromt/_utils/log.py
LuukBlom: Agreed, I will revert then!

pyproject.toml
deltamarnix: I'm afraid I cannot allow this yet. We are supporting all versions of zarr that are still supported in the scientific python specification: https://scientific-python.org/specs/spec-0000/.

So if it's possible to keep it the way it is for now, I would like to keep it.

deltamarnix (changes requested):
Please don't upgrade zarr yet.

pyproject.toml
```diff
@@ -43,7 +43,7 @@ dependencies = [
 	"xarray",                 # ndim data arrays
 	"xmltodict",              # xml parser also used to read VRT
 	"xugrid>=0.9.0",          # xarray wrapper for mesh processing
-	"zarr>=3,<4",             # zarr
+	"zarr>=2,<3",             # zarr
```
deltamarnix: [This PR](https://github.com/Deltares/hydromt/pull/1278/commits) already upgraded zarr to 3.x. And now this PR is downgrading it, making the tests fail.

I would say let's look at the [spec 0](https://scientific-python.org/specs/spec-0000/) and support >=2.17,<4

LuukBlom: Not true, tests were failing due to another reason.

Also, I want to merge this PR and be done with it.

Lets make an issue for resolving the zarr versions and pick it up in another PR. It is not part of the [issue](https://github.com/Deltares/hydromt/issues/778).

PR #1276, opened 2025-09-23
Add DriverOptions class and its usage

### LINKED ISSUE #1252: Driver options are not documented
### HydroMT version checks

- [x] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/hydromt)

### Kind of issue

Docs are missing

https://deltares.github.io/hydromt/stable/guides/advanced_user/data_types.html#vector-data-geodataframe

Eg. for rasterdataset

Image

### Documentation problem

The user is able in his catalog or the get_data methods to pass specific driver options to the driver.read function. However what the function is and then which options are possible cannot be found anywhere. Worse in a few driver, some of the options are popped and therefore not passed to the final reading options. 

Eg, the raster_xarray_driver will pop `preprocess` and `ext_override` from options before passing the rest to `open_mfdataset`....

Because the options are driver specific, they cannot be documented at the ``data_catalog.get_data`` methods. But we should at least help the user with the syntax. Now we just say kwargs passed to ``DataSource`` constructor which is of course also not documented.... I would add example in the doctrings eg:

driver = {"name": "pyogrio", "options": {"layer": country}} and data_adapter = {"rename": {}, "unit_add": {}}

Finally the table above can then directly link to the specific driver documentation where I guess available options can then documented. Probelm is that the current page is pretty unreadable... See for example for [PyogryoDriver](https://deltares.github.io/hydromt/stable/_generated/hydromt.data_catalog.drivers.PyogrioDriver.html). The read pages are more readable (description of the arguments is missing btw) but because options is not a function argument but a property of the driver, this may not really be the place to document it? Example for [PyogryoDriver.read](https://deltares.github.io/hydromt/stable/_generated/hydromt.data_catalog.drivers.PyogrioDriver.read.html)

LuukBlom:
## Issue addressed

Fixes #1252 

## Explanation

To document the options in a nice way, I created a `DriverOptions` class;
a BaseModel that allows for arbitrary attribute setting, so essentially its a dict with some default key-value pairs and some helper functions to retrieve the kwargs for the underlying `open` function.

### autodoc_pydantic
I added [autodoc_pydantic](https://autodoc-pydantic.readthedocs.io/en/stable) to the docs extensions. 
This is automatically applied to all subclasses of `pydantic.BaseModel` that we `autosummary`.
Default configuration options are set in `conf.py`, but can be overwritten in place with [sphinx directives](https://autodoc-pydantic.readthedocs.io/en/stable/users/configuration.html)

This required some updating of the .rst files, but I think these changes only make them more readable.

For review, please have a look at the docs created from [this](https://github.com/Deltares/hydromt/pull/1276/commits/e1e7e0262556d13b21c3f1b25519e6484d6efca1) commit: [html.zip](https://github.com/user-attachments/files/22737853/html.zip)

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/data_catalog/drivers/__init__.py
+++ b/hydromt/data_catalog/drivers/__init__.py
@@ -1,18 +1,23 @@
 """Drivers responsible for reading and writing data."""
 
-from hydromt.data_catalog.drivers.base_driver import BaseDriver
+from hydromt.data_catalog.drivers.base_driver import BaseDriver, DriverOptions
 from hydromt.data_catalog.drivers.dataframe import DataFrameDriver, PandasDriver
 from hydromt.data_catalog.drivers.dataset import DatasetDriver, DatasetXarrayDriver
+from hydromt.data_catalog.drivers.dataset.xarray_driver import DatasetXarrayOptions
 from hydromt.data_catalog.drivers.geodataframe import (
     GeoDataFrameDriver,
     GeoDataFrameTableDriver,
     PyogrioDriver,
+)
+from hydromt.data_catalog.drivers.geodataframe.table_driver import (
+    GeoDataFrameTableOptions,
 )
 from hydromt.data_catalog.drivers.geodataset import (
     GeoDatasetDriver,
     GeoDatasetVectorDriver,
     GeoDatasetXarrayDriver,
 )
+from hydromt.data_catalog.drivers.geodataset.geodataset_driver import GeoDatasetOptions
 from hydromt.data_catalog.drivers.preprocessing import (
     harmonise_dims,
     remove_duplicates,


pyproject.toml
```diff
@@ -43,7 +43,7 @@ dependencies = [
 	"xarray",                 # ndim data arrays
 	"xmltodict",              # xml parser also used to read VRT
 	"xugrid>=0.9.0",          # xarray wrapper for mesh processing
-	"zarr>=3,<4",             # zarr
```
aslowwriter: Why this drop in zarr version?

Issue #1164, opened 2025-04-03
High impact sonarcube issues should be resolved or accepted


savente93:
### HydroMT version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

https://sonarcloud.io/project/issues?issueStatuses=OPEN%2CCONFIRMED&id=Deltares_hydromt

### Current behaviour

There are 126 issues of which sonarcube has marked 58 as high impact. 

### Desired behaviour

High impact issues are either resolved or accepted so that we know when new issues come up and can resolve them.

PR #1280, opened 2025-09-29
Reduce cognitive complexity for basin_mask

deltamarnix:
It was 109, now under 15.

I have removed this line:

```py
    elif (
        ds[basins_name].raster.nodata != 0 and ds[basins_name].raster.nodata is not None
    ):
        # TODO: Does this do anything? It seems to be a check only...
        ds[basins_name].where(ds[basins_name] != ds[basins_name].raster.nodata, 0)
```

Because the where is not stored anywhere. Tests are still passing.

## Issue addressed

Part of #1164

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst


deltamarnix:
@LuukBlom This one was the hardest to do and it seems that there are still some things to nitpick on the types. Do you think it's worth all the effort? It took me more than 4 hours. Perhaps it's better if we have to make changes to a function, first check the complexity and fix that with all the tests before we add more functionality to such a function.

PR #1291, opened 2025-10-17
fix: error on incorrect function signature data_catalog

### LINKED ISSUE #1287: Throw an error if time_tuple is used instead of time_range in get_data methods
### HydroMT version checks

### Reproducible Example

get_rasterdataset(time_tuple = (starttime, endtime))

### Current behaviour

time_tuple has been renamed to time_range for all the the get_data methods. However the change is not documented in the migration guide and barely visible in the changelog while this is a very important user function and not just for plugin developers!

And so we almost missed it in hydromt_wflow because no error or warning was thrown if time_tuple was still used.
And I think this is very error-prone for future v1 users.

### Desired behaviour

Throw an error if time_tuple is use mentioning to use time_range instead (just how any normal python function would do if you use an argument that does not exist..)
+ document that change better

### Additional context

_No response_

LuukBlom:
## Issue addressed

Fixes #1287 

## Explanation

The problem is that we have the entrypoint `DataCatalog.get_rasterdataset()` (and the other get_data functions), which had **kwargs as a part of the function signature.

This made it difficult because:
1. not possible to distinguish between kwargs meant for DataSource construction, and kwargs meant for passing to DataSource.read_data -> Driver.read
2. Subclasses of `DataSource` & `BaseDriver` extended the `DataSource.read_data` & `BaseDriver.read/-write` function signature which, due to how it has been set up, might be misinterpreted as kwargs as well when the subclass doesnt use all arguments.
3. If a user miss spells an argument, that we may or may not have updated (e.g. `time_tup` vs `time_range`), python will think it is actually a key word argument provided by the user, and pass it along, when it should actually error.

Having **kwargs in the function signature will always have that behaviour.

Solution:
Having a dict `open_kwargs` as a part of the signature, gives the user the same freedom for arbitrary options as the **kwargs, but it will also error when unexpected args are passed to the function.

## General Checklist

- [x] Updated documentation
- [x] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -35,7 +35,8 @@ Changed
 - ``NoDataStrategy`` is now part of of ``hydromt.error`` and thus public (#1277)
 - ``hydromt.typing`` and its functions are now public (underscore removed). (#1286)
 - remove ``io`` submodule and move ``readers`` and ``writers`` to the root module (#1308)
-- Function signatures of `DataCatalog` functions `get_rasterdataset`, `get_geodataframe`, `get_geodataset`, `get_dataset` and `get_dataframe` now have: ``source_kwargs`` (passed to `DataSource.__init__`), ``open_kwargs`` (passed to `DataSource.driver.read`) instead of just ``kwargs``, and ``time_tuple`` was renamed to ``time_range``. (#1291)
+- `write_nc` function no longer accepts kwargs, instead a dict `to_netcdf_kwargs` is used to pass arguments to `xarray.to_netcdf`. The function raises an error when "compute" is one of the arguments.
+- Function signatures of `DataCatalog` functions `get_rasterdataset`, `get_geodataframe`, `get_geodataset`, `get_dataset` and `get_dataframe` now have: ``source_kwargs`` (passed to `DataSource.__init__`) instead of just ``kwargs``, and ``time_tuple`` was renamed to ``time_range``. (#1291)
 
 Fixed
 -----
--- a/hydromt/_utils/log.py
+++ b/hydromt/_utils/log.py
@@ -33,7 +33,6 @@ def initialize_logging() -> None:
         console.setLevel(logging.INFO)
         console.setFormatter(_DEFAULT_FORMATTER)
         _ROOT_LOGGER.addHandler(console)
-    _ROOT_LOGGER.info(f"HydroMT version: {__version__}")
 
 
 def set_log_level(log_level: int) -> None:
@@ -89,6 +88,10 @@ def _add_filehandler(
     else:
         logger.debug(f"Writing log messages to new file {path}.")
     return filehandler
+
+
+def log_version() -> None:
+    _ROOT_LOGGER.info(f"HydroMT version: {__version__}")
 
 
 @contextmanager
--- a/hydromt/cli/main.py
+++ b/hydromt/cli/main.py
@@ -237,6 +237,7 @@ def build(
     """  # noqa: E501
     log_level = max(10, 30 - 10 * (verbose - quiet))
     log.set_log_level(log_level=log_level)
+    log.log_version()
     # Model.build will manage the filehandlers and logging
 
     modeltype, kwargs, steps = read_workflow_yaml(config, modeltype=model)
@@ -318,6 +319,7 @@ def update(
     # logger
     log_level = max(10, 30 - 10 * (verbose - quiet))
     log.set_log_level(log_level=log_level)
+    log.log_version()
     # Model.update will manage the filehandlers and logging
 
     mode = "r+" if model_root == model_out else "r"
@@ -450,6 +452,7 @@ def check(
     log_path = Path.cwd() / "hydromt_check.log"
     log.set_log_level(log_level=log_level)
     with log.to_file(log_path):
+        log.log_version()
         results = []
         for cat_path in data:
             results.append(_validate_catalog(Path(cat_path), format, upgrade))
@@ -526,6 +529,7 @@ def export(
     log.set_log_level(log_level=log_level)
     log_path = Path(export_dest_path, "hydromt_export.log")
     with log.to_file(log_path):
+        log.log_version()
         if error_on_empty:
             handle_nodata = NoDataStrategy.RAISE
         else:
--- a/hydromt/model/components/datasets.py
+++ b/hydromt/model/components/datasets.py
@@ -1,7 +1,7 @@
 """Xarrays component."""
 
 import logging
-from typing import TYPE_CHECKING, Union, cast
+from typing import TYPE_CHECKING, Any, Union, cast
 
 import xarray as xr
 from pandas import DataFrame
@@ -143,7 +143,7 @@ class DatasetsComponent(ModelComponent):
         gdal_compliant: bool = False,
         rename_dims: bool = False,
         force_sn: bool = False,
-        **kwargs,
+        to_netcdf_kwargs: dict[str, Any] | None = None,
     ) -> None:
         """Write dictionary of xarray.Dataset and/or xarray.DataArray to netcdf files.
 
@@ -161,17 +161,17 @@ class DatasetsComponent(ModelComponent):
         filename: str, optional
             filename relative to model root and should contain a {name} placeholder
             Can be a relative path.
-        gdal_compliant: bool
+        gdal_compliant: bool, optional
             If True, convert xarray.Dataset and/or xarray.DataArray to gdal compliant
             format using :py:meth:`~hydromt.raster.gdal_compliant`
-        rename_dims: bool
+        rename_dims: bool, optional
             If True, rename x_dim and y_dim to standard names depending on the CRS
             (x/y for projected and lat/lon for geographic). Only used if
             ``gdal_compliant`` is set to True. By default, False.
-        force_sn: bool
+        force_sn: bool, optional
             If True, forces the dataset to have South -> North orientation. Only used
             if ``gdal_compliant`` is set to True. By default, False.
-        **kwargs:
+        to_netcdf_kwargs: dict, optional
             Additional keyword arguments that are passed to the `to_netcdf`
             function.
         """
@@ -183,7 +183,8 @@ class DatasetsComponent(ModelComponent):
             )
             return
 
-        kwargs.setdefault("engine", "netcdf4")
+        to_netcdf_kwargs = to_netcdf_kwargs or {}
+        to_netcdf_kwargs.setdefault("engine", "netcdf4")
 
         filename = filename or self._filename
 
@@ -200,7 +201,7 @@ class DatasetsComponent(ModelComponent):
                 rename_dims=rename_dims,
                 force_sn=force_sn,
                 force_overwrite=self.root.mode.is_override_mode(),
-                **kwargs,
+                to_netcdf_kwargs=to_netcdf_kwargs,
             )
             if close_handle is not None:
                 self._deferred_file_close_handles.append(close_handle)
--- a/hydromt/model/components/grid.py
+++ b/hydromt/model/components/grid.py
@@ -1,7 +1,7 @@
 """Grid Component."""
 
 import logging
-from typing import TYPE_CHECKING, Dict, Optional, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union, cast
 
 import geopandas as gpd
 import numpy as np
@@ -146,7 +146,7 @@ class GridComponent(SpatialModelComponent):
         gdal_compliant: bool = False,
         rename_dims: bool = False,
         force_sn: bool = False,
-        **kwargs,
+        to_netcdf_kwargs: dict[str, Any] | None = None,
     ) -> None:
         """Write model grid data to netcdf file at <root>/<fn>.
 
@@ -165,7 +165,7 @@ class GridComponent(SpatialModelComponent):
         force_sn: bool, optional
             If True and gdal_compliant, forces the dataset to have
             South -> North orientation.
-        **kwargs : dict
+        to_netcdf_kwargs : dict, optional
             Additional keyword arguments to be passed to the `write_nc` method.
         """
         self.root._assert_write_mode()
@@ -190,7 +190,7 @@ class GridComponent(SpatialModelComponent):
             rename_dims=rename_dims,
             force_overwrite=self.root.mode.is_override_mode(),
             force_sn=force_sn,
-            **kwargs,
+            to_netcdf_kwargs=to_netcdf_kwargs,
         )
         if close_handle is not None:
             self._deferred_file_close_handles.append(close_handle)
--- a/hydromt/model/components/spatial.py
+++ b/hydromt/model/components/spatial.py
@@ -2,7 +2,7 @@
 
 import logging
 from abc import ABC, abstractmethod
-from typing import TYPE_CHECKING, Dict, Optional, Tuple, cast
+from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, cast
 
 import geopandas as gpd
 from geopandas import GeoDataFrame
@@ -89,7 +89,7 @@ class SpatialModelComponent(ModelComponent, ABC):
         filename: Optional[str] = None,
         *,
         to_wgs84: bool = False,
-        **write_kwargs,
+        to_file_kwargs: dict[str, Any] | None = None,
     ) -> None:
         """Write the model region to file.
 
@@ -103,7 +103,7 @@ class SpatialModelComponent(ModelComponent, ABC):
             The filename to write the region to. If None, the filename provided at initialization is used.
         to_wgs84 : bool
             If True, the region is reprojected to WGS84 before writing.
-        **write_kwargs:
+        to_file_kwargs: dict, optional
             Additional keyword arguments passed to the `geopandas.GeoDataFrame.to_file` function.
         """
         self.root._assert_write_mode()
@@ -130,7 +130,7 @@ class SpatialModelComponent(ModelComponent, ABC):
             self.region,
             file_path=full_path,
             to_wgs84=to_wgs84,
-            **write_kwargs,
+            to_file_kwargs=to_file_kwargs,
         )
 
     def test_equal(self, other: ModelComponent) -> Tuple[bool, Dict[str, str]]:
--- a/hydromt/model/components/spatialdatasets.py
+++ b/hydromt/model/components/spatialdatasets.py
@@ -2,7 +2,7 @@
 
 import logging
 from pathlib import Path
-from typing import TYPE_CHECKING, Dict, List, Optional, Tuple, Union, cast
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union, cast
 
 import xarray as xr
 from geopandas import GeoDataFrame
@@ -170,7 +170,7 @@ class SpatialDatasetsComponent(SpatialModelComponent):
         gdal_compliant: bool = False,
         rename_dims: bool = False,
         force_sn: bool = False,
-        **kwargs,
+        to_netcdf_kwargs: dict[str, Any] | None = None,
     ) -> None:
         """Write dictionary of xarray.Dataset and/or xarray.DataArray to netcdf files.
 
@@ -197,7 +197,7 @@ class SpatialDatasetsComponent(SpatialModelComponent):
         force_sn: bool, optional
             If True, forces the dataset to have South -> North orientation. Only used
             if ``gdal_compliant`` is set to True. By default, False.
-        **kwargs:
+        to_netcdf_kwargs: dict, optional
             Additional keyword arguments that are passed to the `to_netcdf`
             function.
         """
@@ -209,7 +209,8 @@ class SpatialDatasetsComponent(SpatialModelComponent):
             )
             return
 
-        kwargs.setdefault("engine", "netcdf4")
+        to_netcdf_kwargs = to_netcdf_kwargs or {}
+        to_netcdf_kwargs.setdefault("engine", "netcdf4")
         filename = filename or self._filename
 
         for name, ds in self.data.items():
@@ -226,7 +227,7 @@ class SpatialDatasetsComponent(SpatialModelComponent):
                 rename_dims=rename_dims,
                 force_sn=force_sn,
                 force_overwrite=self.root.mode.is_override_mode(),
-                **kwargs,
+                to_netcdf_kwargs=to_netcdf_kwargs,
             )
             if close_handle is not None:
                 self._deferred_file_close_handles.append(close_handle)
--- a/hydromt/model/components/vector.py
+++ b/hydromt/model/components/vector.py
@@ -3,7 +3,7 @@
 
 import logging
 from os.path import basename, dirname, isfile, join
-from typing import TYPE_CHECKING, Optional, Union, cast
+from typing import TYPE_CHECKING, Any, Optional, Union, cast
 
 import geopandas as gpd
 import numpy as np
@@ -250,7 +250,8 @@ class VectorComponent(SpatialModelComponent):
         geometry_filename: Optional[str] = "vector/vector.geojson",
         *,
         ogr_compliant: bool = False,
-        **kwargs,
+        to_netcdf_kwargs: dict[str, Any] | None = None,
+        to_gdf_kwargs: dict[str, Any] | None = None,
     ) -> None:
         """Write model vector to combined netcdf and geojson files.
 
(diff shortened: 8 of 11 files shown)


savente93:
I haven't reviewed the full PR yet, and I like the direction it's going in, but looking at the code reminded me that we already used to have a solution like this, and I just want to understand where we lost that and if it was for a particular reason. That may take a bit of time, but I'll keep you up to date

hydromt/data_catalog/drivers/geodataframe/table_driver.py
```diff
@@ -42,6 +44,7 @@ class GeoDataFrameTableDriver(GeoDataFrameDriver):
 
     name: ClassVar[str] = "geodataframe_table"
     SUPPORTED_EXTENSIONS: ClassVar[set[str]] = {".csv", ".xlsx", ".xls", ".parquet"}
+    supports_writing: ClassVar[bool] = False
```
deltamarnix: Could you elaborate on this class variable? The default of BaseDriver is already False, and there is no redefinition in GeoDataFrameDriver.

hydromt/data_catalog/drivers/geodataframe/geodataframe_driver.py
```diff
@@ -22,29 +23,70 @@ def read(
...
     def write(
         self,
-        path: StrPath,
-        gdf: gpd.GeoDataFrame,
-        **kwargs,
+        path: Path | str,
+        data: gpd.GeoDataFrame,
+        *,
+        write_kwargs: dict[str, Any] | None = None,
     ) -> str:
```
deltamarnix: mypy complains that this return type doesn't match the one from the BaseDriver. The string is returned by the pyogriodriver. And I also see that this destination path is being used in dataframe.py. Should we update the base class' signature?

hydromt/data_catalog/drivers/geodataframe/table_driver.py
```diff
@@ -52,13 +55,52 @@ def read(
...
+            If multiple URIs are provided.
+
+        Warning
+        -------
+        The `mask` and `variables` parameters are not used directly in this driver, but are included
+        for consistency with the GeoDataFrameDriver interface.
+        """
+        _warn_on_unused_kwargs(
+            self.__class__.__name__, {"mask": mask, "variables": variables}
+        )
```
deltamarnix, line 101: Should we actually warn when these parameters are not used? The call from GeoDataFrameSource will always pass the mask and the variables, so that means using tables will always result in warnings. Warnings in your model mean that the model might not be correct. I don't think this is a useful warning.

hydromt/data_catalog/drivers/geodataset/geodataset_driver.py
deltamarnix, line 42: Given that these functions have different signatures than then one in BaseDriver (.write() and .read()), I am wondering if those functions are ever called. Perhaps those base functions should actually be removed, because we don't rely on them at all. We never call a BaseDriver's .write(), we only call very specific DataFrameDriver.write() or others.

hydromt/data_catalog/drivers/raster/raster_xarray_driver.py
```diff
@@ -77,38 +76,78 @@ def read(
...
+        but are included for consistency with the GeoDataFrameDriver interface.
         """
         _warn_on_unused_kwargs(
             self.__class__.__name__,
             {
+                "mask": mask,
+                "variables": variables,
                 "zoom": zoom,
+                "chunks": chunks,
+                "metadata": metadata,
```
deltamarnix, line 133: Same here. I don't think we want so many warnings in the user's logs.

hydromt/data_catalog/sources/rasterdataset.py
```diff
@@ -217,9 +219,13 @@ def _detect_time_range(
...
             ds = self.read_data()
+
+        if not np.issubdtype(ds[ds.raster.time_dim].dtype, np.datetime64):
+            ds = ds.convert_calendar("standard")
+
         return TimeRange(
-            start=ds[ds.raster.time_dim].min().values,
-            end=ds[ds.raster.time_dim].max().values,
+            start=ds[ds.raster.time_dim].values.min(),
+            end=ds[ds.raster.time_dim].values.max(),
```
deltamarnix, line 228: What makes you do this change?

hydromt/gis/raster.py
```diff
@@ -84,15 +85,29 @@ def get_attrs(self, key, placeholder=None) -> Any:
...
-        return self.get_attrs("time_dim")
+            return None
+        elif len(tdims) == 1:
+            self.set_attrs(time_dim=tdims[0])
+            return tdims[0]
+        else:
+            raise ValueError(
+                f"Multiple time dimensions found: {tdims}. "
+                "Set 'time_dim' attribute manually to resolve."
+            )
```
deltamarnix, line 111: I don't see this reflected in new tests.

savente93:
> The longer I think about this whole inheritance structure, the less sense it makes. It seems like we are not looking for generic solutions, we are looking for very specific solutions. Drivers require their own arguments, source require their own driver. I wonder if we should try to make it less generic and more specific, determining the strategy of the data that is about to be read/written up front.

I don't intend this as an accusation/defensive comment, but this is not the first time I've seen you say this. Would you like a design session to reevaluate (and perhaps rework) this architecture?

hydromt/data_catalog/drivers/geodataset/geodataset_driver.py
LuukBlom: I dont really understand, they are `abstractmethods`, so they are indeed never called and must be overwritten in the subclasses. Putting the functions in the base classmakes sense to me since all subclasses have these functions, so why not enfore that by putting it in the base.

Note: 
In python, we can only extend the function signature in subclasses by adding a key word argument (including default value). Changing the order, number, or names of the positional arguments is not allowed and will error (which is what we want in this case). 
Python's idea is to still support Liskov, but also allow extensions in subclasses

hydromt/data_catalog/sources/rasterdataset.py
LuukBlom: I got an error when it was determining this, and the return types are different:
- `ds[ds.raster.time_dim].values.min()`  -- this was datetime
- `ds[ds.raster.time_dim].min().values` -- this was some kind of array

hydromt/data_catalog/drivers/raster/raster_xarray_driver.py
LuukBlom: I dont like it either, but didnt want to start scope creeping, so I just made everything consistent.

I think the inheritance structure is fine for most cases, but perhaps we should get rid of the intermediate base classes like `RasterDatasetDriver`. That would not force the function signatures to be the same, so we can remove the unused args and the warning, we would have to accomodate this change in the get_rasterdataset function somewhere. 

Lets discuss and make a decision

hydromt/data_catalog/drivers/geodataframe/table_driver.py
LuukBlom: good catch

hydromt/data_catalog/drivers/geodataframe/geodataframe_driver.py
LuukBlom: Good catch, i have updated.

hydromt/data_catalog/drivers/geodataframe/table_driver.py
LuukBlom: Well, the defaults for these arguments is `None`.

This is the warning function:
```python
def _warn_on_unused_kwargs(obj_name: str, name_value_dict: Dict[str, Any]):
    """Warn on unused kwargs."""
    for name, value in name_value_dict.items():
        if value is not None:
            logger.warning(
                f"object: {obj_name} does not use kwarg {name} with value {value}."
            )
```

So there will only be a warning if the user explicitely provides one of these parameters to the function, when using this driver.

hydromt/data_catalog/drivers/geodataset/geodataset_driver.py
aslowwriter: sorry for the drive by comment but this is not true if I understand your comment correctly: 

image

deltamarnix: Python is very flexible and actually lets you do whatever. Anything that is redefined in the subclass is a complete overwrite of whatever was in the base class.
mypy complains when you are not doing liskov and that's why I think it's better to not have the abstract method in the first place.

What I wanted to say regarding the base class.
There is no point in our code that has this:

```py
driver: BaseDriver = get_driver()
driver.write(**driver_kwargs_for_write)
```

This is also very anti-Liskov, because Liskov expects that the arguments of BaseDriver.write() should be the same as GeoDataframeDriver.write().

Instead, we see very specific: `GeoDataFrameDriver().write(specific, arguments, here)`, and I think it could be even more specific with a `PyOgrioDriver().write(ogrio, arguments, here)`.

hydromt/gis/raster.py
LuukBlom: True, there were no tests for this property, but I have added them now.

hydromt/data_catalog/drivers/geodataset/geodataset_driver.py
LuukBlom: Huh interesting example Sam, I couldve sworn you'd get an error when doing that. 

> This is also very anti-Liskov, because Liskov expects that the arguments of BaseDriver.write() should be the same as GeoDataframeDriver.write().

Well, The BaseDriver has all the arguments that are valid for all of its subclasses (if you dont you get mypy/pyright errors as well), then subclasses of BaseDriver extend using arguments with a default, so we could use the BaseDrivers signature and call it on any other Driver class. I know now this isnt enforced by python, but in this case this is true for this implementation. 

I do see the point that we never actually call `BaseDriver.read()`, so that can indeed be a bit misleading. 
Noteworthy though: There is quite some duplicated code in the `DataSource` (look at all `DataSource` subclasses -> `read_data()` and `write_data()`, that I think could be removed if we come up with a good idea. 
Essentially, every `DataSource.read_data()` function calls these functions (with some small modifications for each)

```python
...
def read_data()
    self._mark_as_used()
    self._log_start_read_data()
    time_range: TimeRange = self.data_adapter._to_source_timerange(time_range)
    uris: List[str] = self.uri_resolver.resolve(...)
    data = self.driver.read(...)
    return self.data_adapter.transform(...)

```

LuukBlom: So, to conlude and write down what we decided.

For this PR we will not change more behaviour, only implement the error on `time_tuple` and the `source_kwargs` and `open_kwargs`. 

In the future, we should have a design session to think of a solution and perhaps remove all of the DataSource classes, so we can actually use BaseDriver as an interface. We could enforce exactly the same function signature for all `Driver.read` and then have some other way of adding the optional arguments currently added in the `Driver.read` functions. 

https://github.com/Deltares/hydromt/issues/1295

hydromt/data_catalog/drivers/raster/raster_xarray_driver.py
LuukBlom: https://github.com/Deltares/hydromt/issues/1295

hydromt/data_catalog/drivers/dataframe/pandas_driver.py
```diff
@@ -52,8 +73,8 @@ def read(
         else:
             uri = uris[0]
             extension: str = uri.split(".")[-1]
-            kwargs_for_open = kwargs_for_open or {}
-            kwargs = self.options.get_kwargs() | kwargs_for_open
+            open_kwargs = open_kwargs or {}
+            kwargs = self.options.get_kwargs() | open_kwargs
```
deltamarnix: Did you also look at the DriverOptions? It seems that a lot of these drivers also have their own DriverOptions child class that contains information about what driver kwargs are available. See rasterio_driver.RasterioOptions (KWARGS_FOR_OPEN = {"mosaic_kwargs"}).

I wonder if and how these are used and combined in one's yaml config. Should those also be renamed? Or are we okay the way it is? Let's do some research.

hydromt/data_catalog/drivers/geodataset/geodataset_driver.py
```diff
@@ -45,34 +46,63 @@ def read(
...
         *,
         handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
-        kwargs_for_open: dict[str, Any] | None = None,
+        open_kwargs: dict[str, Any] | None = None,
         mask: Geom | None = None,
         predicate: Predicate = "intersects",
-        variables: list[str] | None = None,
-        time_range: TimeRange | None = None,
         metadata: SourceMetadata | None = None,
     ) -> xr.Dataset | None:
```
deltamarnix: This return type is optional, but the one from vector driver is not. You can see a mypy error in GeoDatasetSource that transform(ds,...) (ds could be None)

hydromt/data_catalog/drivers/dataframe/pandas_driver.py
LuukBlom: Yes I did. I implemented the DriverOptions classes to be able to document them better in https://github.com/Deltares/hydromt/pull/1276

The main difference is that DriverOptions are defined in the DataCatalog.yml, while the open_kwargs are runtime arguments from DataCatalog functions.

How `DriverOptions` works:
- Each Driver has a an attr `options: DriverOptions`, which is a `BaseModel` with `extra="allow"` to allow any arbitrary insertions like a dict (it used to be a dict, but that was very difficult to document nicely).
- In the previous implementation, some keywords were not passed to `open` because they were popped from the kwargs. So, not the entire dict is passed.
- Therefore DriverOptions base class has the main public function: `get_kwargs`, and a `ClassVar KWARGS_FOR_OPEN`. to mimick `kwargs.pop()` behaviour
- `get_kwargs` will return all attributes not defined at the class level (e.g. set by the user), but also include the attributes mentioned in `KWARGS_FOR_OPEN` (any keywords that are in this set, are always passed as kwargs well.). 

`DriverOptions.get_kwargs` is always combined with the `open_kwargs`, in the order that is shown here (so `open_kwargs` overwrites `DriverOptions.get_kwargs` if they are trying to set the same keyword), and then passed to the `open` function of the driver. 

Can you be more specific in what you are worried about?

deltamarnix: I was wondering if it should be possible to overwrite, or if these sets of arguments are actually different. Why would one set it in both the data catalogue and in the call?
And indeed I was wondering why one has priority over the other.

hydromt/data_catalog/drivers/geodataset/geodataset_driver.py
LuukBlom: Strange, mypy is enabled in my vscode but i dont get any errors

hydromt/data_catalog/drivers/dataframe/pandas_driver.py
LuukBlom: Essentially, the driver options are the defaults to be used for a DataSource when reading data. These are bound to the source.
While the runtime arguments are used to overwrite the defaults and are user defined when DataCatalog.get_x is called.

I think the advantage of supporting both, is clear; 
- we don't require users to know for each source the specific kwargs that might be required for the open func (because we (let them) define it in the data_catalog.yml).
- we still allow for flexible overwrites on a per-case basis
- we separate the data_catalog.yml and workflow.yml. (perhaps they are using some artifact data and cannot easily change the catalog.yml, but they can easily change their workflow.yml)

Currently, people using the DataCatalog to read from file paths directly will provide one of 2 to the `DataCatalog.get_x` functions:
- `source_kwargs`: a dict, used to create a new `DataSource` and add it to the catalog. (Now that Im looking at it, `source_kwargs` are only used if the a new `DataSource` is actually created. Which only happens if `data_like` is a `Path` or a `str` that isnt matched to any `DataSource` already in the `DataCatalog`). So, passing `source_kwargs` to `data_catalog.get_rasterdataset("era5", source_kwargs={...})` silently ignores `source_kwargs` when `era5` already exists in the datacatalog. 
- `open_kwargs`: dict that is always passed directly to open funcs, with or without creating a new DataSource.

Im not against changing the behaviour at all, since its a bit confusing. 
I think, with some refactoring, it could be done with just the `source_kwargs`, which might also make the function signatures of the drivers more similar, but I would prefer to do that in a different PR. 
I think this PR makes everything consistent and this idea seems like it will be a large PR itself.

.vscode/settings.json
```diff
@@ -43,5 +43,6 @@
     "sonarlint.connectedMode.project": {
         "connectionId": "deltares",
         "projectKey": "Deltares_hydromt"
-    }
+    },
+    "python.analysis.typeCheckingMode": "standard"
```
deltamarnix: This repository is using mypy, instead of pyright for its type checking. Were you missing the mypy plugin, or is it perhaps disabled in your repository? I would recommend reverting this line.

hydromt/data_catalog/adapters/dataset.py
```diff
@@ -65,6 +65,7 @@ def transform(
             return ds
         except NoDataException:
             exec_nodata_strat("No data to export", strategy=handle_nodata)
+            return None  # handle_nodata == ignore
```
deltamarnix: The comment is confusing for me. The code speaks for itself.

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -130,18 +133,49 @@ def read(
...
+            The path where the dataset was written.
+
+        Raises
+        ------
+        ValueError
+            If the provided file extension is unsupported.
         """
-        ext = splitext(path)[-1]
+        if isinstance(path, str):
+            path = Path(path)
```
deltamarnix:
```suggestion
        path = Path(path)
```

You can always do this, even when it's a Path.

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -130,18 +133,49 @@ def read(
...
+        ext = path.suffix
         if ext == ".zarr":
-            ds.to_zarr(path, **kwargs)
+            data.to_zarr(path, **(write_kwargs or {}))
         elif ext in [".nc", ".netcdf"]:
-            ds.to_netcdf(path, **kwargs)
+            data.to_netcdf(path, **(write_kwargs or {}))
         else:
             raise ValueError(f"Unknown extension for DatasetXarrayDriver: {ext} ")
```
deltamarnix, line 181: Would it make sense to just set write_kwargs once?


```suggestion
        write_kwargs = write_kwargs or {}
        ext = path.suffix
if ext == ".zarr":
            data.to_zarr(path, **write_kwargs)
elif ext in [".nc", ".netcdf"]:
            data.to_netcdf(path, **write_kwargs)
else:
raise ValueError(f"Unknown extension for DatasetXarrayDriver: {ext} ")
```

hydromt/data_catalog/drivers/geodataset/xarray_driver.py
```diff
@@ -115,17 +131,44 @@ def read(
...
-        ext = splitext(path)[-1]
+        if isinstance(path, str):
+            path = Path(path)
+        ext = path.suffix
         if ext == _ZARR_EXT:
-            ds.vector.to_zarr(path, **kwargs)
+            data.vector.to_zarr(path, **(write_kwargs or {}))
         elif ext in _NETCDF_EXT:
-            ds.vector.to_netcdf(path, **kwargs)
+            data.vector.to_netcdf(path, **(write_kwargs or {}))
```
deltamarnix: Same here

hydromt/data_catalog/drivers/raster/raster_xarray_driver.py
```diff
@@ -77,48 +74,86 @@ def read(
...
-        preprocessor: Callable | None = self.options.get_preprocessor()
+        preprocessor = self.options.get_preprocessor()
 
         # Check for the override flag
         first_ext = self.options.get_ext_override(uris)
 
         # When is zarr, open like a zarr archive
-        kwargs_for_open = kwargs_for_open or {}
+        open_kwargs = open_kwargs or {}
+        kwargs = self.options.get_kwargs() | open_kwargs
```
deltamarnix: Maybe be careful with a local variable called kwargs. Might someone every decide to change the signature again, this local variable might override things.
Maybe just arguments? Or overwrite open_kwargs once more?

hydromt/data_catalog/drivers/geodataframe/geodataframe_driver.py
```diff
@@ -22,29 +23,66 @@ def read(
...
+            or any geometry-like object depending on driver support.
+        variables : str | list[str] | None, optional
+            Optional variable(s) or column(s) to read from the source file.
+
+        Returns
+        -------
+        gpd.GeoDataFrame
+            The loaded geospatial data.
+        """
         ...
```
deltamarnix, line 55: Would it be useful to have throw NotImplementedError in all these driver cases where it's now ...? Or will that generate more problems?

hydromt/data_catalog/sources/dataset.py
```diff
@@ -66,22 +71,20 @@ def read_data(
...
             handle_nodata=handle_nodata,
         )
 
         ds: xr.Dataset = self.driver.read(
             uris,
-            time_range=tr,
-            variables=vrs,
-            metadata=self.metadata,
             handle_nodata=handle_nodata,
+            open_kwargs=open_kwargs or {},
```
deltamarnix: I think the drivers can also handle None, so no need to pass an empty dictionary here?

hydromt/data_catalog/sources/geodataset.py
```diff
@@ -59,23 +60,24 @@ def read_data(
...
         )
 
-        ds: Optional[xr.Dataset] = self.driver.read(
+        ds: xr.Dataset = self.driver.read(
             uris,
-            time_range=tr,
-            variables=vrs,
-            metadata=self.metadata,
             handle_nodata=handle_nodata,
+            open_kwargs=open_kwargs or {},
```
deltamarnix: same here

hydromt/data_catalog/sources/rasterdataset.py
```diff
@@ -80,13 +82,13 @@ def read_data(
 
         ds: xr.Dataset = self.driver.read(
             uris,
+            handle_nodata=handle_nodata,
+            open_kwargs=open_kwargs or {},
```
deltamarnix: same here

deltamarnix (changes requested):
I think we're almost there. 
I see a newly added typeCheckingMode in the settings that can go, and I think you're adding too much safety in some places with `open_kwargs=open_kwargs or {}`. See my comments.

hydromt/data_catalog/drivers/dataframe/pandas_driver.py
```diff
@@ -5,15 +5,11 @@
...
 import pandas as pd
 
-from hydromt._utils.unused_kwargs import _warn_on_unused_kwargs
 from hydromt.data_catalog.drivers.dataframe import DataFrameDriver
 from hydromt.error import NoDataStrategy, exec_nodata_strat
 from hydromt.typing import (
-    StrPath,
-    TimeRange,
     Variables,
 )
```
deltamarnix:
```suggestion
from hydromt.typing import Variables
```

hydromt/data_catalog/sources/rasterdataset.py
```diff
@@ -133,25 +135,32 @@ def to_file(
...
             buffer=buffer,
             time_range=time_range,
             zoom=zoom,
             handle_nodata=handle_nodata,
+            single_var_as_array=False,
         )
         if ds is None:  # handle_nodata == ignore
             return None
+        elif isinstance(ds, xr.DataArray):
+            ds = ds.to_dataset()
```
deltamarnix: Can you double check this?

hydromt/data_catalog/data_catalog.py
```diff
@@ -1753,22 +1785,23 @@ def get_dataframe(
...
                 )
                 if not isinstance(source, DataFrameSource):
                     raise ValueError(f"Source '{source.name}' is not a DataFrame.")
             else:
-                if "provider" not in kwargs:
-                    kwargs.update({"provider": "user"})
-                driver: str = kwargs.pop(
+                source_kwargs = source_kwargs or {}
+                if "provider" not in source_kwargs:
+                    source_kwargs.update({"provider": "user"})
```
deltamarnix: You could use setdefault

hydromt/data_catalog/sources/data_source.py
deltamarnix: @LuukBlom mentioned to me that this `to_file` function also has a kwargs and should be changed to dictionaries. We can do it in another PR, but we do need to change it. @savente93 , what do you say?

aslowwriter: Personally I'd prefer a separate PR, since this one has been back and forth a bunch of time already, but I'm happy to defer to whoever is picking it up in place of Luuk, I don't think it's a super big issue

deltamarnix: I've added it here anyways. It were just a couple of lines.

PR #1308, opened 2025-11-05
Remove io submodule

### LINKED ISSUE #1289: Rename 'io' submodule
### Kind of request

### Enhancement Description

When debugging in python, submodules that carry the same name as a Python build-in library (e.g. 'io') can result in unexpected behaviour. Most of the time it will result in the debugger just closing after a certain point. Therefore it is highly discouraged (also from a practical standpoint when importing modules) to name a submodule the same as build-in Python libraries. I would suggest to rename the 'io' submodule to something else, even something as stupid as 'hio' (standing for e.g. hydromt io) can be enough.

### Use case

Debugging.

### Additional Context

No.

aslowwriter:
## Issue addressed

Fixes #1289 

## Explanation

We decided that rather than rename the io module, we might as well get rid of it and just have the readers and writers in the root module. 

## General Checklist

- [x] Updated documentation
- [x] Updated changelog.rst

## Data/Catalog checklist


tests/test_readers.py
deltamarnix, line 1: Please move this test

tests/test_writers.py
deltamarnix, line 1: Please move this test

tests/test_readers.py
aslowwriter: good point

Issue #1295, opened 2025-10-23
Cleanup code duplication in the `DataSource` and `Driver` classes


LuukBlom:
Currently there is a lot of code duplication between all the `DataSource` classes, especially in the `read_data` functions. 

Essentially every function has this structure: 
```python
    def read_data(...)
        # required:
        self._mark_as_used()
        self._log_start_read_data()

        # optional:
        vrs: Optional[List[str]] = self.data_adapter._to_source_variables(variables)
        if bbox is not None or (mask is not None and buffer > 0):
            mask = _parse_geom_bbox_buffer(mask, bbox, buffer)

        # required:
        time_range: TimeRange = self.data_adapter._to_source_timerange(time_range)
        uris: List[str] = self.uri_resolver.resolve(...)
        data: pd.DataFrame = self.driver.read(...)
        return self.data_adapter.transform(...)
```
With some modifications (one or two extra function calls / variables) depending on the DataSource.

We should have a design session to think of a solution and perhaps remove all of the `DataSource` classes, so we can actually use `BaseDriver` as an interface. We could enforce exactly the same function signature for all `Driver.read` and then have some other way of adding the optional arguments currently added in the `Driver.read` functions.

deltamarnix:
In the documentation we see that both `data_type` and `driver` are mandatory. https://deltares.github.io/hydromt/latest/guides/advanced_user/data_prepare_cat.html#data-yaml

That means that the user needs to specify which driver is going to be used anyways. This could just be the name of the driver, otherwise it also requires some information with driver options. Since a driver can only return one data type, it is superfluous in a data catalog to have the data_type in there.

In code we have a not-so-generic RasterDatasetSource. It can actually only handle two drivers. xarray and rasterio. I would simply this by actually specifying more types of data sources: XarraySource, RasterIoSource. And let those handle very specific drivers. We would get rid of RasterDatasetSource and arguments to the drivers can be more specific instead of more generic. You need generic when things need to be handled the same. In our case, we need to handle everything differently. The base classes `DataAdapterBase`, `BaseDriver`, and `DataSource` are only there for the plugin to find those classes. Instead, I would opt for the plugin to only search for different DataSources, and instead of a base class, we could also go for a decorator.

An example could be like so:

```py
@hydromt_datasource
class RasterIoDataSource:
    def read_data():
        driver = RasterIoDriver(...)
        ds = driver.read(...)
        adapter = GeoDatasetAdapter(...)
[...]

PR #1312, opened 2025-11-07
Remove open_kwargs from get_* functions in DataCatalog

### LINKED ISSUE #1287: Throw an error if time_tuple is used instead of time_range in get_data methods

deltamarnix:
## Issue addressed

Fixes #1287

## Explanation

Supercedes PR of @LuukBlom PR (#1291). Discussed with @hboisgon and we came to the conclusion that these `open_kwargs` were introduced after release 1.2.0 and make it ambiguous which arguments are used when. It was introduced with an update on the documentation, when the DriverOptions class was also introduced.

We have decided that it's better to make the DataCatalog leading in the arguments. So if you want to make any changes in the read functionality, make sure that you do it in your data catalog, or the init of your DataSource that you create or pass along.

If you need to change additional driver kwargs before reading, the workaround is to pass a DataSource child object directly into the function.
Example:

```py
dc = DataCatalog()
source = RasterDatasetSource()
ds = dc.get_rasterdataset(source, ...)
```

But it's preferred that the information is correct in the DataCatalog that the user provides.

We're still discussing how we can make this easier in the DataCatalog. The writer of the data_catalog.yml shouldn't have to know the difference between source, driver, and adapter. See #1295.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/model/processes/region.py
+++ b/hydromt/model/processes/region.py
@@ -209,10 +209,8 @@ def parse_region_geom(
 
 
 def parse_region_grid(
-    region: dict,
-    *,
-    data_catalog: Optional[DataCatalog],
-) -> Union[xr.DataArray, xr.Dataset]:
+    region: dict[str, Any], *, data_catalog: DataCatalog | None = None
+) -> xr.DataArray | xr.Dataset | None:
     """Parse a region of kind grid and return the corresponding xarray object.
 
     Note that additional arguments from the region can be passed to the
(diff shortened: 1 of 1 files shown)


hboisgon (approved):
LGTM

PR #1331, opened 2025-11-26
Fix/bug nodata datacatalog export

Tjalling-dejong:
## Issue addressed

Encountered a bug with exporting artifact data with the datacatalog.export_data method.
If no data is found within the given bounding box the export_data method will break instead of handle the no data.

I also saw that some logging and error messaging in the drivers was not properly working. I improved that as well.

## Explanation

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -50,6 +50,6 @@ jobs:
       - name: Test
         run: pixi run --locked test-cov
       - name: SonarQube Scan
-        uses: SonarSource/sonarqube-scan-action@v6
+        uses: SonarSource/sonarqube-scan-action@v7
         env:
           SONAR_TOKEN: ${{ secrets.SONARCLOUD_TOKEN_2025 }}


hydromt/data_catalog/sources/geodataset.py
```diff
@@ -134,6 +134,10 @@ def to_file(
             handle_nodata=handle_nodata,
         )
         if ds is None:  # handle_nodata == ignore
+            exec_nodata_strat(
+                handle_nodata,
+                f"Reading file(s) for {self.name} returned no data.",
+            )
```
hboisgon: Did you check the log? Feels like we might be warning twice? Already in read_data just above and now here again?

Tjalling-dejong: Yes that could be possible. It depends on how the drivers are implemented. For geodataset the warning is indeed logged twice if the handle_nodata is set to WARN, by default it is raises an exception. 

We could remove this code block for this data source. I just saw that the to_file method of every data source contains this code block and changed them because some drivers can return None instead of a dataset or dataframe.

Tjalling-dejong: I went through the code and saw that the nodata for geodataframe source was not properly handled. I removed quite some code and there should now not be any duplicate warnings.

tests/data_catalog/test_data_catalog.py
```diff
@@ -589,6 +589,31 @@ def test_export_dataframe(tmp_path: Path, df, df_time):
...
+        bbox=bbox,
+        force_overwrite=True,
+    )
+    # test if data catalog can be read
+    new_data_catalog = DataCatalog(
+        data_libs=[str(data_catalog_reread_path / "data_catalog.yml")]
+    )
+    assert (
+        len(new_data_catalog) == 45
+    )  # Number of exported sources, not all exported due to nodat for this bbox
```
LuukBlom: Dont love the  `== 45` part, that can break due to many factors and if this test fails, it will be very cryptic to resolve. 

What do you think of using this exported data catalog as a starting point for testing?

1. read artifact_data
2. export for a small region
3. read created datacatalog
4. export it again
5. assert equality (can be just source names, or more in-depth source comparisons? these can then be put in the error msg when the assertion fails for some extra debugging info)

tests/data_catalog/sources/test_dataset_source.py
```diff
@@ -13,14 +13,14 @@
 
 
 class TestDatasetSource:
-    def test_read_data(
+    @pytest.fixture
+    def MockDatasetSource(
```
LuukBlom: We are actually returning an instance, not a class, so the fixture name is a bit misleading to me.
Could you rename `MockDatasetSource` to `mock_dataset_source`

tests/data_catalog/test_data_catalog.py
```diff
@@ -604,9 +605,23 @@ def test_export_data_bulk(tmp_path: Path, caplog: pytest.LogCaptureFixture):
...
+    data_catalog_export_path = tmp_path / "bulk_exported_2"
+
+    new_data_catalog.export_data(
+        data_catalog_export_path,
+    )
+
+    new_data_catalog_reread = DataCatalog(
+        data_libs=[str(data_catalog_export_path / "data_catalog.yml")]
+    )
+    assert len(new_data_catalog_reread.sources) == len(new_data_catalog.sources)
```
LuukBlom: Using the `.sources` property directly is not recommended in hydromt v1 I believe.
To get all sources / loop over them, we should use `list_sources()`, or since the DataCatalog implements `__len__` (as `return len(self.list_sources())`) 

so we can do: `assert len(new_data_catalog_reread) == len(new_data_catalog)

PR #1332, opened 2025-11-26
fix: datacatalog export data

### LINKED ISSUE #1323: Error in `DataCatalog.export_data()` when exporting .tif files with wildcards in the filename
### HydroMT version checks

### Reproducible Example

This code:

```python

data_catalog = DataCatalog(data_libs=["deltares_data"])

folder_name = "tmp_data_export"
data_catalog.export_data(
    new_root=folder_name,
    source_names=["modis_lai"],
)
```

Produces the following logging:

```
2025-11-12 18:32:31,096 - hydromt.data_catalog.data_catalog - data_catalog - INFO - Reading data catalog deltares_data latest
2025-11-12 18:32:31,097 - hydromt.data_catalog.data_catalog - data_catalog - INFO - Parsing data catalog from [C:\Users\buitink\.hydromt\deltares_data\v1.1.0\data_catalog.yml](file:///C:/Users/buitink/.hydromt/deltares_data/v1.1.0/data_catalog.yml)
2025-11-12 18:32:31,523 - hydromt.data_catalog.sources.data_source - data_source - INFO - Reading modis_lai RasterDataset data from [p:\wflow_global\hydromt\landuse\modis\MODIS_MCD15A3H_LAI\](file:///P:/wflow_global/hydromt/landuse/modis/MODIS_MCD15A3H_LAI/)*.tif
2025-11-12 18:32:35,237 - hydromt.data_catalog.drivers.raster.raster_xarray_driver - raster_xarray_driver - WARNING - Unknown extension for RasterDatasetXarrayDriver: .tif,switching to zarr
```

and the following error

```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'c:\\Users\\buitink\\Downloads\\tmp\\tmp_data_export\\*.zarr'
```

### Current behaviour

The `modis_lai` data consists of multiple .tif files, indicated in the data_catalog via `*.tif`, see these lines: https://github.com/Deltares/hydromt/blob/e022de8ebf5b80d32ef7df7ce06220780a5cdcd6/data/catalogs/deltares_data/v1.1.0/data_catalog.yml#L1620-L1623

In the export_data function, this data is being read, and (tried) to be written to a .zarr file. However, due to this file name in the data_catalog, it produces an incorrect filename.

Please note that is also happens with the `soilgrids` data (and probably many others with wildcards in the filename).

### Desired behaviour

Able to export these files. Also, from some files, .tif might be a more user friendly format compared to .zarr, but that is a completely different discussion ;)

### Additional context

Comment from @deltamarnix : the error comes from `RasterDatasetSource.to_file` (+- line 121)

Tjalling-dejong:
## Issue addressed

Fixes #1323 

I added write support for the rasterio driver. I had to add some logic to handle multivariate data like xr.Dataset and multidimensional data arrays. This data is written to multiple files with a datacatalog path containing a wildcard. 

I also encountered a bug when exporting data  with the datacatalog.export_data method. When you want to export certain variables from a data source, the variables were not sent to the to_file method and the whole datasource was exported instead.

## Explanation

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist


hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -248,17 +252,61 @@ def write(
...
-        raise NotImplementedError()
+        no_ext, ext = splitext(path)
+        write_kwargs = write_kwargs or {}
+        # set filepath if incompat
+        if ext not in self.SUPPORTED_EXTENSIONS:
+            logger.warning(
+                f"Unknown extension for RasterioDriver: {ext}, switching to .tif"
+            )
+            path = no_ext + _TIFF_EXT
+            ext = _TIFF_EXT
```
LuukBlom: I prefer errors in cases like this

hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -234,8 +237,9 @@ def _open() -> xr.Dataset:
     def write(
         self,
         path: Path | str,
-        data: xr.Dataset,
+        data: xr.Dataset | xr.DataArray,
         *,
+        source_name: str,
```
LuukBlom: All `Driver.write` functions have the exact same signature.
So, Im not a fan of changing the signature to handle this specific use case of writing to multiple files with a wildcard.

It is the callers responsibility to provide a correct path. 

from discussion: look at where `DataSource._get_uri_basename` gets called in `Datacatalog.export_data()`, update the path there

I think a solution without changing the signature would be ideal.

tests/data_catalog/drivers/raster/test_rasterio_driver.py
```diff
@@ -78,6 +78,45 @@ def test_sets_mosaic_kwargs(self, fake_open_mfraster: MagicMock):
...
+            ValueError,
+            match="Writing multiple files with wildcard requires at least 3 dimensions in data array",
+        ):
+            RasterioDriver().write(uri, rioda)
+
+        # expand dims to have at least 3 dimensions
+        rioda_expanded = rioda.expand_dims("time").assign_coords(time=[0])
+        p = RasterioDriver().write(uri, rioda_expanded)
+        assert p == tmp_path / "test_write_wildcard0_0.tif"
+        assert (tmp_path / "test_write_wildcard0_0.tif").is_file()
```
LuukBlom: This seems strange to me. 

We want to write to `tmp_path / "test_write_*.tif"`

But we end up writing to `tmp_path / "test_write_wildcard0_0.tif"` 

Note: 
tmp_path returns this path: `{temproot}/pytest-of-{user}/pytest-{num}/{testname}/` so the name of the dir for this test is `pytest test_write_wildcard`.

This test only works because the filename `"test_write_*.tif"` overlaps with the test function name `test_write_wildcard`.

Reading from the path `tmp_path / "test_write_*.tif"` should read in the same data again. I think it would be good to that full circle test to ensure that behaviour.

hydromt/data_catalog/drivers/raster/raster_xarray_driver.py
```diff
@@ -178,6 +178,7 @@ def write(
         data: xr.Dataset,
         *,
         write_kwargs: dict[str, Any] | None = None,
+        source_name: str | None = None,
```
LuukBlom: dont forget to remove

hydromt/data_catalog/drivers/raster/raster_dataset_driver.py
```diff
@@ -74,6 +74,7 @@ def write(
         data: xr.Dataset,
         *,
         write_kwargs: dict[str, Any] | None = None,
+        source_name: str | None = None,
```
LuukBlom: dont forget to remove

hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -248,17 +250,57 @@ def write(
...
+            dim0 = data.dims[0]
+            for i in data[dim0]:
+                ds_sel = data.sel({dim0: i})
+                file_path = file_dir / f"{file_dir.stem}_{i.values}{ext}"
+                ds_sel.raster.to_raster(
+                    file_path,
+                    driver=gdal_driver,
+                    **write_kwargs,
+                )
+            return file_path if len(data[dim0]) == 1 else file_dir / f"*{ext}"
```
LuukBlom: I think this is not what we want.

You are throwing away the part of the filename before the `*`.

I think something like this would be best: 

```python

if "*" in str(path) and isinstance(data, xr.DataArray):
    if len(data.dims) < 3:
        raise ValueError(
            "Writing multiple files with wildcard requires at least 3 dimensions in data array"
        )

    path = Path(path)
    path.parent.mkdir(parents=True, exist_ok=True)
    if path.name.count("*") != 1:
        raise ValueError(
            "There must be exactly one wildcard `*` in the filename when multiple outputs required"
        )

    dim0 = data.dims[0]

    for label in data[dim0]:
        ds_sel = data.sel({dim0: label})
        file_name = path.name.replace("*", f"{label.values}")

        ds_sel.raster.to_raster(
            path.with_name(file_name),
            driver=gdal_driver,
            **write_kwargs,
        )

    return path  # return path with the wildcard still there
```

hydromt/data_catalog/data_catalog.py
```diff
@@ -1883,3 +1882,22 @@ def _denormalise_data_dict(data_dict) -> List[Tuple[str, Dict]]:
...
+    basename: str = source._get_uri_basename(handle_nodata, **source_kwargs)
+    if "*" in basename or (
+        source_kwargs.get("variables", None)
+        and isinstance(source.driver, RasterioDriver)
+    ):
+        p = Path(new_root) / source.name / basename
+        p.parent.mkdir(parents=True, exist_ok=True)
+    else:
+        p = Path(new_root) / basename
+    return p
```
LuukBlom: Tbh, I dont like that we are changing the path here. In my opinion we should just use whatever was provided by the user, and error if it is not valid.

So if people use wildcards for their sources, they should contain enough information to actually read and write the correct files. 
e.g.: the dir/source name relative to the datacatalog root should be in the uri as well: 
- `source: ..., uri: myname/*.tif` 
- `source: ..., uri: myname_*.tif`

hydromt/data_catalog/drivers/raster/rasterio_driver.py
LuukBlom: please see https://github.com/Deltares/hydromt/pull/1337 for my actual proposed fix

hydromt/data_catalog/sources/data_source.py
```diff
@@ -171,17 +172,13 @@ def _get_uri_basename(self, handle_nodata: NoDataStrategy, **query_kwargs) -> st
...
             # if multiple_uris, use the first one:
             if len(uris) > 0:
-                uri: str = uris[0]
+                uri = PurePath(uris[0])
             else:
                 raise NoDataException("!")
-        else:
-            uri: str = self.uri
+        if self.root and uri.is_absolute():
+            uri = uri.relative_to(self.root)
```
LuukBlom, line 179: this will error if self.root is not a parent of the uri.

```python
if self.root is not None and uri.is_absolute():
    root = Path(self.root)
    if uri.is_relative_to(root)): 
        # stored under the data catalog root, so return a relative path
        uri = uri.relative_to(root).as_posix()
    else:
        # some path outside of the data catalog root, so we make up a relative path to write to
        uri = f"{self.name}/{uri.name} 
        # should we include more than just uri.name to capture possible wildcards/ variables further up the uri? 
```

hydromt/data_catalog/drivers/raster/raster_dataset_driver.py
```diff
@@ -88,6 +88,8 @@ def write(
         write_kwargs : dict[str, Any] | None, optional
             Additional keyword arguments to pass to the underlying write function.
             Default is None.
+        source_name : str | None, optional
+            Optional name of the source dataset, used for naming conventions if needed.
```
LuukBlom: dont forget to remove

hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -248,17 +250,71 @@ def write(
...
+
+        gdal_driver = GDAL_DRIVER_CODE_MAP.get(path.suffix.lstrip(".").lower())
+
+        if "*" in str(path) and isinstance(data, xr.DataArray):
+            if len(data.dims) < 3:
+                raise ValueError(
+                    "Writing multiple files with wildcard requires at least 3 dimensions in data array"
+                )
+
+            path = Path(path)
```
LuukBlom, line 295: remove, see L 265

hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -248,17 +250,71 @@ def write(
...
+            if len(data.data_vars) == 1:
+                data = data[list(data.data_vars.keys())[0]]
+            else:
+                path.parent.mkdir(parents=True, exist_ok=True)
+                for var in data.data_vars:
+                    if "*" in path.name:
+                        file_name = path.name.replace("*", var)
+                        file_path = path.with_name(file_name)
+                    else:
+                        file_path = path.parent / f"{var}{path.suffix}"
```
LuukBlom, line 323: Here we throw away the last part of the path again. 

`path.stem` returns the filename without suffix

```python
if "*" in path.name:
    file_name = path.name.replace("*", var)
    file_path = path.with_name(file_name)
else:
    file_path = path.parent / f"{path.stem}_{var}{path.suffix}"
    # or: file_path = path.with_stem(f"{path.stem}_{var})")
```

hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -248,17 +250,71 @@ def write(
...
+                    data_raster = data[var]
+                    data_raster.raster.to_raster(
+                        file_path, driver=gdal_driver, **write_kwargs
+                    )
+                return path.parent / f"*{path.suffix}"
```
LuukBlom: Same thing where we throw away the stem. 
I think returning `path` as is should be fine? that still has the wildcard, but also the correct stem.

hydromt/data_catalog/drivers/dataframe/pandas_driver.py
```diff
@@ -63,7 +63,11 @@ def read(
                 f"{self.__class__.__name__} driver is not supported."
             )
         elif len(uris) == 0:
-            df = pd.DataFrame()
+            exec_nodata_strat(
+                "No URIs provided to read data from.",
+                strategy=handle_nodata,
+            )
+            return None  # handle_nodata == ignore
```
LuukBlom: Looking at it now, do you agree these checks should actually live in the URI resolver? It also gets `handle_nodata` as an argument.

hydromt/data_catalog/drivers/geodataframe/pyogrio_driver.py
```diff
@@ -87,7 +87,11 @@ def read(
                 f"{self.__class__.__name__} driver is not supported."
             )
         elif len(uris) == 0:
-            gdf = gpd.GeoDataFrame()
+            exec_nodata_strat(
+                "No URIs provided to read data from.",
+                strategy=handle_nodata,
+            )
+            return None  # handle_nodata == ignore
```
LuukBlom: Looking at it now, do you agree these checks should actually live in the URI resolver? It also gets handle_nodata as an argument.

hydromt/data_catalog/drivers/geodataframe/table_driver.py
```diff
@@ -99,7 +99,13 @@ def read(
...
         if not metadata:
             metadata = SourceMetadata()
-        if len(uris) > 1:
+        if len(uris) == 0:
+            exec_nodata_strat(
+                "No URIs provided to read data from.",
+                strategy=handle_nodata,
+            )
+            return None  # handle_nodata == ignore
+        elif len(uris) > 1:
```
LuukBlom: Looking at it now, do you agree these checks should actually live in the URI resolver? It also gets handle_nodata as an argument.

hydromt/data_catalog/drivers/geodataset/vector_driver.py
```diff
@@ -82,7 +82,13 @@ def read(
...
             The dataset read from the source.
         """
-        if len(uris) > 1:
+        if len(uris) == 0:
+            exec_nodata_strat(
+                "No URIs provided to read data from.",
+                strategy=handle_nodata,
+            )
+            return None  # handle_nodata == ignore
+        elif len(uris) > 1:
```
LuukBlom: Looking at it now, do you agree these checks should actually live in the URI resolver? It also gets handle_nodata as an argument.

hydromt/data_catalog/drivers/geodataset/xarray_driver.py
```diff
@@ -90,6 +90,12 @@ def read(
             },
         )
         preprocessor = self.options.get_preprocessor()
+        if len(uris) == 0:
+            exec_nodata_strat(
+                "No URIs provided to read data from.",
+                strategy=handle_nodata,
+            )
+            return None  # handle_nodata == ignore
```
LuukBlom: Looking at it now, do you agree these checks should actually live in the URI resolver? It also gets handle_nodata as an argument.

hydromt/data_catalog/drivers/raster/rasterio_driver.py
```diff
@@ -280,3 +346,22 @@ def _get_zoom_levels_and_crs(uri: str) -> tuple[dict[int, float], int]:
...
+            y_coords.size < 2
+            or (y_coords.ndim == 2 and y_coords.shape[0] < 2)
+            or x_coords.size < 2
+            or (x_coords.ndim == 2 and x_coords.shape[1] < 2)
+        ):
+            exec_nodata_strat(
+                f"Cannot write raster data with insufficient spatial dimensions: {data.raster.y_dim} size {y_coords.size}, {data.raster.x_dim} size {x_coords.size}",
+                strategy=NoDataStrategy.RAISE,
+            )
+            return None
```
LuukBlom: Ntpick: since we will always be raising here, I prefer `raise NoDataException(msg)`,

hydromt/data_catalog/sources/data_source.py
```diff
@@ -171,17 +172,20 @@ def _get_uri_basename(self, handle_nodata: NoDataStrategy, **query_kwargs) -> st
...
+                uri = PurePath(uris[0])
             else:
-                raise NoDataException("!")
-        else:
-            uri: str = self.uri
+                exec_nodata_strat(
+                    f"No URIs could be resolved from the DataSource: {self.name}.",
+                    handle_nodata,
+                )
+                return None
```
LuukBlom: Looking at it now, do you agree these checks should actually live in the URI resolver? It also gets handle_nodata as an argument.

hydromt/data_catalog/drivers/dataframe/pandas_driver.py
Tjalling-dejong: We can remove the exec_nodata_strat indeed. The uri resolver already implements this if no uris are found

Issue #1326, opened 2025-11-25
binder is not working in the documentation


deltamarnix:
### HydroMT version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

See https://github.com/Deltares/hydromt_wflow/issues/632

### Current behaviour

it's not importing

### Desired behaviour

binder examples should work

PR #1333, opened 2025-11-27
fix docker build

aslowwriter:
## Issue addressed

Fixes part of #1326 

## Explanation

A bit more might be necessary but I suspect that a large part of the reason it is broken is that the docker images seemingly haven't been published since v10.1. This step should fix the docker build tasks so we can publish it, after that I'll have to see if additional work is necessary.

Locally the binder test works when I built the docker image locally. it will just have to be published as well after this is merged. 

In addition to fixing the docker I've also reworked the test workflow for the binder. instead of invoking the `repo2docker` cli which tests a lot less than I thought, we build the images themselves, spin them up and see if we can connect to the jupyter server. I think this is as good as we can reasonably test this for now. This doesn't guarantee that the binder will work since that one will try to pull from dockerhub. If that is also important we could implement a routine test with the latest published one, but I'll leave that for now to avoid scope creep. 

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/.github/workflows/test-docker.yml
+++ b/.github/workflows/test-docker.yml
@@ -42,7 +42,7 @@ jobs:
           docker run -d -p 8888:8888 deltares/hydromt:binder
 
           # give it a second or two to start up
-          sleep 2
+          sleep 10
 
           # check if we can reach the server
           curl -f "http://localhost:8888/api/"


.github/workflows/test-docker.yml
```diff
@@ -5,12 +5,14 @@ on:
   push:
     branches: [main]
     paths:
+      - .github/workflows/test-docker.yml
       - .binder/*
```
deltamarnix:
```suggestion
      - .binder/**
```

.github/workflows/test-docker.yml
```diff
@@ -5,12 +5,14 @@ on:
...
       - Dockerfile
       - pyproject.toml
   pull_request:
     branches: [main]
     paths:
+      - .github/workflows/test-docker.yml
       - .binder/*
```
deltamarnix:
```suggestion
      - .binder/**
```

pyproject.toml
```diff
@@ -300,7 +300,7 @@ docker-build = { cmd = [
 	"--build-arg",
 	"PIXIENV=$DOCKER_TARGET-latest",
 	".",
-], env = { DOCKER_TAG = "<DOCKER_TAG>", DOCKER_TARGET = "<DOCKER_TARGET>" } }
+] }
```
deltamarnix, line 303: I was expecting that docker-build would be called in the github action, but instead a separate action is used. Consider if you want to make use of the pixi tasks during github actions, because it makes it perhaps easier to maintain and reproduce locally vs on our servers.

So this `docker-build` task is local only? How does one add DOCKER_TAG and DOCKER_TARGET? If it's with environment variables... you can now use task arguments in pixi: https://pixi.sh/latest/workspace/advanced_tasks/#task-arguments, which is more locally controlled during the pixi step.

```toml
docker-build = { cmd = [
"docker",
"build",
"-t",
"deltares/hydromt:{{ docker_tag }}",
"--target={{ docker_target }}",
"--build-arg",
"PIXIENV={{ docker_target }}-latest",
".",
],
args=[
  { "arg"="docker_target", "default"="slim" },
  { "arg"="docker_tag",  "default"="dev"}
] }
```

call it with `pixi run docker-build` if you want to use the defaults, or e.g. `pixi run docker-build full v1.3.0` if you want more control.

aslowwriter: I'll rewrite it to use the pixi tasks

savente93:
> Then perhaps we can also add pushing binder to our release pipeline?

binder is a pull based system, so I'm not aware of a way to do this

pyproject.toml
deltamarnix: Are you keeping the environment variables?

pyproject.toml
```diff
@@ -322,6 +322,14 @@ docker-build-latest = { cmd = [
 	"docker-build",
 ], env = { DOCKER_TAG = "latest", DOCKER_TARGET = "slim" } }
 
+docker-build-binder = { cmd = [
+	"docker",
+	"build",
+	"-t",
+	"deltares/hydromt:binder",
+	".binder/",
+]}
```
deltamarnix: Since the binder docker depends on `deltares/hydromt:slim`, we can add this dependency here in pixi like so:

```suggestion
], "depends-on"=["dockker-build-slim"]}
```

Then you won't have to add two pixi steps in the github action

deltamarnix (approved):
Some minor comments on the pixi file. But it looks good to me.

pyproject.toml
aslowwriter: yes, the environment variables stay in the tasks that actually define them, but leaving these "marker" variables was causing issues so I removed them. The environment variables are still defined in the specific tasks though

pyproject.toml
aslowwriter: good point

Issue #1194, opened 2025-05-21
GeomsComponent region returns only the total bounds of the first geometry


dalmijn:
### HydroMT version checks

- [x] I have checked that this issue has not already been reported.
- [x] I have checked that this bug exists on the latest version of HydroMT.

### Reproducible Example

Create a GeomsComponent and add two non overlapping geometries to it.
Request the region by simply calling `component.region`. 

### Current behaviour

Returns the total bounds of the first geometry.

### Desired behaviour

Returns the total bounds of the combined geometries.

I think it's as simple as replacing these lines:
https://github.com/Deltares/hydromt/blob/main/hydromt/model/components/geoms.py#L91-L97

with:
```Python
        bounds = np.column_stack([geom.total_bounds for geom in self.data.values()])
        total_bounds = (
            bounds[0, :].min(),
            bounds[1, :].min(),
            bounds[2, :].max(),
            bounds[3, :].max(),
        )
```

As when the bounds properties of GeoDataFrames are put into a columns stack if appends them as rows...
Which means that the 0 to 3rd column are just the bounds of the first GeoDataFrame.

Issue #1127, opened 2025-03-13
GeomsComponent return geometry instead bounds as region


dalmijn:
### Kind of request

Changing existing functionality

### Enhancement Description

GeomsComponent returns a bounding box as a region, which feels counterintuitive when using this component as your region component. If ones supplies a region vector file to be used (for e.g. wflow) to setup the data, doesnt it then setup data for the entire bounding box?

### Use case

General.

### Additional Context

No.

savente93:
I would like some more opinions on this, but as the geoms component could contain multiple geodataframes, a couple of options we could consider instead of the total bounds:
- Return the union of all geoms
- Return a list of arrays
- return all geodataframes concatenated

Just from a pure semantics point the first one sounds the most intuitive to me, but I wonder what others think. @hboisgon I remember you had some opinions on this back when we discussed it, what do you think?

deltamarnix:
We have the following implementations for `region`, and I think it is the best that we are consistent.

* GeomsComponent: Makes a total bounds calculation and returns an axis-aligned bounding box.
* GridComponent: Returns an axis-aligned bounding box of the grid
* MeshComponent: Makes an axis-aligned bounding box of the bounds of the mesh
* VectorComponent: Creates an axis-aligned bounding box of the vector geometry

If we change it, we should change it for all of them. Even for a grid we would have to think if we rather return the transformed bounding box, or a convex hull in that sense, of the grid. Rather than AABB.

Also, I would like to know what the functions that use `component.region` do with it. For example, `get_rasterdataset` uses the region of the model. Will it still work if it's a more complex shape?

DirkEilander:
I agree with Brendan that it would be nice to get the region geometry back, possibly as a union. We can make the axis-aligned bounding box a new `region_bounds` property. `get_rasterdataset` can deal with more complex geometries as well, see `raster.clip_geom`.

dalmijn:
In addition: for hydromt_fiat v1, we developed our own RegionComponent, which was relatively easy to do and behaves how we want it to. Which we did because, like @DirkEilander said, `get_rasterdataset` is able to use complex polygons to clip the data.
So maybe there is the discussion whether we want this functionality in core, or that it is the responsibility of the plugins to create something tailor-made in regards to the region. Although I still believe that having the functionality to work with polygons would be handy in core.

PR #1334, opened 2025-11-27
changes for UWBM

LuukBlom:
Changes to get the tests for https://github.com/Deltares-research/hydromt_uwbm/pull/4 green.

## Related issues: 

- https://github.com/Deltares-research/hydromt_uwbm/issues/1
- https://github.com/Deltares/hydromt/issues/1194
- https://github.com/Deltares/hydromt/issues/1127

## Explanation

- config.py: also read  `.ini` files with read_toml
- geoms.py: updated `_region_data` to return the union of all geoms, not the bbox. Also added `crs` property that gets the crs from self.data if possible. There already is a property `bounds` in `SpatialModelComponent` that returns `self.region.total_bounds`. So the bbox is still accessible (alternative: `model.region.total_bounds` also works.) 
- spatial.py: return `_region_data` instead of `region` to prevent infinite loop

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/model/components/geoms.py
+++ b/hydromt/model/components/geoms.py
@@ -80,10 +80,18 @@ class GeomsComponent(SpatialModelComponent):
 
     @property
     def crs(self) -> Optional[CRS]:
-        """Return the CRS of the geometries.
+        """
+        Return the coordinate reference system associated with the geometries.
 
-        If multiple geometries with different CRS are present, return the CRS of the first one found.
-        If no geometries have a CRS, return None.
+        If multiple GeoDataFrames are present with differing CRS values, the CRS of the first
+        GeoDataFrame that defines one is returned. If none of the stored geometries provide a
+        CRS, the method returns None.
+
+        Returns
+        -------
+        CRS or None
+            The CRS of the first geometry with a defined coordinate reference system, or None
+            if no CRS is available.
         """
         for geom in self.data.values():
             if geom.crs is not None:
@@ -93,15 +101,17 @@ class GeomsComponent(SpatialModelComponent):
     @property
     def _region_data(self) -> Optional[GeoDataFrame]:
         """
-        Provide the region as the union of all geometries in the data dictionary.
+        Return the region as the geometric union of all polygonal features stored in the data dictionary.
 
-        The region is constructed by calling explode on each GeoDataFrame to in self.data
-        to split multi-part geometries into single-part geometries, followed by computing the union.
+        Each GeoDataFrame in ``self.data`` is exploded to ensure that multi-part geometries are treated
+        as individual components before computing the union. Only geometries with area contribute to the
+        resulting region; non-area features such as points and lines are ignored.
 
         Returns
         -------
         geopandas.GeoDataFrame or None
-            The region geometry as the union of all geometries in the data dictionary,
+            A GeoDataFrame containing the unified region geometry, or None if no polygonal features
+            are available.
         """
         # Use the union of all geometries as region
         if len(self.data) == 0:
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -43,6 +43,8 @@ Changed
 - remove ``io`` submodule and move ``readers`` and ``writers`` to the root module (#1308)
 - `write_nc` function no longer accepts kwargs, instead a dict `to_netcdf_kwargs` is used to pass arguments to `xarray.to_netcdf`. The function raises an error when "compute" is one of the arguments.
 - Function signatures of `DataCatalog` functions `get_rasterdataset`, `get_geodataframe`, `get_geodataset`, `get_dataset` and `get_dataframe` now have: ``source_kwargs`` (passed to `DataSource.__init__`) instead of just ``kwargs``, and ``time_tuple`` was renamed to ``time_range``. (#1291)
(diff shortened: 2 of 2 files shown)


hydromt/model/components/config.py
```diff
@@ -136,7 +136,7 @@ def read(self, path: Optional[str] = None) -> None:
         match read_path.suffix.lower():
             case ".yaml" | ".yml":
                 self._data = read_yaml(read_path)
-            case ".toml":
+            case ".toml" | ".ini":
```
aslowwriter: we've deprecated the `ini` format a long time ago and also I'm not sure that a toml reader can read `ini` files? can you explain your thoughts behind this a bit more?

LuukBlom: uwbm uses .ini, and that file can be read using the toml reader. I added this here so i didnt have to copy paste the read fn of the config component and only change this part.

But from a quick google I can see that .ini is not always readable by toml readers so I will revert and just copy paste the function.

aslowwriter: ~~We deprecated ini files purposefully and they are not supported anymore, so UWBM will have to upgrade to yaml I'm afraid~~

Sorry, shouldn't have pressed comment that soon, I only realise now I'm confusing two things. never mind

hboisgon: Quick note here but we deprecated it mainly for the core workflow file no? Or altogether for model config component too? If the change is small I dont think it's that annoying to support ini in core but you can decide.
Reading the hydromt workflow yaml goes through a different function (read_workflow_yml)

aslowwriter: yes exactly, that's what I was confusing

hydromt/model/components/geoms.py
```diff
@@ -79,21 +78,45 @@ def _initialize(self, skip_read=False) -> None:
...
 
-        return region
+        # Flatten all geometries from all GeoDataFrames
+        all_geoms = [
+            geom
+            for gdf in self.data.values()
+            if not gdf.empty
+            for geom in gdf.geometry.explode(index_parts=False)
+            if geom is not None and not geom.is_empty and geom.geom_type == "Polygon"
+        ]
```
LuukBlom, line 127: @dalmijn @DirkEilander you mentioned this in https://github.com/Deltares/hydromt/issues/1127.

Do you agree with this implementation:
1. expode all multipolygons, into polygons
2. compute the union of all non empty polygons, while ignoring (Multi)Linestrings and Points.

We can include the linestrings and points as well, but that can make the region very strangely shaped when they fall outside of polygons. 

If I recall corectly, in hydromt-sfincs, points and linestrings that fall outside of the region are ignored/error as well. But im not sure. 

Of course, the bbox is accessible by calling `model.region.total_bounds`

hydromt/model/components/geoms.py
```diff
@@ -79,21 +78,45 @@ def _initialize(self, skip_read=False) -> None:
...
+    def crs(self) -> Optional[CRS]:
+        """Return the CRS of the geometries.
+
+        If multiple geometries with different CRS are present, return the CRS of the first one found.
+        If no geometries have a CRS, return None.
+        """
+        for geom in self.data.values():
+            if geom.crs is not None:
+                return geom.crs
+        return None
```
deltamarnix, line 99:
```suggestion
        return next((geom.crs for geom in self.data.values() if geom.crs is not None), None)
```

I thought I'd give it a go with a one liner, but Python is terrible for that 😅

aslowwriter: I actually dislike one-liners like that because they are much harder to read. I'll leave it up to Luuk, but I do think the original is better.

deltamarnix: Totally agree, let's resolve

hydromt/model/components/geoms.py
dalmijn: I think the idea of a union is good! Nice :).
Im wondering though whether this solves correctly the principle of mulitple geodataframe's in the component data that are in different crs's? From what I see these polygons keeps their coordinates from the old system, just with the new slapped on top op it. I guess a routine for reprojecting is needed there.
I really like the complex polygon, but know that `hydromt_wflow` relies on the bounding-box region as for certain interpolation steps it's necessary to pull in data beyond the basin domain. You might want to pass the `total_bounds` property there.

LuukBlom: It should not be possible to have geometries with different crs in one GeomsComponent. 
The `set` function ensures any geometries that are added are in the crs of the model. (or if that is not defined, of the same crs as the first geom found in its data, see L140)

So, unless people start messing around with the underlying `data/_data` attributes and not use `set()`, this should never happen and all geoms have the same crs.

dalmijn: And if the model crs is still None, i.e. no regioncomponent data?

LuukBlom: When converting geometries to model crs in the set function, this happens:
1. check the property `model.crs`, if this returns a valid crs, use that and convert everything to this.
2. If the property `model.crs` is None, use `self.crs` or `GeomsComponent.crs` to deterimine crs and convert to that.
3. if both above properties are None, there is no crs to use, so we use whatever the data has (can be None or some crs, which will then get used the next time a geometry is added.)

dalmijn: Sounds like it's covered. It's discouraged to use _data yourself, so that seems fine. Then I only have the comment regarding the `total_bounds` for hydromt_wflow. So the region return is fine as far as I'm concerned.

PR #1351, opened 2025-12-23
feat: Add function `sample_geoms` to the raster accessor as a performant alternative to `zonal_stats`.

LuukBlom:
## Issue addressed

Fixes [#521](https://github.com/Deltares/hydromt/issues/1358)

## Explanation

After discussing with @dalmijn, came up with this solution where we provide a quicker alternative to `raster.zonal_stats`.
Essentially, the function captures this behaviour:

- find sample points: one per geometry. (determined by `sampling_strategy`) & determine index (row, col) for each sample point
- directly sample the xr.Dataset. (this keeps xarray from doing massive amounts of overhead when the dataset is large)
- spatial datasets with no time dim, skip any stat calculations and return the value immediately. (makes no sense to take the mean/min/max of a single value)
- spatial datasets with time dim: sample each point at each time step, reduce by applying the `stats`

## Visualization
When I was testing this, I wrote a little notebook to help visualize bugs:
[showcase-sample.ipynb](https://github.com/user-attachments/files/24314192/showcase-sample.ipynb)

## Performance
Performance compared to zonal_stats: (will not add it to the repo since it can take quite a while, and its more proof of concept)
[test_performance.py](https://github.com/user-attachments/files/24524745/test_performance.py)
```console
---------------------------------------------------------- Captured stdout call ---------------------------------------------------------- 
sample_points: 0.102 s
zonal_stats:   80.086 s
speedup:      788.5x
```

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/gis/raster.py
+++ b/hydromt/gis/raster.py
@@ -909,7 +909,7 @@ class XRasterBase(XGeoBase):
             {'count', 'min', 'max', 'sum', 'mean', 'std', 'median', 'q##'}.
             Multiple percentiles can be calculated using comma-separated values,
             e.g.: 'q10,50,90'. Statistics ignore the nodata value and are applied
-            along the x and y dimension. By default 'mean'
+            along the x and y dimension.
         variables: list of str, optional
             List of variable names to calculate zonal statistics for. If None, all variables
             are used. By default None


hydromt/gis/raster.py
```diff
@@ -981,6 +986,190 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
...
+            reduce_dims = [d for d in da.dims if d not in (x_dim, y_dim)]
+
+            if not reduce_dims:  # single band raster, no stats needed
+                if stats is not None:  # no stats requested
+                    logger.warning(
+                        "Statistics requested for single-band raster; returning sampled value."
+                    )
+                out[var] = _to_index_1d(sampled_vals)
+            else:  # compute stats
+                stats_to_apply = stats or ["mean"]
```
hboisgon: I would just use "mean" as default argument. No need for None as possible type then and this would be more transpararent.

hydromt/gis/raster.py
```diff
@@ -981,6 +986,190 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
...
+            if geom is None or geom.is_empty:
+                continue
+
+            if sampling_strategy == "centroid":
+                p = geom.centroid
+                if not geom.contains(p):
+                    msg = f"Centroid sample point for geometry {i} is not within geometry."
+                    if geom.geom_type.startswith("Multi"):
+                        msg += " Consider using 'representative_point' sampling strategy when working with multi-part geometries."
+                    logger.warning(msg)
```
hboisgon: Nice alternative for speed. The loop over geoms will make it slow though...
I guess you need it for the warning about using centroid? Is there another way?

Else I would simplify by just deriving the points in a couple of lines (centroid ore representative points) and then why not just call ``raster.sample`` with a window of 0? And in case of 3D do the stats afterwards on the results?
Also the name might be a little confusing because of the ``sample`` method but I'm really bad at naming...

hydromt/gis/raster.py
```diff
@@ -981,6 +986,190 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
...
+                    elif stat == "sum":
+                        vals = np.nansum(sampled_vals, axis=0)
+                    elif stat == "std":
+                        vals = np.nanstd(sampled_vals, axis=0)
+                    elif stat == "median":
+                        vals = np.nanmedian(sampled_vals, axis=0)
+                    elif callable(stat):
+                        vals = np.apply_along_axis(stat, 0, sampled_vals)
+                    else:
+                        raise ValueError(f"Stat {stat} not valid for point sampling")
```
hboisgon: Why not stay closer to ``zonal_stats`` or at least no need for this large repeat?

See
```
                for stat in stats:
                    if stat in _ST:
                        ds1_stat = getattr(ds1, stat)(dims)
                        dss.append(ds1_stat.rename(rmd(ds1, stat)))
                    elif isinstance(stat, str) and stat.startswith("q"):
                        qs = np.array([float(q) for q in stat.strip("q").split(",")])
                        dss.append(
                            ds1.quantile(qs / 100, dims).rename(rmd(ds1, "quantile"))
                        )
                    elif callable(stat):
                        dss.append(
                            ds1.reduce(stat, dims).rename(rmd(ds1, stat.__name__))
                        )
                    else:
                        raise ValueError(f"Stat {stat} not valid.")
```

I think good that the two methods are consistent in their arguments and the objects (and array names) that it will return. It will make it easier for plugin to add an option for if stats=simple then ds_out=raster.sample.stats and else ds_out=raster.zonal_stats. If the return objects are the same any lines after that wont need to be tailor made to the use of one function or the other :)

hboisgon (commented):
Good idea. Wonder why you did not use ``raster.sample`` in the background? Good it the method is very simialr in terms of arguments and return object to ``raster.zonal_stats``.

hydromt/gis/raster.py
```diff
@@ -25,7 +25,6 @@
 import pyproj
 import rasterio.fill
 import rasterio.warp
-import rioxarray  # noqa: F401
```
hboisgon, line 28: My python knowledge is not the best here but are you sure about deleting this line?

hydromt/gis/raster.py
```diff
@@ -892,7 +892,12 @@ def sample(self, gdf, wdw=0):
             obj_out = obj_out.raster.mask(ds_sel["mask"])
         return obj_out
 
-    def zonal_stats(self, gdf, stats, all_touched=False):
+    def zonal_stats(
+        self,
+        gdf: gpd.GeoDataFrame,
+        stats: list[str] | Callable,
```
hboisgon: I think it can also be str like your new method? If not good to harmonise!
Also in the docstrings we say the default is mean but does look like it anymore? I would add the default here like you did for your method.

hydromt/gis/raster.py
```diff
@@ -981,6 +985,140 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
...
+
+        if gdf.crs is not None and self.crs is not None and gdf.crs != self.crs:
+            gdf = gdf.to_crs(self.crs)
+
+        if sampling_strategy == "centroid":
+            if gdf.crs.is_geographic:
+                raise ValueError(
+                    "Centroid sampling is not supported for geographic CRS, use 'sampling_strategy=representative_point' instead."
+                )
+            points = gdf.geometry.centroid
```
hboisgon: Really? Why? Geopandas geometry.centroid does work for geographic crs. Is there another reason? Else it's really easy to reproject a geometry to projected CRS I think with hydromt.gis.gis_utils.utm_crs or something like that.

hydromt/gis/raster.py
```diff
@@ -981,6 +985,140 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
...
+
+        gdf_points = gdf_points[valid_mask]
+        valid_indices = gdf_points.index.values
+
+        sampled = self.sample(gdf_points)
+        if isinstance(sampled, xr.DataArray):
+            sampled = sampled.to_dataset(name=sampled.name or "values")
+
+        if variables is not None:
+            sampled = sampled[variables]
```
hboisgon: Method might be even faster if done before calling self.sample :)
Can you also add variables argument to ``zonal_stats``?

hydromt/gis/raster.py
```diff
@@ -981,6 +985,140 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
 
         return ds_out
 
+    def sample_points(
```
hboisgon: Naming is tough ^^' but ``sample`` already sample points so I still find it confusing. Maybe ``sample_geoms`` so it sounds more that the expected geoms should be a polygon and because you call it sample it's clear we wont be doing zonal_stats?

hydromt/gis/raster.py
```diff
@@ -981,6 +985,140 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
...
+            # single-band or single-point: no reduction
+            if not dims_to_reduce:
+                out_vars[var] = da
+                continue
+
+            # multiple dims: apply stats
+            for stat in stats:
+                if stat in _STAT_REDUCERS:
+                    res = getattr(da, stat)(dim=dims_to_reduce)
+                    name = var if not dims_to_reduce else f"{var}_{stat}"
```
hboisgon: Do you need this if you already asserted dims_to_reduce is not None?

```suggestion
                    name = f"{var}_{stat}"
```

hydromt/gis/raster.py
```diff
@@ -981,6 +985,140 @@ def gen_zonal_stat(ds, geoms, stats, all_touched=False):
...
+        nodata = da.raster.nodata
+        if nodata is not None:
+            bad_mask |= (arr == nodata).any(dim=reduce_dims).values
+
+        if np.any(bad_mask):
+            exec_nodata_strat(
+                f"{int(np.sum(bad_mask))} out of {len(valid_indices)} geometries sampled nodata values.",
+                handle_nodata,
+            )
+            ds_out = ds_out.isel(index=~bad_mask)
```
hboisgon: Is the nodata strategy really interesting here? Or should we let the plugin deal with it after calling this method?
Note that this is now not valid. You are sampling a dataset possibly and not da where the nodata values can differ per variable that you sample. You'll probably need to check inside your above loop or create a new one here over sampled.data_vars

tests/gis/test_raster.py
```diff
@@ -21,6 +24,20 @@
 )
 from hydromt.readers import open_raster
 
+
+@pytest.fixture
+def sample_points_da():
```
hboisgon: Dont we already have such a fixture that you can reuse? See in the main conftest.py. Eg rioda, demda, lulcda, or even raster_ds with extra time dim to test on stats.

hydromt/gis/raster.py
LuukBlom: good catch, ruff automatically deleted this one even though there is the # noqa comment

hydromt/gis/raster.py
LuukBlom: I got warnings while testing about this, so i chose to raise with the idea to enforce users to do the reprojection themselves if they want it, or use representative point (which is a cheaply computed point guaranteed to be in the polygon.)
I can use `raster.planar_operation_in_utm`, but that will always silently reproject, reducing the performance benefit of this func. 
I can also accept that we just do it and let the warning be a problem for users to handle. 

What do you prefer?

hydromt/gis/raster.py
LuukBlom: Looking at it again, I agree with removing it and documenting the possibility of returning nodata values + that its the callers responsibility. zonal_stats doesnt do this either so its already similar in design. I will up-date the zonal_stats docstring also.

hydromt/gis/raster.py
```diff
@@ -892,19 +891,28 @@ def sample(self, gdf, wdw=0):
...
             GeoDataFrame with geometries
-        stats: list of str, callable
+        stats: str | list[str] | Callable
             Statistics to compute from raster values, options include
             {'count', 'min', 'max', 'sum', 'mean', 'std', 'median', 'q##'}.
-            Multiple percentiles can be calculated using comma-seperated values,
+            Multiple percentiles can be calculated using comma-separated values,
             e.g.: 'q10,50,90'. Statistics ignore the nodata value and are applied
-            along the x and y dimension. By default ['mean']
+            along the x and y dimension. By default 'mean'
```
hboisgon: stats is a mandatory argument actually in this one. Shall we edit?
```suggestion
            along the x and y dimension.
```

hboisgon (approved):
Looks great! Thanks @LuukBlom 

One last small thing but I approve already.

PR #1362, opened 2026-01-29
fix: respect `ext_override` when reading files in Driver classes

### LINKED ISSUE #1361: `ext_override` option not respected for netCDF files in `RasterDatasetXarrayDriver`
### HydroMT version checks

### Reproducible Example

The `ext_override` option in `RasterXarrayOptions` is not respected when reading netCDF files, causing files with non-standard extensions (like NWM's `.LDASIN_DOMAIN1`) to be skipped even when `ext_override: ".nc"` is set.

1. Create a data catalog with files that don't have `.nc` extension but are valid netCDF files:

```yaml
meta:
  roots:
    - /path/to/data
nwm_retro_meteo:
  data_type: RasterDataset
  driver:
    name: raster_xarray
    options:
      ext_override: ".nc"
  uri: meteo/nwm_retro/*.LDASIN_DOMAIN1
```

2. Try to load the dataset:

```python

dc = DataCatalog().from_yml("data_catalog.yml")
ds = dc.get_rasterdataset("nwm_retro_meteo")
```

### Current behaviour

All files are skipped because the driver compares each file's extension against `.nc` and skips files that don't match.

```
WARNING - Reading netcdf and /path/to/data/meteo/nwm_retro/202106110000.LDASIN_DOMAIN1 was not, skipping...
WARNING - Reading netcdf and /path/to/data/meteo/nwm_retro/202106110100.LDASIN_DOMAIN1 was not, skipping...
...
```

### Desired behaviour

When `ext_override` is set, all files matching the URI pattern should be read as netCDF files, regardless of their actual extension.

### Additional context

In [`raster_xarray_driver.py`](https://github.com/Deltares/hydromt/blob/478d81d669aad5b7b1d4b461597c31e666e9c7f1/hydromt/data_catalog/drivers/raster/raster_xarray_driver.py#L145), the zarr branch correctly checks `ext_override`:

```python
if ext != first_ext and not self.options.ext_override:
    logger.warning(f"Reading zarr and {_uri} was not, skipping...")
```

But the [netCDF branch](https://github.com/Deltares/hydromt/blob/478d81d669aad5b7b1d4b461597c31e666e9c7f1/hydromt/data_catalog/drivers/raster/raster_xarray_driver.py#L156) does NOT check `ext_override`:

```python
if ext != first_ext:
    logger.warning(f"Reading netcdf and {_uri} was not, skipping...")
```

A fix can be to change the netCDF branch to match the zarr branch behavior:

```python
if ext != first_ext and not self.options.ext_override:
    logger.warning(f"Reading netcdf and {_uri} was not, skipping...")
```
An undersirable workaround is creating symlinks with `.nc` extension pointing to the original files.

This affects reading NWM (National Water Model) files which use extensions like `.LDASIN_DOMAIN1` and `.CHRTOUT_DOMAIN1` but are valid netCDF4 files that can be opened directly with `xarray`:

```python
import xarray as xr
# This works fine
ds = xr.open_dataset("/path/to/202106110000.LDASIN_DOMAIN1")
```

LuukBlom:
## Issue addressed

Fixes #1361 

## Explanation

Updated the if statements to check for `ext_override`.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/.github/workflows/check-data-catalogs.yml
+++ b/.github/workflows/check-data-catalogs.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
-      - uses: prefix-dev/setup-pixi@v0.9.3
+      - uses: prefix-dev/setup-pixi@v0.9.4
         with:
           pixi-version: "v0.59.0"
       - name: Validate the catalogs
--- a/.github/workflows/create-release-artifacts.yml
+++ b/.github/workflows/create-release-artifacts.yml
@@ -77,7 +77,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
-      - uses: prefix-dev/setup-pixi@v0.9.3
+      - uses: prefix-dev/setup-pixi@v0.9.4
         with:
           pixi-version: "v0.59.0"
 
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -31,7 +31,7 @@ jobs:
     steps:
       - name: checkout code
         uses: actions/checkout@v6
-      - uses: prefix-dev/setup-pixi@v0.9.3
+      - uses: prefix-dev/setup-pixi@v0.9.4
         with:
           pixi-version: "v0.59.0"
 
--- a/.github/workflows/finish-release.yml
+++ b/.github/workflows/finish-release.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - id: download-artifact
-        uses: dawidd6/action-download-artifact@v11
+        uses: dawidd6/action-download-artifact@v14
         with:
           github_token: ${{secrets.GITHUB_TOKEN}}
           workflow: create-release-artifacts.yml
@@ -35,7 +35,7 @@ jobs:
     steps:
     - uses: actions/checkout@v6
     - id: download-artifact
-      uses: dawidd6/action-download-artifact@v11
+      uses: dawidd6/action-download-artifact@v14
       with:
         github_token: ${{secrets.GITHUB_TOKEN}}
         workflow: create-release-artifacts.yml
--- a/.github/workflows/pixi_auto_update.yml
+++ b/.github/workflows/pixi_auto_update.yml
@@ -14,7 +14,7 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Set up pixi
-        uses: prefix-dev/setup-pixi@v0.9.3
+        uses: prefix-dev/setup-pixi@v0.9.4
         with:
           pixi-version: "v0.59.0"
           run-install: false
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -44,7 +44,7 @@ jobs:
       - uses: actions/checkout@v6
         with:
           fetch-depth: 0
-      - uses: prefix-dev/setup-pixi@v0.9.3
+      - uses: prefix-dev/setup-pixi@v0.9.4
         with:
           pixi-version: "v0.59.0"
       - name: Test
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, windows-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ['311','312','313']
         dependencies: ['min', 'full']
 
@@ -50,7 +50,7 @@ jobs:
       cancel-in-progress: true
     steps:
       - uses: actions/checkout@v6
-      - uses: prefix-dev/setup-pixi@v0.9.3
+      - uses: prefix-dev/setup-pixi@v0.9.4
         with:
           pixi-version: "v0.59.0"
           environments: ${{ matrix.dependencies }}-py${{ matrix.python-version }}
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,12 +12,12 @@ repos:
       - id: debug-statements
       - id: mixed-line-ending
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.7
+    rev: v0.14.14
     hooks:
       - id: ruff
         args: [--fix, --exit-non-zero-on-fix]
       - id: ruff-format
   - repo: https://github.com/kynan/nbstripout
-    rev: 0.8.2
+    rev: 0.9.0
     hooks:
       - id: nbstripout
--- a/hydromt-core-sbom.json
+++ b/hydromt-core-sbom.json
@@ -6,20 +6,20 @@
     "creators": [
       "Tool: sbom4python-0.12.4"
     ],
-    "created": "2025-12-03T03:31:15Z",
+    "created": "2026-01-03T03:34:14Z",
     "licenseListVersion": "3.26"
   },
   "name": "Python-hydromt",
   "dataLicense": "CC0-1.0",
-  "documentNamespace": "http://spdx.org/spdxdocs/Python-hydromt-47eb4431-90d0-4f72-9b2e-0e403e847467",
+  "documentNamespace": "http://spdx.org/spdxdocs/Python-hydromt-f6b6e67d-fa1c-4de0-9d52-beb2e1541e68",
   "packages": [
     {
       "SPDXID": "SPDXRef-1-hydromt",
       "name": "hydromt",
-      "versionInfo": "1.2.0.dev0",
+      "versionInfo": "1.3.0.dev0",
       "primaryPackagePurpose": "APPLICATION",
       "supplier": "Organization: Dirk Eilander Helene Boisgontier Sam Vente (sam.vente@deltares.nl)",
-      "downloadLocation": "https://pypi.org/project/hydromt/1.2.0.dev0/#files",
+      "downloadLocation": "https://pypi.org/project/hydromt/1.3.0.dev0/#files",
       "filesAnalyzed": false,
       "licenseConcluded": "MIT",
       "licenseDeclared": "NOASSERTION",
@@ -40,12 +40,12 @@
         {
           "referenceCategory": "PACKAGE-MANAGER",
           "referenceType": "purl",
-          "referenceLocator": "pkg:pypi/hydromt@1.2.0.dev0"
+          "referenceLocator": "pkg:pypi/hydromt@1.3.0.dev0"
         },
         {
           "referenceCategory": "SECURITY",
           "referenceType": "cpe23Type",
-          "referenceLocator": "cpe:2.3:a:dirk_eilander_helene_boisgontier_sam_vente:hydromt:1.2.0.dev0:*:*:*:*:*:*:*"
+          "referenceLocator": "cpe:2.3:a:dirk_eilander_helene_boisgontier_sam_vente:hydromt:1.3.0.dev0:*:*:*:*:*:*:*"
         }
       ]
     },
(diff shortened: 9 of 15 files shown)


hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -96,7 +96,7 @@ def read(
             datasets = []
             for _uri in uris:
                 ext = splitext(_uri)[-1]
-                if ext != first_ext:
+                if ext != first_ext and not self.options.ext_override:
                     logger.warning(f"Reading zarr and {_uri} was not, skipping...")
                 else:
                     datasets.append(preprocessor(opn(_uri)))
```
deltamarnix: It seems like `self.options.get_ext_override()` is already solving this problem and we're checking twice for the same thing. Would it be easier to remove the `first_ext` check and use the `self.options.get_ext_override()` with a uri?

```suggestion
for _uri in uris:
  ext = self.options.get_ext_override(uri)
  if ext  == ".zarr":
    # do zarr stuff
  elif ext in [".nc", ".netcdf"]:
    # do netcdf stuff
  else:
    # warn
```

deltamarnix (changes requested):

hydromt/data_catalog/drivers/dataset/xarray_driver.py
LuukBlom: Please have a look at the current state, I think this is what we want.
Basically each `RasterDatasetDriver` and the `GeoDatasetXarrayDriver` now do this:

```python
        ...
        # Filter uris based on extension
        filtered_uris = []
        for _uri in uris:
            ext = self.options.get_reading_ext(_uri) 
            # if `self.options.ext_override` is set, this will always return the override.
            # So, it will be equal and added to the filtered uri list. 
            # If the override is not set, get_reading_ext returns the uri's file extension, which can be different than first_ext.
            # So this might filter some unwanted files from the uri list.
            
            # Question: Is this still what we want or is this old behaviour that we no longer want to support? 
            # We could also read all of the files without any filtering. 
 
            if ext != first_ext:
                logger.warning(
                    f"Reading {first_ext} and {_uri} has a different extension, skipping..."
                )
            else:
                filtered_uris.append(_uri)

        # Read and merge based on first_ext
        ...
```

@hboisgon can you explain why the filtering was added to the Drivers?
I would say that filtering the uri's is not something that a Driver should do.
The driver should just be given a list of uri's by the URI Resolver and then it should read them.
But I can imagine it was added for a good reason that I just dont know about.

hydromt/data_catalog/drivers/geodataset/xarray_driver.py
```diff
@@ -90,28 +89,25 @@ def read(
...
             opn = partial(xr.open_zarr, **self.options.get_kwargs())
-            datasets = []
-            for _uri in uris:
-                ext = splitext(_uri)[-1]
-                if ext != first_ext:
-                    logger.warning(f"Reading zarr and {_uri} was not, skipping...")
-                else:
-                    datasets.append(preprocessor(opn(_uri)))
-
+            datasets = [preprocessor(opn(_uri)) for _uri in filtered_uris]
```
deltamarnix: I am not an expert on partial functions, but since we call the partial here right away, I think it can be removed.

```
datasets = [preprocessor(xr.open_zarr(_uri, **self.options.get_kwargs()) for _uri in filtered_uris]
```

deltamarnix (changes requested):
The implementation looks nice, but can you write me some tests?

hydromt/data_catalog/drivers/dataset/xarray_driver.py
deltamarnix: The driver in this case is a combined driver. We are looking at XarrayDriver, so the technology is leading, not the type of data. We could decide to split the drivers to ZarrDriver and NetCdfDriver.

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -90,27 +93,35 @@ def read(
...
-                ext = splitext(_uri)[-1]
-                if ext != first_ext:
-                    logger.warning(f"Reading zarr and {_uri} was not, skipping...")
-                else:
-                    datasets.append(preprocessor(opn(_uri)))
+        first_ext = self.options.get_reading_ext(uris[0])
 
+        # Determine reading extensions based on first file
+        if first_ext in _NETCDF_EXT:
+            reading_extentions = set(_NETCDF_EXT)
```
deltamarnix: I see different implementations in the different drivers. In this version you use `set(_NETCDF_EXT)`, while in geodataset driver you redefine `reading_extentions = {".nc", ".netcdf"}`. Maybe it's easiest to have both `_ZARR_EXT` and `_NETCDF_EXT` as a `set` and just write `reading_extensions = _ZARR_EXT`?

deltamarnix: Then the check below can always be `if ext in reading_extensions`

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -90,27 +93,35 @@ def read(
...
-                    datasets.append(preprocessor(opn(_uri)))
+        first_ext = self.options.get_reading_ext(uris[0])
 
+        # Determine reading extensions based on first file
+        if first_ext in _NETCDF_EXT:
+            reading_extentions = _NETCDF_EXT
+        elif first_ext in _ZARR_EXT:
+            reading_extentions = _ZARR_EXT
+        else:
+            raise ValueError(f"Unknown extension for DatasetXarrayDriver: {first_ext}")
```
deltamarnix:
```suggestion
        if first_ext not in SUPPORTED_EXTENSIONS:
            raise ValueError(f"Unknown extension for DatasetXarrayDriver: {first_ext}")
```

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -90,27 +93,35 @@ def read(
...
+
+        # Filter uris based on extension
+        filtered_uris = []
+        for _uri in uris:
+            ext = self.options.get_reading_ext(_uri)
+            if ext not in reading_extentions:
```
deltamarnix: I'm wondering if `reading_extentions` is a bit too much. Can I be as simple as checking that `first_ext` is the same as `ext`? Since we use the same function for it.

```suggestion
            if ext != first_ext:
```

deltamarnix (changes requested):
Sorry Luuk, just wondering if this can be even a little simpler. It feels like we're doing too many variable updates and it looks like the code can be simpler for future us.

hydromt/data_catalog/drivers/dataset/xarray_driver.py
LuukBlom: This suggestion is actually close to the previous implementation.

We need to have the variable `reading_extensions` for this case: 

Driver where ext_override is not defined, gets called with uris: `["file.nc", "file.netcdf"]`

We would expect it to read both files, but because `first_ext` was used, and `.nc != .netcdf`, the second file is skipped and not read.

hydromt/data_catalog/drivers/dataset/xarray_driver.py
LuukBlom: see https://github.com/Deltares/hydromt/pull/1362#discussion_r2783420454

hydromt/data_catalog/drivers/dataset/xarray_driver.py
deltamarnix: I'm not sure if that is an issue. If you have a set of netcdf files, I don't expect that one will be .nc and the other .netcdf if they are stored as a group.

The issue for ext_override to work is what you have completely custom extensions.

If you want to support this case, you could also let `get_reading_ext()` return `.nc` where `.netcdf` is the actual extension.

LuukBlom: > I'm not sure if that is an issue. If you have a set of netcdf files, I don't expect that one will be .nc and the other .netcdf if they are stored as a group.

Okay, but should we then implement a safe guard anyways since the expectation is to read them both anyways?

> If you want to support this case, you could also let `get_reading_ext()` return `.nc` where `.netcdf` is the actual extension.

Okay I will implement this. But tbh, I think its adding more complexity where its not really needed.

deltamarnix: I'm just looking for fewer if-statements and simpler read flow. If you think that the end result is more complex I would not do it.

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -1,48 +1,26 @@
...
 
 from hydromt.data_catalog.drivers.base_driver import (
     DRIVER_OPTIONS_DESCRIPTION,
-    DriverOptions,
 )
 from hydromt.data_catalog.drivers.dataset.dataset_driver import DatasetDriver
-from hydromt.data_catalog.drivers.preprocessing import Preprocessor, get_preprocessor
+from hydromt.data_catalog.drivers.xarray_options import (
+    _NETCDF_EXT,
+    _ZARR_EXT,
```
deltamarnix: Could be public now, since they are exported.
Maybe make `_xarray_options.py` private?

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -55,10 +33,10 @@ class DatasetXarrayDriver(DatasetDriver):
 
     name: ClassVar[str] = "dataset_xarray"
     supports_writing: ClassVar[bool] = True
-    SUPPORTED_EXTENSIONS: ClassVar[set[str]] = {".zarr", ".nc", ".netcdf"}
+    SUPPORTED_EXTENSIONS: ClassVar[set[str]] = _ZARR_EXT | _NETCDF_EXT
 
-    options: DatasetXarrayOptions = Field(
-        default_factory=DatasetXarrayOptions, description=DRIVER_OPTIONS_DESCRIPTION
+    options: XarrayDriverOptions = Field(
```
deltamarnix, line 39: I really like a shared class in this case. Nice how SonarQube sort of forced to rethink that.

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -90,27 +68,17 @@ def read(
...
-            datasets = []
-            for _uri in uris:
-                ext = splitext(_uri)[-1]
-                if ext != first_ext:
-                    logger.warning(f"Reading zarr and {_uri} was not, skipping...")
-                else:
-                    datasets.append(preprocessor(opn(_uri)))
-
+        io_format = self.options.get_io_format(uris[0])
+        filtered_uris = self.options.filter_uris_by_format(uris, io_format)
```
deltamarnix: In theory the `io_format` doesn't need to be passed to `filter_uris_by_format`, because `self.options` can do that for itself.

hydromt/data_catalog/drivers/dataset/xarray_driver.py
```diff
@@ -167,10 +138,10 @@ def write(
         path = Path(path)
         ext = path.suffix
         write_kwargs = write_kwargs or {}
-        if ext == ".zarr":
+        if ext in _ZARR_EXT:
```
deltamarnix: Can we use the same check here like we do in `read()` with `get_io_format`?

hydromt/data_catalog/drivers/geodataset/xarray_driver.py
```diff
@@ -90,28 +89,17 @@ def read(
...
-                if ext != first_ext:
-                    logger.warning(f"Reading zarr and {_uri} was not, skipping...")
-                else:
-                    datasets.append(preprocessor(opn(_uri)))
-
+        io_format = self.options.get_io_format(uris[0])
+        filtered_uris = self.options.filter_uris_by_format(uris, io_format)
+
+        # Read and merge
+        if io_format == "zarr":
```
deltamarnix: Instead of a string comparison you could decide to go for an enum to avoid mistyping.

hydromt/data_catalog/drivers/raster/raster_xarray_driver.py
```diff
@@ -217,9 +186,9 @@ def write(
                 f"Unknown extension for RasterDatasetXarrayDriver: {ext},"
                 "switching to zarr"
             )
-            path = no_ext + _ZARR_EXT
-            ext = _ZARR_EXT
-        if ext == _ZARR_EXT:
+            path = no_ext + next(iter(_ZARR_EXT))
+            ext = next(iter(_ZARR_EXT))
```
deltamarnix: Does it make sense to swap these two lines?
```suggestion
            ext = next(iter(_ZARR_EXT))
            path = no_ext + ext
```

PR #1365, opened 2026-02-05
fix: parsing relative paths in workflow yml

### LINKED ISSUE #1350: Parsing relative data in the global section of the workflow yaml
### HydroMT version checks

### Reproducible Example

Workflow yaml starting with:

``` yaml
global:
  data_libs:
    - "examples/local_sources.yml"

```

### Current behaviour

The relative path to data_libs is not parsed. It will then be interpreted as relative to the working dir and not relative to the workflow file itself.

### Desired behaviour

Parse to a relative path based on the path of the workflow yml.

### Additional context

_No response_

LuukBlom:
## Issue addressed

Fixes #1350 

## Explanation

Added the validator `resolve_paths_or_catalogs` in `HydromtGlobalConfig`, which is used to validate the workflow yaml.

Refactored `read_workflow_yaml` to now actually return `HydromtModelSetup` instead of 3 dicts.

Moved all of the validation of the steps from the `Model` class into the step classes in `model_config.py`
Now has:
- `RawStep`: this class validates the structure of the yaml, no runtime function binding yet.
- `HydromtModelStep`: class that validates the `hydromt_step` part, has the function as an attribute. 

## Cyclic import errors

While fixing this, I encountered some problematic cyclic import errors, and after some thought I have re-introduced `hydromt.io` again.

The problem is that some of the readers depend on other parts of the hydromt codebase, when they should only read the files and return content. This means importing any reader (even just `read_toml`), will import a lot of other hydromt modules.

My idea for the separation: 
- `_io`: just file io, should not depend / know about any other part of hydromt
- `parsers`: take in raw data and return validated hydromt objects/classes
- `readers`: combine `io` and `parsers` to read and validate files, also exposes _io functions like `read_toml` for backwards compatibility

This separation means we can import the low-level _io of `read_yaml` or `read_toml` pretty much everywhere in hydromt since they have no hydromt dependencies.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist


hydromt/_validators/data_catalog_v0x.py
```diff
@@ -19,7 +19,7 @@
 from pyproj.exceptions import CRSError
 
 from hydromt import __version__ as HYDROMT_VERSION
-from hydromt.readers import _yml_from_uri_or_path
+from hydromt._io import _yml_from_uri_or_path
```
deltamarnix: If the module is protected, I would assume it's okay to make the function public.

hydromt/model/example/example_model.py
```diff
@@ -38,20 +40,29 @@ class ExampleModel(Model):
...
-                self,
-                filename=str(config_filename),
-            ),
-            "grid": ExampleGridComponent(self),
-        }
+        self.config = ConfigComponent(
+            self,
+            filename=str(config_filename),
+        )
+        self.grid = ExampleGridComponent(self)
```
deltamarnix: I prefer getter functions, because it clarifies that callers shouldn't override. Otherwise it's a public variable that can easily be overridden.

hydromt/model/model.py
```diff
@@ -476,8 +470,11 @@ def read(self, components: Optional[List[str]] = None) -> None:
             c.read()
 
     @staticmethod
-    def _options_contain_write(steps: List[Dict[str, Dict[str, Any]]]) -> bool:
-        return any("write" in next(iter(step_dict)) for step_dict in steps)
+    def _steps_contain_write(steps: list[RawStep]) -> bool:
+        for step in steps:
+            if "write" in step.name:
+                return True
+        return False
```
deltamarnix:
```suggestion
        return any("write" in step.name for step in steps)
```

hydromt/readers.py
```diff
@@ -983,32 +917,66 @@ def open_ncs(
     return ncs
 
 
-def read_yaml(path: str | Path) -> dict[str, Any]:
```
deltamarnix, line 986: I think it's fine, but you removed `read_yaml` and `read_toml` from the public API. Did you check with users or PO?

hydromt/parsers.py
```diff
@@ -0,0 +1,87 @@
...
+    ]
+
+    return HydromtModelSetup(
+        modeltype=PLUGINS.model_plugins.get(resolved_modeltype, resolved_modeltype),
+        globals_=global_cfg,
+        steps=raw_steps,
+    )
+
+
+def _deep_update(base: dict, override: dict) -> dict:
```
deltamarnix: We already have this function in _utils/deep_merge.py

hydromt/readers.py
deltamarnix: The functions are single-line functions almost, so we might say that the users of these functions can copy paste it if they want.

deltamarnix (changes requested):
Let's have a look at the names of the config classes.
We also talked about what `Model` should know about. Should it know about the config classes, or should the config classes know about the model? Perhaps don't convert during the validation, but only validate and keep the data raw.

hydromt/readers.py
LuukBlom: I agree they are not huge functions and they are copy-pastable / not really important.
They are currently part of the exposed api, so I left them.
They are still imported and exposed in `hydromt.readers` so for users, this will not change anything. 
For use inside of hydromt, it is nice that we can now import `read_toml` etc without requiring the importing of the entire `hydromt.readers` module.

examples/working_with_models.ipynb
hboisgon, line 1: Depending on what we choose but there are 4 pages in the docs that need updating (search for read_workflow_yaml) + update changelog
Also should API be updated? Users will need to know what object they are getting when calling read_workflow_yaml so this should be documented.

hydromt/cli/main.py
```diff
@@ -9,20 +9,19 @@
 from typing import Any, Optional
 
 import click
-import numpy as np
 from pydantic import ValidationError
 
 from hydromt import __version__
+from hydromt._io import read_yaml
```
hboisgon, line 15: Better keep using readers to import ``read_yaml`` then if I understand well?

hydromt/cli/main.py
```diff
@@ -240,26 +239,30 @@ def build(
...
-        mode = "w+" if fo else "w"
-        if modeltype not in PLUGINS.model_plugins:
-            raise ValueError("Unknown model")
-        mod = PLUGINS.model_plugins[modeltype](
+        mod = workflow.modeltype(
             root=model_root,
             mode=mode,
-            data_libs=data_libs,
-            **kwargs,
+            **workflow.globals_.model_dump(),
```
hboisgon, line 260: Not sure I am fond of ``model_dump`` here. Can we simplify for beginner python users? This is basic hydromt use building a model, I dont want to make it too complex at this stage.

hydromt/model/components/config.py
```diff
@@ -5,10 +5,10 @@
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Dict, Optional, Tuple, Union, cast
 
+from hydromt._io import read_toml, read_yaml
```
hboisgon, line 8: Guess you can undo and go back to readers. Most likely plugins will copy and modify this component so we should limit calls to private functions

hydromt/model/example/example_model.py
hboisgon, line 1: Why any change in this file? I dont think we want model components to be initialised this way. ``_config`` is private for a reason. Same for the property definition, not sure we want to envourage plugins to define their components this way.

hydromt/model/model.py
```diff
@@ -6,15 +6,28 @@
 import shutil
 import typing
 from abc import ABCMeta
-from inspect import _empty, signature
 from os.path import isabs, isfile, join
 from pathlib import Path
-from typing import Any, Dict, List, Optional, TypeVar, Union, cast
+from typing import (
```
hboisgon, line 11: Instead of adding more lines, can we get rid of Dict, List, Optional, Union in this file?

hydromt/model/model.py
```diff
@@ -6,15 +6,28 @@
...
+    TypeVar,
+    Union,
+    cast,
+)
 
 import geopandas as gdp
 from pyproj import CRS
 
-from hydromt._utils import _rgetattr, _validate_steps, log
+from hydromt._io import read_yaml
```
hboisgon, line 24: Back to ``readers`` import

hydromt/model/model.py
```diff
@@ -141,13 +153,25 @@ def _add_components(self, components: Dict[str, Any]) -> None:
         for name, value in components.items():
             if isinstance(value, ModelComponent):
                 self.add_component(name, value)
+            elif isinstance(value, HydromtComponentConfig):
+                self.add_component(
+                    name, value.type(self, **value.model_dump(exclude={"type", "name"}))
+                )
```
hboisgon, line 159: Why this addition? Was adding components using a workflow yaml not working?

hydromt/model/model.py
```diff
@@ -141,13 +153,25 @@ def _add_components(self, components: Dict[str, Any]) -> None:
...
             elif isinstance(value, dict):
                 type_name = value.pop("type")
-                component_type = PLUGINS.component_plugins[type_name]
+                if type_name in PLUGINS.component_plugins:
```
hboisgon, line 162: Hum... we did not force plugins to register their custom components so this will break all our published v1 plugins I think. At least wflow only exposes ``WflowSbmModel`` and ``WflowSedimentModel`` not ``WflowConfigComponent`` etc... Is this something we really want to enforce?

hydromt/model/model.py
```diff
@@ -299,32 +323,17 @@ def build(
...
-                if kwargs:
-                    merged.update(**kwargs)
-                for k, v in merged.items():
-                    logger.info(f"{step}.{k}={v}")
 
-                method(**merged)
+            # Execute steps
+            steps = steps or []
+            raw_steps = [create_raw_step(step) for step in steps]
+            bound_steps = [step.to_model_step(self) for step in raw_steps]
```
hboisgon, line 330: Can you maybe add a comment here of what these two lines are doing and how do they differ? Is there a reason to keep two lines here or can we combine into one to to simplify build and update?

hydromt/model/model.py
```diff
@@ -299,32 +323,17 @@ def build(
         """
         with log.to_file(Path(self.root.path) / "hydromt.log"):
             log.log_version()
-            steps = steps or []
-            _validate_steps(self, steps)
```
hboisgon, line 303: I still have to review what the below functions are doing but it can be risky to remove validation here.

Example: region is not in the command line anymore in v1. So for automation of model building re-using exactly the same settings but applied to a different region, region needs to be manually added to the steps in between ``read_workflow_yaml`` and ``model.build``.

hydromt/parsers.py
hboisgon, line 1: Should this go to model folder or somewhere else? To keep the root sort of organised. (same for _io.py)
Or maybe in model/steps? Seems the python classes should not be hidden in validators especially if these are the return types of read_workflow_yaml which users will have to interact with?

hydromt/readers.py
```diff
@@ -983,32 +917,66 @@ def open_ncs(
...
-def read_toml(path: str | Path) -> dict[str, Any]:
-    """Read toml file and return as dict."""
-    with open(path, "rb") as f:
-        data = tomli.load(f)
+    Returns
+    -------
+    HydromtModelSetup
+        A dataclass containing the model type, global configuration, and runtime steps.
+        Access the model type with `.modeltype`, the global configuration with `.globals_`,
+        and the runtime steps with `.steps`.
```
hboisgon, line 947: Not clear what `modeltype`, `globals` and `steps` are: str, ?, dict? needs documentation of the HydroMTModelSetup class in public API

hboisgon: I guess ``.global`` is not possible? Then why not ``.global_`` to try and keep close to the workflow yaml? global goes to model __init__ so could also be ``.model_init``?

hydromt/parsers.py
```diff
@@ -0,0 +1,75 @@
...
+        The validated HydroMT model setup.
+    """
+    # Merge defaults
+    if defaults:
+        data = deep_merge(defaults, data)
+
+    # Determine model type
+    resolved_modeltype = modeltype or data.get("modeltype")
+    if resolved_modeltype is None:
+        raise ValueError("Model type not specified in workflow or arguments")
```
hboisgon, line 52:
```suggestion
        raise ValueError("Model type not specified in workflow YAML or arguments")
```

hydromt/parsers.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # Context for path resolution
+    context = {}
+    if (
+        abs_path
+        and root is not None
+        and (skip_abspath_sections is None or "global" not in skip_abspath_sections)
```
hboisgon, line 59: I'm having trouble following but what happens if "global" in skip_abspath_sections? Does this means that relative paths in steps will not be converted? Ie using a local data file instead of catalog entry for the setup functions?

hydromt/parsers.py
```diff
@@ -0,0 +1,75 @@
...
+    if defaults:
+        data = deep_merge(defaults, data)
+
+    # Determine model type
+    resolved_modeltype = modeltype or data.get("modeltype")
+    if resolved_modeltype is None:
+        raise ValueError("Model type not specified in workflow or arguments")
+
+    # Context for path resolution
+    context = {}
```
hboisgon, line 55: Do you expect anything else in context than just root? If not why make it a dictionnary?

hydromt/_validators/model_config.py
hboisgon, line 1: Reviewed in the other PR. See comments there.

hboisgon (commented):
Some food for thoughts and a couple of warnings on the current implementation.

As highlighted in the ther PR, I think we need a design session about in general how do we build a model with hydromt before we can really finalise this PR. This is major change for python users for now if we do this we should do it well and avoid changing (again) in the future.

PR #1370, opened 2026-02-09
Fix: zero as nodata in full_like

dalmijn:
## Issue addressed

Fixes same as #1366 

## Explanation

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/.github/workflows/check-data-catalogs.yml
+++ b/.github/workflows/check-data-catalogs.yml
@@ -27,7 +27,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v6
-      - uses: prefix-dev/setup-pixi@v0.9.3
+      - uses: prefix-dev/setup-pixi@v0.9.4
         with:
           pixi-version: "v0.59.0"
       - name: Validate the catalogs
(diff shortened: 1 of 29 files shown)


hydromt/gis/raster_utils.py
LuukBlom, line 229: I like these updates, but would prefer to fix it in a way that will not bite us in the future.
I think this will error in this case: 
nodata would be set to `np.nan` (which is a float), but the dtype would be resolved to `int`

Perhaps a helper function like this could be nice:
```python
def _determine_nodata_value(*candidates, fallback: Any) -> Any:
    for c in candidates:
         if c is not None:
             return c
    return fallback
```

Could also add the `dtype` to this helper function

LuukBlom (changes requested):
Nice bug finding! 

I think with the addition of a helper func and some more updates to other parts of the file as well, this can be merged quickly.

hydromt/gis/raster_utils.py
dalmijn: I like the function, though I do think that it should just error in your case. If one tries to set nodata to nan and dtype to int, it's a user error in my opinion.

LuukBlom (approved):

PR #1372, opened 2026-02-10
feat: cleanup config classes

LuukBlom:
This PR is meant as a suggested improvement to https://github.com/Deltares/hydromt/pull/1365 to make reviewing these changes easier.

## Issue addressed

Fixes #<issue number>

## Explanation

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/hydromt/_validators/model_config.py
+++ b/hydromt/_validators/model_config.py
@@ -4,12 +4,13 @@ import inspect
 import logging
 from keyword import iskeyword
 from pathlib import Path
-from typing import Any, Callable, Protocol, Type
+from typing import Any, Callable, Type
 
 from pydantic import (
     BaseModel,
     ConfigDict,
     Field,
+    PrivateAttr,
     field_serializer,
     field_validator,
     model_validator,
@@ -20,13 +21,16 @@ from hydromt.plugins import PLUGINS
 logger = logging.getLogger(__name__)
 
 
-class Model(Protocol):
-    components: dict[str, Any]
+class ComponentSpec(BaseModel):
+    """Represents the specification for a single model component.
 
+    The `name` field is the name of the component, which will be used to reference it in workflow steps.
+    The `type` field specifies the component class, which can be given as a string referring to a registered component plugin or as the class itself.
+    The remaining fields are arbitrary and will be passed as kwargs to the component constructor when the model is instantiated.
+    """
 
-class HydromtComponentConfig(BaseModel):
     name: str
-    type: Type
+    type: Type  # perhaps rename attr to `component_type`
 
     model_config = ConfigDict(extra="allow", arbitrary_types_allowed=True)
 
@@ -47,11 +51,19 @@ class HydromtComponentConfig(BaseModel):
         return v
 
 
-class RawStep(BaseModel):
-    """Represents the YAML structure BEFORE runtime wiring."""
+class WorkflowStep(BaseModel):
+    """Represents a single step in the workflow, which will be executed as part of the model run.
+
+    The `name` field specifies the function to call, which can be in the format `<component_name>.<function>` or just `<function>` for top-level model methods.
+    The `kwargs` field contains the keyword arguments to pass to the function when executing the step.
+    The `_method` field is populated at runtime when the step is bound to a model instance by calling `bind_to_model`, and is not part of the input validation.
+    """
 
     name: str
-    args: dict[str, Any] = Field(default_factory=dict)
+    kwargs: dict[str, Any] = Field(default_factory=dict)
+
+    # runtime-only, set with `bind_to_model`
+    _method: Callable[..., Any] | None = PrivateAttr(default=None)
 
     model_config = ConfigDict(extra="forbid")
 
@@ -59,9 +71,21 @@ class RawStep(BaseModel):
     @classmethod
     def unpack(cls, v):
         if not isinstance(v, dict):
-            raise ValueError("Each step must be a single-key dictionary")
-        name, args = next(iter(v.items()))
-        return {"name": name, "args": args or {}}
+            raise ValueError("Each step must be a dictionary")
+
+        # already normalized shape
+        if "name" in v:
+            return v
+
+        # YAML shape: exactly one key, value is kwargs
+        if len(v) != 1:
+            raise ValueError(
+                "Each step must be a single-key dictionary or contain 'name' and 'kwargs'"
+            )
+
+        # dict with one key as name and value as kwargs
+        name, kwargs = next(iter(v.items()))
+        return {"name": name, "kwargs": kwargs or {}}
 
     @field_validator("name", mode="after")
     def validate_name(cls, name: str) -> str:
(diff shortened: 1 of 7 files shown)


hydromt/_validators/model_config.py
hboisgon, line 1: See my comment in the other PR. I think users should know the content of the these classes if this is what they get when calling the ``read_workflow_yaml`` function. This should move into hydromt.model and be renamed to model_workflow.py to avoid confusion with the ConfigComponent

hydromt/_validators/model_config.py
```diff
@@ -71,7 +95,20 @@ def validate_name(cls, name: str) -> str:
...
+    def bind_to_model(self, model_instance: type) -> None:
+        """Bind this workflow step to a specific model instance.
+
+        This is done by resolving the function to call based on self.name and the model's components.
+        """
+        if not hasattr(model_instance, "components"):
+            raise ValueError(
+                f"Model instance of type {type(model_instance).__name__} does not have 'components' attribute"
+            )
+        elif not isinstance(model_instance.components, dict):
```
hboisgon, line 107: Is this case at all possible?

hydromt/_validators/model_config.py
```diff
@@ -47,21 +51,41 @@ def validate_type(cls, v):
         return v
 
 
-class RawStep(BaseModel):
-    """Represents the YAML structure BEFORE runtime wiring."""
+class WorkflowStep(BaseModel):
```
hboisgon, line 54: I thought ``HydroMTModelStep`` was clearer also with the @hydromt_step decorator. Again steps can be defined and called outside of the workflow yaml so we maybe are too restrictive by using the name Workflow here?

hydromt/_validators/model_config.py
```diff
@@ -89,51 +126,40 @@ def to_model_step(self, model_instance: Model) -> "HydromtModelStep":
...
                 f"Step function '{self.name}' argument validation failed: {e}"
             )
-        return self
+        self._method = fn
 
     def execute(self):
-        kwargs = self.args or {}
-
+        """Execute this workflow step by calling the bound function with the provided arguments."""
+        if self._method is None:
```
hboisgon, line 144: Can't you first try and call bind_to_model and then if still None return an error? I'm a little fuzy about this works still when calling model.build and model.update sorry.

hydromt/_validators/model_config.py
```diff
@@ -89,51 +126,40 @@ def to_model_step(self, model_instance: Model) -> "HydromtModelStep":
...
             logger.info(f"{self.name}.{k}={v}")
+        self._method(**kwargs)
 
-        sig = inspect.signature(self.fn)
-        bound_args = {
-            param: p.default
-            for param, p in sig.parameters.items()
-            if p.default != inspect._empty
-        }
-        bound_args.update(kwargs)
```
hboisgon, line 130: Is this still happening somewhere? It's important for reproducibility if the complete function arguments used appera in the log and not just the one the user provided/changed.

hydromt/_validators/model_config.py
```diff
@@ -89,51 +126,40 @@ def to_model_step(self, model_instance: Model) -> "HydromtModelStep":
...
+    """Represents the 'global' part of the workflow that defines the model configuration.
+
+    The attributes of this instance will be passed as kwargs to the model constructor,
+    so it is designed to be flexible and allow for any keys.
+    """
 
-class HydromtGlobalConfig(BaseModel):
-    components: list[HydromtComponentConfig] = Field(default_factory=list)
-    data_libs: list[str | Path] = Field(default_factory=list)
+    components: list[ComponentSpec] = Field(default_factory=list)
```
hboisgon, line 159: I never looked into this before but why a list? In both model init and in the workflow yaml they are defined as dictionnaries so why change the type here?

hydromt/_validators/model_config.py
```diff
@@ -172,10 +198,21 @@ def serialize_components_field(self, v, info):
...
-class HydromtModelSetup(BaseModel):
-    modeltype: type
-    globals_: HydromtGlobalConfig = Field(alias="global")
-    steps: list[RawStep]
+class WorkflowSpec(BaseModel):
+    """Represents the entire workflow, including the model type, global model configuration and the list of workflow steps.
+
+    The `modeltype` field specifies the model class to instantiate, which can be given as a string referring to a registered model plugin or as the class itself.
+    The `globals_` field contains the model configuration that will be passed as kwargs to the model constructor when instantiating the model.
+    The `steps` field is a list of workflow steps that will be executed in order as part of the model run.
```
hboisgon, line 206: I dont think the documentation is clear enough for users to be able to use whatever comes out of ``read_workflow_yaml``.

modeltype is "wflow_sbm" or ``WflowSbmModel`` (second).

globals: will any argument of model.__init__ be a property that you can call and keep the type model __init__ expects? I see components is a list here and model __init__ wants a dict? WflowSbmModel has extra a ``config_filename``. How is this dealt with? If a model has extra arguments in __init__ that also should be converted from relative to absolute paths how will this work? Do plugins need to extend this class? If so this should be documented in the plugins docs.

steps: steps won't really be a list of dict anymore right? already the step name is added with name property to the arguments dict? How can users modify steps arguments before calling model.build or model.update if they need to? Not that I mind working with python classes instead of dict but I do need documentation then.

hydromt/_validators/model_config.py
```diff
@@ -172,10 +198,21 @@ def serialize_components_field(self, v, info):
         }
 
 
-class HydromtModelSetup(BaseModel):
-    modeltype: type
-    globals_: HydromtGlobalConfig = Field(alias="global")
-    steps: list[RawStep]
+class WorkflowSpec(BaseModel):
```
hboisgon, line 201: Same as above, I'm not sure workflow is the right name? Is ``ModelStepSpec`` or ``StepSpec`` clearer.

hboisgon (commented):
Comments are mixed between this PR and the other one. I saw this one after I started reviewing. Still not fully sure what to think about this refactor. I guess better to validate when reading the yaml before calling build and update to get errors a little earlier but wonder if this does not make it harder to use for our users. Definitely requires docs.

Also would be nide to have a step back on what we are trying to do and steps to build/update a model with hydromt and terminology?

It was an important comment when we had an external audit on the code but the CLI build function is not linked to a model function ie ``CLI hydromt build`` does extra things including ``read_workflow_yaml`` which ``model.build`` does not do. Maybe moving validation around now highlights we should definitely work on this. Hence would be good to step back from the implementation and think again about what we are trying to do when building a model.

deltamarnix:
I haven't done a full review myself yet, but I did have a conversation with Luuk before he started on this PR.

I understand Helene's concern that validation is now separate from the actual model building. One of the things that I discussed with Luuk was to make the core `Model` base class a `pydantic.BaseModel`. And then we could make it so that the yaml is directly read into a Model, or WflowModel, or whatever plugin model. Then we can get rid of all these workflow classes, because the components are pydantic serializable themselves. But that would be quite a change, because it might have an impact on the plugins too. Although I do think that would be a nice addition. It will simplify the whole reading process.

PR #1373, opened 2026-02-12
fix: relative paths for data_libs should be made absolute

### LINKED ISSUE #1350: Parsing relative data in the global section of the workflow yaml

LuukBlom:
## Issue addressed

Fixes #1350 

## Explanation

This is a minimal fix for the resolving of the relative to absolute paths.

Essentially what is done when reading a workflow.yml file at `ROOT/workflow.yml`:
- read in the yml as dict
- recursively go through the values of the entire dict (skip keys based `skip_abs_path_section`)
- for every string `x`: convert it to `Path(ROOT, x)`
- if the path exists, we keep the conversion, if it doesn't we leave the string untouched.

I don't like the current design, which is why I started changing it in the other PR. 
What I don't like about it is that we are linking a file's existence on disk to whether or not a path should be absolute.
And that we are changing the read in data before we actually validate it. Why not just change it during the validation like pydantic is supposed to be used. That way we will also have access to the expected format and types of it all, instead of a recursive dict approach that is difficult to extend/debug/update.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,7 @@ New
 
 Changed
 -------
+- ``open_ncs`` returns either ``dict[Path, xr.Dataset]`` or ``dict[str, xr.Dataset]``. When the uri contains no ``{name}`` placeholder, the full paths are used as keys. If there is a placehodler the key is the value of the ``{name}`` placeholder.
 -  Instead of skipping the ``global`` section by default, ``read_workflow_yaml`` now tries to convert all strings in all sections into absolute paths by default, by checking if the absolute path exists.
 
 Fixed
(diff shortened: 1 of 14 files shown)


LuukBlom:
Discussed with @hboisgon and this can be merged now

PR #1377, opened 2026-02-16
fix: implement 'expand_uri_paths' function and update related components

### LINKED ISSUE #1376: hydromt.readers.open_ncs does not open all files correctly
### HydroMT version checks

### Reproducible Example

Wildcard with * and not {} in filename_template
eg filename_template = 'forcing/*.nc'

### Current behaviour

All paths are found, but because there is no regex, ``name`` is '' so all files will be overwritten in the output ``ncs`` dict instead of being all listed:

``` python
    ncs = {}
    path_template = root / filename_template


    path_glob, _, regex = _expand_uri_placeholders(
        str(path_template), placeholders=list(_placeholders)
    )
    paths = glob(path_glob)
    for path in paths:
        name = ".".join(regex.match(path).groups())  # type: ignore
        ds = open_nc(filepath=path, **kwargs)
        ncs[name] = ds
    return ncs
```

### Desired behaviour

All files are read so name should be unique (why not use just filename in the first place?)

### Additional context

This is very common to use * for wflow forcing if several forcing files should be read / written. Bumped into it when try WflowSbmModel.forcing.read()

LuukBlom:
## Issue addressed

Fixes #1376 

## Explanation

Instead of having to deal with the regex in when calling `_expand_uri_placeholders`, I have added a function `expand_uri_paths` (name can be improved) that always returns resolved paths and a valid name to use: `tuple(path, name)`.
So callers are no longer forced to deal with the regex.

Went over all functions that use `_expand_uri_placeholders` and updated if possible.
Also went through the Drivers and Components and updated if possible.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

Regex can be very complicated and difficult to understand.
So, its no guarantee that there will never be problems with this again.
But by adding some tests we can at least lock in the behaviour we want.

### DIFF
--- a/hydromt/data_catalog/uri_resolvers/convention_resolver.py
+++ b/hydromt/data_catalog/uri_resolvers/convention_resolver.py
@@ -22,7 +22,7 @@ from hydromt.typing import (
 logger = logging.getLogger(__name__)
 
 
-class SafeDict(dict):
+class _SafeDict(dict):
     """A dict that returns the placeholder string for missing keys instead of raising KeyError."""
 
     def __missing__(self, key):
@@ -60,9 +60,11 @@ class ConventionResolver(URIResolver):
         ) -> Iterable[str]:
             if pair[0] is not None:
                 return map(
-                    lambda uri: self.filesystem.get_fs().unstrip_protocol(uri)
-                    if not uri.startswith(pair[0])
-                    else uri,
+                    lambda uri: (
+                        self.filesystem.get_fs().unstrip_protocol(uri)
+                        if not uri.startswith(pair[0])
+                        else uri
+                    ),
                     pair[1],
                 )
             else:
@@ -158,7 +160,7 @@ class ConventionResolver(URIResolver):
         )
         uris: list[str] = list(
             self._resolve_wildcards(
-                map(lambda fmt: uri_expanded.format_map(SafeDict(**fmt)), fmts)
+                map(lambda fmt: uri_expanded.format_map(_SafeDict(**fmt)), fmts)
             )
         )
         if not uris:


tests/_utils/test_naming_convention.py
```diff
@@ -0,0 +1,89 @@
...
+
+def test_expand_uri_paths_wildcard_and_placeholder_not_unique_names_raises(
+    tmp_path: Path,
+):
+    root = tmp_path / "2020/temp"
+    root.mkdir(parents=True)
+    create_nc(root / "some_forcing.nc")
+    create_nc(root / "other_forcing.nc")
+
+    uri = str(tmp_path / "{year}/{variable}/*_forcing.nc")
```
hboisgon: Why shouldn't we be able to read this? the uri is valid in my sense and just because we can't create unique names is not a good enough reason no? (eg if * is month-day combination in your example). Do we really need to create names using placeholders? Is anything using these names later? 

Or are we just making things complex again for little to no reason? I think this ``uri_resolver`` with the data_catalog placeholders was more meant for ``data_catalog`` than for ``model`` and ``components``. Why not keep using the old way for geoms/tables and open_ncs instead of creating headaches for ourselves?

Eg write_geoms used to be in 0.10 or tables before this PR was the same:

```python
# fn = "geoms/*.geojson"
fns = glob.glob(join(self.root, fn))
        for fn in fns:
            name = basename(fn).split(".")[0]
            self.logger.debug(f"Reading model file {name}.")
            self.set_geoms(gpd.read_file(fn, **kwargs), name=name)
```
It's simple and does the job for core components I think? If plugins want other names, they can then create their own components.

But maybe there is a reason to make it complex that I cannot see?

hboisgon (commented):
Thanks for looking into this Luuk. I'm however not convinced this is the way to go and I get the feeling we are making things complex and I would like to understand if this is needed. You said regex is complex and to be honest I have difficulties reviewing your new functions so I'd be up to simplify.

1) Do we need this complexity for model components and open_ncs? I think not and we can go back to something simpler like tables before your PR?
2) I would think your path that errors in your test should be valid. As a user I would get very confused why it does not (eg files saved per semester or per day etc in your failing examples should not be a weird case we cannot read). For data catalog, is the created name ever used afterwards? If not can we just use the resolved filename as temporary name?

Or only if placeholder is {name} we use it as name for the dict and not the resolved filename as this is clearly user intended compared to other placeholders?

Sorry I really thought it was a one-liner for open_ncs... I think somewhere along the lines functionnality meant for data catalog was also used for model and not sure if this was smart.

tests/_utils/test_naming_convention.py
LuukBlom: See my other comment for my reasoning for this test.

I agree with you that this simple file globbing at the directory level is nice and simple. 
That is what `expand_uri_paths` does when there are no placeholders.
When there are no placeholders (e.g. no `{` in the uri), it skips the whole regex checks, and uses the filename without extension as the name.

LuukBlom:
Im also fine with enforcing the rule that components only use wildcards, and placeholders are only for the datacatalog.
That is not the case now and is not documented / wasnt clear to me.

LuukBlom:
Discussed with @hboisgon and implemented final comments.

hydromt/data_catalog/uri_resolvers/convention_resolver.py
```diff
@@ -22,6 +22,13 @@
 logger = logging.getLogger(__name__)
 
 
+class SafeDict(dict):
```
deltamarnix: Maybe good to keep this private? `_SafeDict`

PR #1385, opened 2026-02-26
Add CI workflow to test downstream plugin compatibility

LuukBlom:
## Issue addressed

Fixes #<issue number>

## Explanation

This PR adds a GitHub Actions workflow that tests downstream plugins against a hydromt release candidate. The workflow builds a hydromt wheel from the current repo, installs it into the plugin’s pixi environment, and runs the plugin test suite. This catches breaking changes before release and reflects how users install hydromt in practice.

# TODO

- hydromt sfincs pixi updates: https://github.com/Deltares/hydromt_sfincs/pull/346
- determine which other plugins to add? 

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/.github/workflows/downstream-plugins.yml
+++ b/.github/workflows/downstream-plugins.yml
@@ -4,7 +4,6 @@ on:
   push:
     tags:
       - "v*"
-    branches: [feat/release-test-automation-downstream-plugins]
   workflow_dispatch:
 
 permissions:


.github/workflows/downstream-plugins.yml
```diff
@@ -0,0 +1,68 @@
+name: Downstream plugin compatibility
+
+on:
+  push:
+    tags:
+      - "v*"
+  workflow_dispatch:
+
```
evetion, line 8: If you add this branch, or the this filepath trigger, could we run CI already in this PR?

.github/workflows/downstream-plugins.yml
```diff
@@ -0,0 +1,68 @@
...
+          # We install the wheel in two ways:
+          # 1. --no-deps (strict API) to ensure the plugin works with only the candidate wheel
+          # 2. with deps (default) to ensure it still works when all dependencies are installed
+
+          pixi install -e ${{ matrix.plugin.environment }}
+          if [ "${{ matrix.with_deps }}" = "true" ]; then
+              pixi run -e ${{ matrix.plugin.environment }} python -m pip install --force-reinstall "$HYDROMT_DIST"/hydromt-*.whl
+          else
+              pixi run -e ${{ matrix.plugin.environment }} python -m pip install --force-reinstall --no-deps "$HYDROMT_DIST"/hydromt-*.whl
+          fi
```
evetion: If we want these two steps, might be good to split them into seperate steps, so it's easy to see what failed exactly.

evetion:
Btw, if this works, it would be good for plugins to test their downstream projects in a similar way (if such repos exist).

.github/workflows/downstream-plugins.yml
LuukBlom: No I tried that and it wasnt working unfortunately.

For that to work, Google/stackoverflow told me that it needs to be in main first. Then when changes occur on a PR into main, the updated workflow from that feature branch will be used.

.github/workflows/downstream-plugins.yml
LuukBlom: The name of the job is `name: Test ${{ matrix.plugin.name }} (deps=${{ matrix.with_deps }})`.

So, when the matrix kicks off, we will see all combinations of each plugin + with/without deps. similar to the python versions + os for `test.yml`

And since the only diff between the two cases is whether the command has `--no-deps` or not, I thought that duplicating it all would be a bit overkill. 
Did you have something in mind?

evetion: I just meant a new -name item.

.github/workflows/downstream-plugins.yml
```diff
@@ -0,0 +1,110 @@
...
+        working-directory: plugin
+        run: |
+          pixi install -e ${{ matrix.plugin.environment }}
+          pixi run -e ${{ matrix.plugin.environment }} \
+            python -m pip install --force-reinstall --no-deps "$HYDROMT_DIST"/hydromt-*.whl
+
+      - name: Echo solved environment
+        run: |
+          # what pixi thinks is installed
+          echo "pixi list output:\n"
```
evetion:
```suggestion
          echo "pixi list output:"
```

PR #1398, opened 2026-03-09
Fix: model root set issues

dalmijn:
## Issue addressed

Fixes issues with `ModelRoot`

## Explanation

@michaelohanrahan has found an issue where when switching to write mode and another directory, that directory was not created on the fly as this was purely dealt with at initialisation instead of when setting a root. I therefore altered the `ModelRoot` class so it more heavily revolves around the `set` method. I also moved the `ModelMode` class to the `model` submodule as having it under typing didnt make sense to me. Also created a helper `_check_directory` function as a the similar method that existed in `ModelRoot` was stateless... Also created a bunch of tests as these were missing.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

### DIFF
--- a/tests/model/test_root.py
+++ b/tests/model/test_root.py
@@ -56,6 +56,19 @@ def test_model_root__assert_append(tmp_path: Path):
     root._assert_write_mode()
 
 
+def test_model_root__clean_up(tmp_path: Path):
+    # Set up the root
+    p = Path(tmp_path, "foo")
+    root = ModelRoot(path=p, mode="w")
+    # Assert it's there
+    assert p.exists()
+
+    # Call the method
+    root._cleanup()
+    # Assert the directory is gone
+    assert not p.exists()
+
+
 def test_model_root_read_check(tmp_path: Path):
     # Set up the root
     root = ModelRoot(path=tmp_path, mode="r")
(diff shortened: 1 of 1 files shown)


hydromt/_utils/path.py
```diff
@@ -4,7 +4,25 @@
...
+    root: Path | str | None = None,
+    fail: bool = True,
+) -> Path:
+    """Very simple check whether a directory exists."""
+    root = root or Path.cwd()
+    path = Path(root, path)
+    if not path.exists() and fail:
+        raise IOError(f"{path.as_posix()} does not exist")
+    path.mkdir(parents=True, exist_ok=True)
+    return path
```
LuukBlom: I dont love this function as a utils function tbh.
Could this be moved into the ModelRoot class? Its quite specific to the model root only.

Second, its called `_check_directory`, but it also creates directories.
If we want this, please add a kwarg `mkdir: bool = True` or something to communicate that.
And I would like a more descriptive name

```python
def _resolve_model_root_dir(
    path: Path | str,
    root: Path | str | None = None,
    mkdir: bool = True,
    must_exist: bool = False,
) -> Path:
    """Resolve a path relative to the model root and optionally create it."""
    root = Path(root) if root is not None else Path.cwd()
    path = (root / path).resolve()

    if must_exist and not path.exists():
        raise IOError(f"{path.as_posix()} does not exist")

    if mkdir:
        path.mkdir(parents=True, exist_ok=True)

    return path
```

PR #1399, opened 2026-03-10
fix: improve logging

LuukBlom:
Fixes [#708](https://github.com/Deltares/hydromt_wflow/issues/708)

Added exception logging to the ``log.to_file`` context manager to ensure any unhandled exceptions are logged to the file before closing it.

Initializing the logger and adding a console handler should not be done whenever hydromt is imported.
It should be done explicitely at the start of the application or script by the person.
If users dont do this, it will still work, there just will not be output in the terminal (which is what we want to be able to support)

## Explanation

move log out of utils and into toplevel
update ``initialize_logging``
dont call ``initialize_logging`` at import time.
update ``to_file``

## Tests

add setup logging sessions fixture
put all static paths into session scoped fixtures and update usage everywhere.
added ``test_log.py`` to lock in the contextmanager behaviour.

### DIFF
--- a/hydromt/log.py
+++ b/hydromt/log.py
@@ -187,7 +187,9 @@ def to_file(
         handler.close()
 
 
-def flags_to_level(verbose: int = 0, quiet: int = 0) -> int:
+def flags_to_level(
+    verbose: int = 0, quiet: int = 0, default: int = logging.WARNING
+) -> int:
     """Compute the log level based on the number of verbose and quiet flags.
 
     Each ``-v`` (verbose) flag decreases the log level by 10 (e.g., from WARNING to INFO to DEBUG),
@@ -199,6 +201,8 @@ def flags_to_level(verbose: int = 0, quiet: int = 0) -> int:
         Number of verbose flags (e.g., -v)
     quiet : int, optional
         Number of quiet flags (e.g., -q)
+    default : int, optional
+        Default log level, by default logging.WARNING
 
     Returns
     -------
@@ -207,5 +211,5 @@ def flags_to_level(verbose: int = 0, quiet: int = 0) -> int:
     """
     return max(
         logging.DEBUG,
-        min(logging.CRITICAL, logging.WARNING - 10 * verbose + 10 * quiet),
+        min(logging.CRITICAL, default - 10 * verbose + 10 * quiet),
     )


hydromt/log.py
```diff
@@ -11,28 +11,70 @@
...
 _LOG_FORMAT = "%(asctime)s - %(name)s - %(module)s - %(levelname)s - %(message)s"
 _DEFAULT_FORMATTER = logging.Formatter(_LOG_FORMAT)
 
 
-def initialize_logging() -> None:
-    """Initialize the hydromt root logger with a formatter, a log level and an optional console handler.
+def initialize_logging(
+    *,
+    file_path: Path | None = None,
+    console: bool = True,
```
dalmijn, line 20: Enable console by default (by that I mean removing the argument), I liked that about the old initialize_logging. This function should do all the default logging when no arguments are provided I would say.

hydromt/log.py
```diff
@@ -52,7 +99,7 @@ def set_log_level(log_level: int) -> None:
     _ROOT_LOGGER.setLevel(log_level)
 
 
-def _add_filehandler(
+def add_filehandler(
```
dalmijn: I thought this would still be private?

hydromt/log.py
```diff
@@ -11,28 +11,70 @@
...
-def initialize_logging() -> None:
-    """Initialize the hydromt root logger with a formatter, a log level and an optional console handler.
+def initialize_logging(
+    *,
+    file_path: Path | None = None,
+    console: bool = True,
+    level: int | None = None,
+    capture_warnings: bool = True,
+    formatter: logging.Formatter = _DEFAULT_FORMATTER,
+    logger: logging.Logger = _ROOT_LOGGER,
```
dalmijn: Do we want `logger` as an optional argument or only use the hydromt logger? I think there is something to be said for both.

hydromt/log.py
```diff
@@ -11,28 +11,70 @@
...
 
-def initialize_logging() -> None:
-    """Initialize the hydromt root logger with a formatter, a log level and an optional console handler.
+def initialize_logging(
+    *,
+    file_path: Path | None = None,
+    console: bool = True,
+    level: int | None = None,
+    capture_warnings: bool = True,
+    formatter: logging.Formatter = _DEFAULT_FORMATTER,
```
dalmijn: I feel like the `formatter` argument could be removed. This is the hydromt logger, the format can be fixed.

dalmijn (changes requested):
Nice changes! 
Not initialising at import seems like the right way to go.
This indeed is up to the user. 

I really like the new fixtures in the testing bench. Nicely done.
Did you verify that the logging still works as expected over cli?

I have a few comments regarding the code.
Also, could you add periods to the end of the parameter descriptions.

hydromt/cli/main.py
```diff
@@ -235,35 +235,28 @@ def build(
     hydromt build sfincs /path/to/model_root  -i /path/to/sfincs_config.yml
     -d /path/to/data_catalog.yml -v
     """  # noqa: E501
-    log_level = max(10, 30 - 10 * (verbose - quiet))
-    log.set_log_level(log_level=log_level)
-    log.log_version()
-    # Model.build will manage the filehandlers and logging
+    log.initialize_logging(level=max(10, 30 - 10 * (verbose - quiet)))
```
evetion: If we're refactoring, it seems good to remove this duplication everywhere, and fix the magic numbers as well.

Something like 

```py
log.initialize_logging(level=default_log_level(verbose, quiet))
```

And then ideally some enums instead of just an integer in the `default_log_level` method.

hydromt/log.py
```diff
@@ -137,8 +179,14 @@ def to_file(
     handler = _add_filehandler(
         path, log_level=log_level, formatter=formatter, logger=logger
     )
+    log_version(logger=logger)
```
evetion: nit: I get that this removes the duplication, but I don't feel that a contextmanager should log a version?

LuukBlom: Good point, I agree actually, just was removing duplication for the sake of it

PR #1406, opened 2026-03-26
feat: add azure support to datacatalog

LuukBlom:
## Issue addressed

Fixes #1405

## Explanation

Add support for Azure datastores. 
I vibe coded quite a bit, but have reviewed every line, and i think its a good start. 
I do want to test it with an actual azure real data store that requires authentication to validate it all works. 

## TODO

- [ ] add integration tests with a non public azure datastore (`azureml://`)


LuukBlom:
@evetion Could you approve this? I Spoke to the Henrique and the people UK project last friday, they can read from their own azureml urls as well.
They are still having problems, but those are not related to this.

evetion (approved):

PR #1417, opened 2026-04-09
Add SlippyTile raster driver and comprehensive test suite

maartenvanormondt:
## Summary
- Add `SlippyTileDriver` for reading XYZ/slippy map tiles stored in `{zoom}/{x}/{y}.png` directory structure
- Returns `xr.Dataset` in EPSG:3857 with terrain-decoded elevation data
- Supports multiple PNG encoders (terrarium, uint8/16/24/32, float8/16/24/32)
- Optional S3 download of missing tiles (unsigned/public access)
- Add 32 tests covering tile math, PNG decoding, driver read/write, zoom resolution, and options

## Test plan
- [x] 32 pytest tests pass (`tests/data_catalog/drivers/raster/test_slippy_tile_driver.py`)
- [ ] Verify CI passes on all platforms

Delft Dashboard finally uses hydromt-sfincs and hydromt-hurrywave. For that to work properly, it needs this slippy tile driver.


hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
```diff
@@ -0,0 +1,595 @@
...
+
+        if tiles_read == 0:
+            if handle_nodata == NoDataStrategy.RAISE:
+                raise IOError(
+                    f"No tiles found in {tile_root} at zoom {izoom} "
+                    f"for bbox {bbox_3857}"
+                )
+            logger.warning(
+                f"No tiles found in {tile_root} at zoom {izoom} for bbox {bbox_3857}"
+            )
```
LuukBlom: Can become 
```python
exec_nodata_strat(handle_nodata, msg)
return None
```

.github/workflows/sonar.yml
LuukBlom, line 1: Please revert changes here

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
```diff
@@ -0,0 +1,595 @@
...
+    dxy = _ZOOM0_PIXEL_SIZE / 2 ** np.arange(24)
+    izoom = np.nonzero(dxy < dx)[0]
+    if len(izoom) == 0:
+        return 23
+    return int(izoom[0])
+
+
+def _webmercator_to_latlon(easting: float, northing: float) -> Tuple[float, float]:
+    """Convert Web Mercator (EPSG:3857) to latitude/longitude (degrees)."""
+    lon = (easting / 20037508.34) * 180.0
```
LuukBlom: Please also make a constant for `20037508.34` to make it more readable and understandable (similar to `_ZOOM0_PIXEL_SIZE`)

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
```diff
@@ -0,0 +1,595 @@
...
+    lat = (180.0 / math.pi) * (
+        2.0 * math.atan(math.exp(northing / 20037508.34 * math.pi)) - (math.pi / 2.0)
+    )
+    return lat, lon
+
+
+def _latlon_to_webmercator(lat: float, lon: float) -> Tuple[float, float]:
+    """Convert latitude/longitude (degrees) to Web Mercator (EPSG:3857)."""
+    x = lon * 20037508.34 / 180.0
+    y = (math.log(math.tan((90.0 + lat) * math.pi / 360.0)) / math.pi) * 20037508.34
```
LuukBlom: https://github.com/Deltares/hydromt/pull/1417/changes#r3078047089

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
```diff
@@ -0,0 +1,595 @@
...
+
+
+def _download_tile(s3_client: Any, bucket: str, key: str, filename: str) -> bool:
+    """Download a single tile from S3.  Returns True on success."""
+    try:
+        Path(filename).parent.mkdir(parents=True, exist_ok=True)
+        s3_client.download_file(Bucket=bucket, Key=key, Filename=filename)
+        return True
+    except Exception:
+        return False
```
LuukBlom: bare exception catching is risky as it can also catch IOErrors in Path.mkdir for example.
Please at least add some logging:

```python
except Exception as e:
    logger.debug(f"Failed to download {key}: {e}")
    return False
```

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
```diff
@@ -0,0 +1,595 @@
...
+    encoder_vmin, encoder_vmax : float, optional
+        Value range for float encoders.
+
+    Returns
+    -------
+    np.ndarray
+        2-D elevation array (npix x npix).  NoData pixels are ``np.nan``
+        for float encoders and terrarium, or ``-1`` for uint encoders.
+    """
+    img = Image.open(png_file)
```
LuukBlom: Please use a context manager here.

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
```diff
@@ -0,0 +1,595 @@
...
+            + rgb[:, :, 1] * 65536
+            + rgb[:, :, 2] * 256
+            + rgb[:, :, 3]
+        )
+        elevation = encoder_vmin + (encoder_vmax - encoder_vmin) * idx / 4294967294
+        elevation[idx == 0] = np.nan
+    else:
+        raise ValueError(f"Unknown encoder: {encoder!r}")
+
+    return elevation
```
LuukBlom, line 289: Please add some comments explaining the “magic numbers” like 16777216, 65536, 4294967295, etc. with their power-of-two expressions (e.g., `2**24`, `2**16`, `2**32 - 1`),

Or add a comment next to each explaining what the number represents (e.g., “# 2**24”)?

roeldegoede:
I actually once started with this on the hydromt-sfincs side, where it actually returned a lazy xarray by delaying the actual reading of the files, see https://github.com/Deltares/hydromt_sfincs/blob/50ddd2c6f4deaa1a7e9ddd018943582232f15108/hydromt_sfincs/data/tiled_web_map_driver.py

hydromt/data_catalog/drivers/__init__.py
```diff
@@ -70,4 +76,5 @@
     "PyogrioDriver",
     "RasterDatasetXarrayDriver",
     "RasterioDriver",
+    "SlippyTileDriver",
```
LuukBlom, line 79: remove duplicate

tests/data_catalog/drivers/raster/test_slippy_tile_driver.py
```diff
@@ -0,0 +1,523 @@
...
+    @patch("hydromt.data_catalog.drivers.raster.slippy_tile_driver._HAS_BOTO3", False)
+    def test_download_missing_tiles_no_boto3(self, tmp_path):
+        """Without boto3, download should return 0."""
+        result = _download_missing_tiles(
+            str(tmp_path), "bucket", "key", "us-east-1", [(1, 0, 0)]
+        )
+        assert result == 0
+
+    @pytest.mark.skipif(not _HAS_BOTO3, reason="boto3 not installed")
+    @patch("hydromt.data_catalog.drivers.raster.slippy_tile_driver._HAS_BOTO3", True)
```
LuukBlom: this `patch` is not compatible with the `skipif`

tests/data_catalog/drivers/raster/test_slippy_tile_driver.py
```diff
@@ -0,0 +1,523 @@
...
+        mock_client = MagicMock()
+        mock_boto3.client.return_value = mock_client
+        result = _download_missing_tiles(
+            str(tmp_path), "bucket", "tiles", "us-east-1", [(1, 0, 0)]
+        )
+        assert result == 1
+        mock_client.download_file.assert_called_once()
+
+    @pytest.mark.skipif(not _HAS_BOTO3, reason="boto3 not installed")
+    @patch("hydromt.data_catalog.drivers.raster.slippy_tile_driver._HAS_BOTO3", True)
```
LuukBlom: same comment

LuukBlom (commented):
Nice implementation and tests! I see you are getting the hang of using Claude!

I dont have many significant comments, but a few small ones that make this PR align better with how hydromt is setup.

Second, before we merge this, I'd like to see an integration test using an s3 bucket and some test tiles in a datacatalog (as described in the module docstring). Currently, we only test under mock, which is good, but we should also add a test where we actually run things.

Also, we are making a release of hydromt core relatively soon. (https://github.com/Deltares/hydromt/milestone/15)
Is there a timeline/deadline for when you'd need this released @maartenvanormondt?

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
maartenvanormondt: it now reads:
   
    if tiles_read == 0:
        exec_nodata_strat(
                f"No tiles found in {tile_root} at zoom {izoom} for bbox {bbox_3857}",
                handle_nodata,
        )
        return None

note: my ai buddy says you got the order (message, handle_nodata) wrong...

.github/workflows/sonar.yml
maartenvanormondt: Sorry, that was claude being too enthousiastic. Reverted.

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
maartenvanormondt: done

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
maartenvanormondt: replaced 20037508.34 with _WEBMERCATOR_HALF_EXTENT

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
maartenvanormondt: done

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
maartenvanormondt: done

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
maartenvanormondt: done

hydromt/data_catalog/drivers/__init__.py
maartenvanormondt: from which list?

tests/data_catalog/drivers/raster/test_slippy_tile_driver.py
maartenvanormondt: removed @pytest.mark.skipif(not _HAS_BOTO3, reason="boto3 not installed")

tests/data_catalog/drivers/raster/test_slippy_tile_driver.py
maartenvanormondt: removed @pytest.mark.skipif(not _HAS_BOTO3, reason="boto3 not installed")

maartenvanormondt:
Hi Luuk,
Thanks for the thorough review. I think I covered everything. I added a test for cached tiles (North Sea, these are stored in hydromt\tests\data\gebco_2024\) and one test to actually fetch the tiles from S3 (somewhere near Florida). Is that what you meant? Not sure if boto3 is available in the test environment.
I actually need these changes ASAP. We have a client that needs the new Delft Dashboard (this week...). 

Thanks!

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
LuukBlom: Ah yes, good catch

hydromt/data_catalog/drivers/__init__.py
LuukBlom: oh nevermind, the github visualizations made it look like the additions were part of the same list.

LuukBlom (approved):
LGTM! I did some small updates to align the optional boto3 more with the rest of the codebase.

Perhaps @DirkEilander or @evetion want to do a review / quick pass as well before merging?

evetion (commented):
Thanks for making a PR @maartenvanormondt. It's good to have a slippy tile driver here, especially if it's used in projects. Even better, we're getting elevation values, my favorite form of data. That said, I feel this PR does mix some things that we should (eventually) separate. That can be a follow up, including functionality suggested by @roeldegoede, while this can have a dev release if needed.

We should add some more context to either the code or documentation, given the encoding logic and web mercator code that are otherwise hard to grasp: Tile servers are widespread on the web, and simple `depth/i/j` schemes are common. These are normally used for visualisation (so pictures of the data, not the raw data values), which is why there are encodings (like the terrarium one here) to store raw height values into the RGB channels of the png format.

This PR introduces a slippy tile client for local or S3 access, and always decodes to floating point values. It also hardcodes a web mercator projection. I assume these choices come directly from a project requirement, but it also limits its use. It's not a generic slippy tile client, we cannot use this (yet) to get terrain over http(s), use mapbox terrains, or even get the actual png/rgb data, nor are other schemes (WMTS comes to mind) supported. It's a S3TerrainSlippyTileDriver 😉.

I would suggest rewriting both the encoding en the web mercator code to use external libraries. We can't be the first to write a web mercator indexing (and we already depend on pyproj), and I expect there's code existing for the terrarium coding as well.

Finally, a small nit, I love *comprehensive* test suites, but I feel that Claude can go a bit overboard with making too many tests that become unreadable after a while. The test suite is 10 lines shorter than the code itself.

pyproject.toml
```diff
@@ -27,7 +27,7 @@ dependencies = [
 	"pyarrow",                # Apache Arrow bindings
 	"pydantic~=2.4",          # Validation
 	"pydantic-settings~=2.2", # Settings Management
-	"pyflwdir>=0.5.10",        # Hight models and derivatives
+	"pyflwdir>=0.5.10",       # Hight models and derivatives
```
evetion: If we're changing these lines, let's make it `Height`

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
```diff
@@ -338,32 +338,36 @@ def _download_missing_tiles(
...
     logger.info(
         f"Downloading {len(to_download)} missing tiles from s3://{s3_bucket}/{s3_key}/ ..."
     )
-    s3_client = boto3.client(
-        "s3", region_name=s3_region, config=BotoConfig(signature_version=UNSIGNED)
+    fs = s3fs.S3FileSystem(
+        anon=True,
+        client_kwargs={
+            "region_name": s3_region,
+            "endpoint_url": f"https://s3.{s3_region}.amazonaws.com",
```
evetion: Is this required?

evetion (approved):
Looks good, @maartenvanormondt can you check if this still works with you with the latest commits (we changed from `boto3` to `aiobotocore` via s3fs)? Especially if you have use cases with a custom S3 configuration.

hydromt/data_catalog/drivers/raster/slippy_tile_driver.py
LuukBlom: Well, probably this is not required, but I could not get it to work quickly with virtual hosted URLS (`https://deltares-ddb.s3.eu-west-1.amazonaws.com/key`) as i was getting strange `PermissionError: Access Denied` to Maartens public bucket.
Path style did work though: (`https://s3.eu-west-1.amazonaws.com/deltares-ddb/key`).

They point to the same bucket, just with a different URL format, according to aws docs: https://docs.aws.amazon.com/AmazonS3/latest/userguide/VirtualHosting.html#path-style-access

That docs page also references this blog from 2020 that says path style will be deprecated
https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

But more googling doesn't result in any other info about this supposed deprecation from 6 years ago...

LuukBlom: I have now fixed it.

`s3fs` does not automatically fall back to anonymous access when no AWS credentials are available. 
By default it always attempts to sign requests, and if the credential chain resolves nothing, this results in a `NoCredentialsError` rather than an unsigned request.

fix: I added `has_credentials` to `slippy_tile_driver.py` that uses `botocore` to see if any AWS credentials are available.
- If credentials are available, requests are signed and private buckets work as expected
- If no credentials are found, `anon=True` is used to enable unsigned requests, which is required for accessing public buckets.

This avoids the error where public S3 data could not be accessed without configured AWS credentials. (this was the reason I was setting `endpoint_url`)

This means: 
- Users can configure the S3Filesystem with environment variables (as defined in the aws docs)
- `s3_region` is no longer needed in the function `_download_missing_tiles`, since region and endpoint resolution are handled either by AWS defaults or the configured environment variables.

PR #1429, opened 2026-04-13
add workflow to create prerelease

### LINKED ISSUE #1428: Test making pre-release
We should be able to create a pre-release (either -devX or -rcX), and make sure everything still works (documentation, pypi, Github, etc), but marked as pre-release where possible.

LuukBlom:
## Issue addressed

Fixes #1428

## Explanation

Add a workflow that creates a pre-release from any branch and publishes it to Pypi and GitHub.

Added a push trigger temporarily to run the workflow and created this pre-release: https://github.com/Deltares/hydromt/releases/tag/v1.4.0.dev20260413131052
Using this succesfull workflow: https://github.com/Deltares/hydromt/actions/runs/24345179576/job/71083973853

Removed the push trigger, so we can merge this now.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -2,10 +2,6 @@ name: Pre-release
 description: Create a pre-release from any branch, publish it to pypi and GitHub to allow for easier testing.
 
 on:
-  # TODO: remove push trigger before merging to main
-  push:
-    branches:
-      - feat/add-prerelease-workflow
   workflow_dispatch:
     inputs:
       description:
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -8,6 +8,24 @@ on:
         description: "Description for this pre-release"
         required: true
         type: string
+      base_version:
+        description: "Base version (e.g. 1.2.0). Defaults to current version in source."
+        required: false
+        type: string
+      identifier:
+        description: "Pre-release identifier. One of: rc, alpha, beta, or timestamp."
+        required: false
+        type: choice
+        options:
+          - timestamp
+          - rc
+          - alpha
+          - beta
+        default: timestamp
+      build_nr:
+        description: "Build number (e.g. 1, 2). Defaults to lowest non-existing tag."
+        required: false
+        type: string
 
 permissions:
   contents: write
@@ -22,7 +40,7 @@ jobs:
 
       - uses: actions/setup-python@v6
         with:
-          python-version: "3.14"
+          python-version: "3.12"
 
       - name: Install build tools
         run: pip install flit twine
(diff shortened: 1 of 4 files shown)


evetion:
Nice work, and quick too! Some thoughts:
- Could we integrate this into the existing workflow? Now we duplicate most of the release workflow, only to change one or two things.
- Is the documentation updated with this release process? (i.e. new version, dropdown works).

LuukBlom:
> * Could we integrate this into the existing workflow? Now we duplicate most of the release workflow, only to change one or two things.

The release process is semi automated and spread out over 3 files: `prep-release.yml`, `create-release-artifacts.yml` and `finish-release.yml`. 

These do quite a bit of stuff that is not relevant to prereleases:
- create PR + release artifacts
- create and publish docs
- branch triggers (only for main / release branches)

To accomodate for prereleases in the existing workflows would introduce more complexity into the workflows, which im not sure is a great direction.

The only things that both pre-release workflows and release workflows do is extract the version and use some actions (upload/download/publish etc).

I actually like having this pre-release workflow fully isolated and separate from true releases. 
It makes it simple to understand and use imo: Actions tab -> workflow dispatch: pre-release -> choose your branch and bam, the prerelease is made.

LuukBlom:
> * Is the documentation updated with this release process? (i.e. new version, dropdown works).

Well, pre-releases are not a part of the documentation, so I don't think we should add it the the `switcher.json` file.
We will use pre-releases only for (internal) testing.

But good point, I will add a section in the docs describing this workflow and how to use it.

evetion:
> Well, pre-releases are not a part of the documentation, so I don't think we should add it the the `switcher.json` file. We will use pre-releases only for (internal) testing.

Fair enough, but wouldn't we want these changes as part of the [unreleased section](https://deltares.github.io/hydromt/stable/changelog.html#unreleased) of the documentation?

My other comment, but does not need to hold this back, is whether we can also make release candidates (rc) using this (or the other normal release) workflow. If not, that might be a good edit/follow up.

evetion (approved):

LuukBlom:
> Fair enough, but wouldn't we want these changes as part of the [unreleased section](https://deltares.github.io/hydromt/stable/changelog.html#unreleased) of the documentation?

Im not sure about how we can achieve this. 
The unreleased section is for changes that are in main, but not in a release yet.
So, if the prerelease is made from main, that is already the case and that is already online.
If the pre-release is not made from main, the changes will not be on the website yet, but it wouldnt make sense to change the docs online for a release that is not from main. That would be the docs people would see if they installed the latest normal release. 
 
I think we have to choose one of:
- pre-releases are strictly from main (this would make sure that the docs and code always align.
- we accept that docs for pre-releases are not represented on the public docs.

> My other comment, but does not need to hold this back, is whether we can also make release candidates (rc) using this (or the other normal release) workflow. If not, that might be a good edit/follow up.

I like that, I will add an option to allow to provide the full name to use instead of the auto generated timestamp.

.github/workflows/pre-release.yml
```diff
@@ -8,6 +8,24 @@ on:
...
+        type: string
+      identifier:
+        description: "Pre-release identifier. One of: rc, alpha, beta, or timestamp."
+        required: false
+        type: choice
+        options:
+          - timestamp
+          - rc
+          - alpha
+          - beta
```
evetion: I like your enthousiasm here! However, I think @deltamarnix his memo prescribed only dev (which is removed now?) and rc, and explicitly forbade alpha and beta (because they are confusing).

LuukBlom: Ah okay, that makes sense. I'll update

PR #1430, opened 2026-04-14
feat: Update ConfigComponent

### LINKED ISSUE #1418: Improve ConfigComponent test
### Kind of request

### Enhancement Description

The current ``ConfigComponent`` ``test_equal`` method only returns "config mismatch" is the configs are not equal which does not tell much (eg which of my 1000 lines actually mismatch?)

I think nice if it can return more info than that.

### Use case

Easiest to debug model compare errors

### Additional Context

This was implemented in hydromt_delft3dfm based on hydromt core v0.10

``` python
    def test_equal(self, other: ConfigComponent) -> tuple[bool, dict[str, str]]:
        """Test if two components are equal.

        Parameters
        ----------
        other : ModelComponent
            The component to compare against.

        Returns
        -------
        tuple[bool, Dict[str, str]]
            True if the components are equal, and a dict with the associated errors per
            property checked.
        """
        errors = {}
        if not isinstance(other, self.__class__):
            errors["__class__"] = f"other does not inherit from {self.__class__}."
        eq = len(errors) == 0
        if not eq:
            return eq, {"mdu": errors}
        other_config = cast(ConfigComponent, other)


        # not enough details in python recursion
        errors.update(**_check_equal(self.data, other_config.data, "mdu"))
        return len(errors) == 0, {"mdu": errors}




def _check_equal(a, b, name="") -> dict[str, str]:
    """Recursive test of model components.

    Returns dict with component name and associated error message.
    """
    errors = {}
    try:
        assert isinstance(b, type(a)), "property types do not match"
        if isinstance(a, dict):
            for key in a:
                assert key in b, f"{key} missing"
                errors.update(**_check_equal(a[key], b[key], f"{name}.{key}"))
        else:
            assert a == b, "values not equal"
    except AssertionError as e:
        errors.update({name: e})
    return errors
```

LuukBlom:
## Issue addressed

Fixes #1418

## Explanation

added the function from the issue, updated it and added some tests

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,7 +15,7 @@ New
 ---
 - Added earthdatahub_data predefined catalog for accessing ERA5_land, ERA5_land_hourly, ERA5_ocean (#1357).
 - Added ``precision`` parameter to ``GeomsComponent.write`` to set the number of decimals used when writing (#1415).
-- More informative error messages when calling ``test_equal`` on ConfigComponent (#1430). 
+- More informative error messages when calling ``test_equal`` on ConfigComponent (#1430).
 
 Changed
 -------
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,17 +14,18 @@ Unreleased
 New
 ---
 - Added earthdatahub_data predefined catalog for accessing ERA5_land, ERA5_land_hourly, ERA5_ocean (#1357).
-- Added ``precision`` parameter to ``GeomsComponent.write`` to set the number of decimals used when writing (#1415)
+- Added ``precision`` parameter to ``GeomsComponent.write`` to set the number of decimals used when writing (#1415).
+- More informative error messages when calling ``test_equal`` on ConfigComponent (#1430). 
 
 Changed
 -------
-- ``log.initialize_logging()`` is no longer called when hydromt is imported. It is now the users responsibility to configure logging at the start of their scripts. (#1399)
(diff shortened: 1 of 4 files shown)


hboisgon (approved):
Nice! You even made it better :)

One remark, in some of the plugins, I return errors = {component_name: errors} because if you have 20 errors on config and 10 on geoms, it becomes hard to read which components are failing and which are passing.

Maybe this is something we can better handle at the Model.test_equal level?

https://github.com/Deltares/hydromt/blob/9d42ac1c58a158e4933f8bfc2dbffeeff7ce5013/hydromt/model/model.py#L558-L564

I approve already but think it would be nice to add the above to Model.test_equal.
And fix the linting of course :)

LuukBlom:
> Nice! You even made it better :)
> 
> One remark, in some of the plugins, I return errors = {component_name: errors} because if you have 20 errors on config and 10 on geoms, it becomes hard to read which components are failing and which are passing.
> 
> Maybe this is something we can better handle at the Model.test_equal level?
> 
> https://github.com/Deltares/hydromt/blob/9d42ac1c58a158e4933f8bfc2dbffeeff7ce5013/hydromt/model/model.py#L558-L564
> 
> I approve already but think it would be nice to add the above to Model.test_equal. And fix the linting of course :)

Good idea! Will do.
I agree, all `component.test_equal` methods should just return a dict with errors, the model class can then call these in a loop like shown and store the returned dicts like `errors[name] = component_errors`

PR #1435, opened 2026-04-21
Add linux paths to deltares catalog.

evetion:
This revives the effort made in #786 to introduce multiple catalog roots, I've made v1.1.1 to do so. Note that the version number in the deltares data catalog v1.1.0 has version v1.0.0 (!), that seems an oversight, but I'm not sure whether we can fix it anymore (although I try to do so here).

edit: The logic for parsing `roots` is different than `root`. When `root` is present, it is popped unvalidated from the meta dict (possibly yielding an error later), while all `roots` get checked. I've had to change that behaviour to allow non-existent paths (p is not available on github).


hydromt/data_catalog/predefined_catalog.py
LuukBlom, line 1: The version and thereby the hash being wrong might not be the biggest problem:
We use `pooch.Pooch.fetch` to download the catalog files.

[API docs](https://www.fatiando.org/pooch/latest/api/generated/pooch.Pooch.html#pooch.Pooch.fetch) for `pooch.Pooch.fetch` say this: 

> If it’s not in the local storage, it will be downloaded. If the hash of the file in local storage doesn’t match the one in the registry, will download a new copy of the file. This is considered a sign that the file was updated in the remote storage. If the hash of the downloaded file still doesn’t match the one in the registry, will raise an exception to warn of possible file corruption.

So, for users I think it should see the hash is different and then re-download.

evetion: Nice, than this retconning should work :)

LuukBlom (approved):

PR #1439, opened 2026-04-23
make pet_debruin able to lazily do operations.

### LINKED ISSUE #1420: `pet_debruin` does not support lazy operations
### HydroMT version checks

### Reproducible Example

When building larger models, the method `pet_debruin` tries to load too much data in memory, causing memory errors (even when building forcing with the `output_frequency` option from hydromt_wflow)

### Current behaviour

Looking at the function, this line tries to put the data in memory as opposed to keep all operations lazy: https://github.com/Deltares/hydromt/blob/5d6a4efabcea67627c2d8dafc55b20ac37791403/hydromt/model/processes/meteo.py#L585

### Desired behaviour

Keep all operations lazy, such that it is easier to build large models without running into memory issues. Perhaps something like this works better, not sure if this is the best solution: 
```python
ep_joule = xr.where(
    xr.apply_ufunc(np.isclose, k_ext, 0.0, dask="allowed"), 0.0, ep_joule
)
```

### Additional context

_No response_

LuukBlom:
## Issue addressed

Fixes #1420 

## Explanation

Update pet_debruin to not use any numpy functions since they are eager and try to load all data. 
New implementation replaces numpy with xarray-native operations.

A boolean mask (abs(k_ext) > 1e-8) is defined once and reused to both avoid division by zero and mask invalid regions. 
All operations (where, arithmetic, clip) are Dask-aware and remain lazy

### DIFF
--- a/hydromt/model/processes/meteo.py
+++ b/hydromt/model/processes/meteo.py
@@ -568,22 +568,30 @@ def pet_debruin(
     """
     # saturation and actual vapour pressure at given temperature [hPa °C-1]
     esat = 6.112 * np.exp((17.67 * temp) / (temp + 243.5))
+
     # slope of vapour pressure curve [hPa °C-1]
     slope = esat * (17.269 / (temp + 243.5)) * (1.0 - (temp / (temp + 243.5)))
-    # compute latent heat of vapourization [J kg-1]
-    lam = (2.502 * 10**6) - (2250.0 * temp)
+
+    # latent heat of vapourization [J kg-1]
+    lam = (2.502e6) - (2250.0 * temp)
+
     # psychometric constant [hPa °C-1]
     gamma = (cp * press) / (0.622 * lam)
-    # compute ref. evaporation (with global radiation, therefore calling it potential)
-    # in J m-2 over whole period
-    ep_joule = (
-        (slope / (slope + gamma))
-        * (((1.0 - 0.23) * k_in) - (cs * (k_in / (k_ext + 0.00001))))
-    ) + beta
-    ep_joule = xr.where(np.isclose(k_ext, 0.0), 0.0, ep_joule)
-    pet = ((ep_joule / lam) * timestep).astype(np.float32)
-    pet = xr.where(pet > 0.0, pet, 0.0)
-    return pet
+
+    # define mask in a lazy way to avoid division by zero
+    mask = abs(k_ext) > 1e-8
+
+    # compute ref. evaporation [J m-2]
+    ratio = xr.where(mask, k_in / k_ext, 0.0)
+    ep_joule = (slope / (slope + gamma)) * (((1.0 - 0.23) * k_in) - cs * ratio) + beta
+
+    # apply same mask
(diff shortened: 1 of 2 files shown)


hydromt/model/processes/meteo.py
```diff
@@ -568,22 +568,30 @@ def pet_debruin(
...
-    ep_joule = (
-        (slope / (slope + gamma))
-        * (((1.0 - 0.23) * k_in) - (cs * (k_in / (k_ext + 0.00001))))
-    ) + beta
-    ep_joule = xr.where(np.isclose(k_ext, 0.0), 0.0, ep_joule)
-    pet = ((ep_joule / lam) * timestep).astype(np.float32)
-    pet = xr.where(pet > 0.0, pet, 0.0)
-    return pet
+
+    # define mask in a lazy way to avoid division by zero
```
evetion, line 581: This removed the `()` part of the comment, which seems useful to have.

hydromt/model/processes/meteo.py
```diff
@@ -568,22 +568,30 @@ def pet_debruin(
...
+    mask = abs(k_ext) > 1e-8
```
evetion, line 582: This is not exactly the same as [`np.isclose`](https://numpy.org/doc/2.1/reference/generated/numpy.isclose.html) no? That said, it's probably small enough not to matter much, is this function tested somewhere?

evetion (approved):

hydromt/model/processes/meteo.py
LuukBlom: good point, I was a bit eager pressing tab in my IDE and didnt notice this. autocomplete more like autodelete.

hydromt/model/processes/meteo.py
LuukBlom: NumPy functions are eager, they load all data into memory, which is exactly the problem described in the issue. So I wanted to avoid all numpy functions here.

Calling `abs` on a lazy loaded dask array calls dask internals under the hood and keeps it all lazy. You can verify this:

```python
import numpy as np
import xarray as xr
import dask.array as da

data = da.from_array(np.array([0.0, 1e-9, 1e-4, 1.0]), chunks=2)
k_ext = xr.DataArray(data)

# stays lazy, still a dask.array.Array
result = abs(k_ext) > 1e-8
print(type(result.data))  # <class 'dask.array.core.Array'>

# forces compute, returns a numpy.ndarray immediately
result = np.isclose(k_ext, 0.0)
print(type(result))  # <class 'numpy.ndarray'>
```

LuukBlom: Also, there is one test: `tests.model.processes.meteo.test_pet_debruin`:

```python
def test_pet_debruin(era5_data):
    pet = pet_debruin(
        era5_data["temp"], era5_data["press_msl"], era5_data["kin"], era5_data["kout"]
    )
    assert isinstance(pet, xr.DataArray)
    np.testing.assert_almost_equal(pet.mean(), 0.8540, decimal=4)
```

so the output is unchanged.

PR #1447, opened 2026-05-01
fix: ``convert to string before calling 'lower()' to prevent TypeError

LuukBlom:
## Explanation

Ran into this issue when using the latest rc2 for hydromt sfincs.
Calling `.lower()` on anything that is not a string will always error.
also pandas docs says: `pd.DataFrame.columns : list[str]`, whereas `pd.DataFrame.index` can be many types, columns is only the labels and should always be strings.


deltamarnix (changes requested):
Can you add a unit test that shows where this goes wrong?

LuukBlom:
Im not sure if this is something we should test, as this is more pandas imo.
This happens when the columns are not a string, which hydromt sfincs does quite a bit. 
They have a lot of headerless input files and use integers for columns (for example https://github.com/Deltares/hydromt_sfincs/blob/main/tests/data/sfincs_test/sfincs.obs)

I have added a test now though, even though I think its a slightly strange test since we cant test the previous version showing it actually goes wrong after having updated it. 

To show it anyways, I copied the entire function added `_safe` to the name and did the string conversion there, and this test passes:
```python
def test_open_vector_from_table_non_string_columns(tmp_path: Path):
    df = pd.DataFrame({0: [1, 2, 3], 1: [4, 5, 6], "X": [7, 8, 9]})
    path = tmp_path / "test.csv"
    df.to_csv(path, sep=r" ", header=False, index=False)
# Old approach fails on integer column names
    with pytest.raises(AttributeError):
        _gdf = open_vector_from_table(
            path,
            driver="csv",
            header=None,
            index_col=False,
            sep=r"\s+",
            x_dim="0",
            y_dim="1",
            crs=32633,
        )

    # Fixed approach works: open_vector_from_table uses str(c).lower()
    gdf = open_vector_from_table_safe(
        path,
        driver="csv",
        header=None,
        index_col=False,
        sep=r"\s+",
        x_dim="0",
        y_dim="1",
        crs=32633,
    )
    assert isinstance(gdf, gpd.GeoDataFrame)
    assert len(gdf) == 3
```

deltamarnix (approved):

PR #1460, opened 2026-05-12
feat: inclusive time slicing

### LINKED ISSUE #1458: Time slicing in get_rasterdataset() should be inclusive
### HydroMT version checks

### Reproducible Example

When providing a time_range to `data_catalog.get_rasterdataset()`, a simple isel() is used to do the slicing. If the user provides a start/stop time that does not exactly match the timesteps of the dataset, the resulting slice is too narrow:
```python
from hydromt.data_catalog import DataCatalog
from hydromt.model.processes.region import parse_region_bbox

data_catalog = DataCatalog(data_libs=["artifact_data"])
region = dict(bbox=[12.4331, 46.4661, 12.5212, 46.5369])
region_crs = 4326
geom = parse_region_bbox(region, crs=region_crs)
tstart = "2010-02-01 12:00:00"
tstop = "2010-02-03 12:00:00"
meteo_data = data_catalog.get_rasterdataset(
            data_like="era5",
            geom=geom,
            buffer=2,
            time_range=(tstart, tstop),
            variables=["temp"],
        )

print("time min", meteo_data.time.to_pandas().min())
print("time max", meteo_data.time.to_pandas().max())
```

### Current behaviour

In the code we request "2010-02-01 12:00:00" to "2010-02-03 12:00:00", but we get "2010-02-02 00:00:00" to "2010-02-03 00:00:00". This is 12 hours too short on both sides.

### Desired behaviour

Inclusive time slicing of the data, in this example it should be "2010-02-01 00:00:00" to "2010-02-04 00:00:00". This can be done like it was done in dfm_tools:
https://github.com/Deltares/dfm_tools/blob/f07ad0a02dd2920776a971893b1382d2c05ade03/dfm_tools/interpolate_grid2bnd.py#L403-L415

### Additional context

Might be related to https://github.com/Deltares/hydromt/issues/665.

LuukBlom:
add `_create_time_slice` that can create inclusive or exclusive timeslices and use that in all `_slice_data` functions that perform slicing.

## Issue addressed

Fixes #1458

## Explanation

This PR makes `inclusive` the default behaviour for time slicing, but allows for users to set `inclusive=True/False` when calling a get_data method on the datacatalog.

1. Is this something we want to keep configurable by users?
2. Should we just remove the option in-/exclusive and always do inclusive time slicing?

Making this the new behaviour without users being able to configure will likely break some tests in the plugins since data will not be the exact same anymore. 

But adding this `inclusive` argument everywhere is also not ideal. 
What do you think @deltamarnix @evetion ?

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist


hydromt/data_catalog/data_catalog.py
```diff
@@ -1835,6 +1850,7 @@ def get_dataframe(
         data_like: Union[str, SourceSpecDict, Path, pd.DataFrame, DataFrameSource],
         variables: Optional[List] = None,
         time_range: TimeRange | tuple | dict | None = None,
+        inclusive: bool = True,
```
evetion: This is a breaking change (depending on our contract), might be good to set it to false first, and ask at our next demo before we make it the default?

hydromt/data_catalog/adapters/adapter_utils.py
```diff
@@ -17,18 +18,129 @@
 logger = logging.getLogger(__name__)
 
 
+def _create_time_slice(
+    data: xr.Dataset | pd.DataFrame,
```
evetion: I miss multiple dispatch

tests/data_catalog/adapters/test_adapter_utils.py
evetion, line 1: Are we using AI for this? It seems overkill, would be good to get it much smaller. Maybe you can parametrize the possible combinations with an expected length and/or first/last value?

hydromt/data_catalog/adapters/adapter_utils.py
evetion: We should probably split this function, maybe move it back to each individual object that it's operating on (not sure if we need to make it into an abstract method)? There is a bit of overlap in the errors, but it reads much more cleanly when split.

evetion (changes requested):
This looks fine to me, good to have it. But maybe not make it the default right now? 

Did you consider adding include and exclusive in the same request for either start or end? Like `[1,2)`. @veenstrajelmer do you have cases like that?

hydromt/data_catalog/data_catalog.py
veenstrajelmer: Indeed, although I would not expect a model to break break from including potentially more times in the slice, except for tests maybe. But then again, I cannot oversee that, so it might be wise to be conservative indeed.

veenstrajelmer:
> Did you consider adding include and exclusive in the same request for either start or end? Like `[1,2)`. @veenstrajelmer do you have cases like that?

I don't have cases like that and since this inclusive requests only pops up now, I think it is valid to assume there are also no such cases in other hydromt plugins. The only case I have is selecting such that the selection is exactly from/to the sliced times or buffered to the next outer times if there are no exact matches. So as far as I can judge, this implementation looks perfect.

I do have one question, I see indexing with 0/-1, is it guaranteed that the times are sorted at this point in the code?

hydromt/data_catalog/adapters/adapter_utils.py
```diff
@@ -17,18 +18,203 @@
...
+    return slice(time_index[start_idx], time_index[stop_idx])
+
+
+def _create_time_slice(
+    data: xr.Dataset | pd.DataFrame,
+    tstart: pd.Timestamp | datetime | str,
+    tstop: pd.Timestamp | datetime | str,
+    inclusive: bool = False,
+    handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
+) -> slice | None:
```
deltamarnix, line 171:
```suggestion
def _create_time_slice(
    data: xr.Dataset | pd.DataFrame,
    *,
    tstart: pd.Timestamp | datetime | str,
    tstop: pd.Timestamp | datetime | str,
    inclusive: bool = False,
    handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
) -> slice | None:
```

I recommend using more named parameters, since there are so many and also defaults. Also for the other private functions.

hydromt/data_catalog/adapters/adapter_utils.py
```diff
@@ -17,18 +18,203 @@
...
-        ds = ds.sel(time=slice(time_range.start, time_range.end))
+        time_slice = _create_time_slice(
+            ds,
+            time_range.start,
+            time_range.end,
+            inclusive=inclusive,
+            handle_nodata=handle_nodata,
+        )
+        if time_slice is None:
+            return None
```
deltamarnix, line 224: Is this in line with the NoDataStrategy or does that only count towards the actual data?

hydromt/data_catalog/adapters/adapter_utils.py
```diff
@@ -17,18 +18,203 @@
...
+    -------
+    slice | None
+        A slice object for the time dimension, or None if no overlap exists.
+    """
+    pd_tstart = pd.Timestamp(tstart)
+    pd_tstop = pd.Timestamp(tstop)
+
+    time_values = pd.to_datetime(ds["time"].values)
+    if not time_values.is_monotonic_increasing:
+        logger.warning("Dataset time dimension is not sorted; sorting for slicing.")
```
deltamarnix: Warnings in hydromt are actions for users to fix and that makes them reiterate on their model. Is this warnings such a case or is it only info?

hydromt/data_catalog/adapters/adapter_utils.py
```diff
@@ -17,18 +18,203 @@
...
+        ds = ds.sortby("time")
+        time_values = pd.to_datetime(ds["time"].values)
+
+    result = _check_time_overlap(
+        pd_tstart, pd_tstop, time_values[0], time_values[-1], handle_nodata
+    )
+    if result is None:
+        return None
```
deltamarnix, line 91: Is this in line with the no data strategy?

hydromt/data_catalog/adapters/adapter_utils.py
```diff
@@ -17,18 +18,203 @@
...
+        A slice object for the DataFrame index, or None if no overlap exists.
+    """
+    pd_tstart = pd.Timestamp(tstart)
+    pd_tstop = pd.Timestamp(tstop)
+
+    time_index = df.index
+    if not isinstance(time_index, pd.DatetimeIndex):
+        raise ValueError("DataFrame index must be a DatetimeIndex.")
+    if not time_index.is_monotonic_increasing:
+        logger.warning("DataFrame DatetimeIndex is not sorted; sorting for slicing.")
```
deltamarnix: Same for this warning. Does the user make changes, or is it info and can we continue?

hydromt/data_catalog/adapters/adapter_utils.py
```diff
@@ -17,18 +18,203 @@
...
+    if inclusive:
+        minimum = ds.sel(time=true_tstart, method="pad")
+        maximum = ds.sel(time=true_tstop, method="backfill")
+    else:
+        minimum = ds.sel(time=true_tstart, method="backfill")
+        maximum = ds.sel(time=true_tstop, method="pad")
+    return slice(minimum.time.values, maximum.time.values)
+
+
+def _create_dataframe_time_slice(
```
deltamarnix, line 103: These two functions look very much alike. Can we do a neat trick with function arguments, or does that only make things more complicated?

hydromt/data_catalog/data_catalog.py
```diff
@@ -1343,6 +1343,7 @@ def get_rasterdataset(
         handle_nodata: NoDataStrategy = NoDataStrategy.RAISE,
         variables: Optional[Union[List, str]] = None,
         time_range: TimeRange | tuple | dict | None = None,
+        inclusive: bool = False,
```
deltamarnix: I am wondering if this is going too far. Are there other ways to indicate that the TimeRange is inclusive or exclusive? The class is ours, we could add `inclusive` or `inclusive_start`, `inclusive_end` as optional parameters to the initializer and not have even more arguments in `get_rasterdataset`. I'm not sure if that makes it easier. But we could say: TimeRange can be inclusive, tuple and ?dict? are always exclusive. Or is TimeRange used throughout the code for other things that don't need this extra option in the class?

hydromt/data_catalog/data_catalog.py
```diff
@@ -1391,6 +1392,9 @@ def get_rasterdataset(
             Start and end date of period of interest. By default the entire time period
             of the dataset is returned. If not None, must be parsable by TimeRange.create,
             by default None
+        inclusive : bool, optional
+            Whether the start and end date of `time_range` are included in the returned data,
+            by default True
```
deltamarnix: default False

hydromt/data_catalog/data_catalog.py
```diff
@@ -1651,6 +1657,9 @@ def get_geodataset(
...
```
deltamarnix: default False

hydromt/data_catalog/data_catalog.py
```diff
@@ -1769,6 +1780,9 @@ def get_dataset(
...
```
deltamarnix: default False

hydromt/data_catalog/data_catalog.py
```diff
@@ -1857,6 +1873,9 @@ def get_dataframe(
...
```
deltamarnix: default False

deltamarnix (changes requested):
I have an idea to upgrade `TimeRange`. Also requested changes because the documentation is wrong.

hydromt/data_catalog/data_catalog.py
evetion: Yeah I dislike the extra argument, but didn't want to make this PR (which I adopted) even bigger. I'll check TimeRange, it seems an excellent place to do so.

hydromt/data_catalog/adapters/adapter_utils.py
evetion: I actually split them from a combined function was a bit too complex due to the constant if/else.

hydromt/data_catalog/adapters/adapter_utils.py
evetion: Yep, that is handled by `_check_time_overlap`.

evetion:
@veenstrajelmer I've adapter your issue test, but as there's no first day of the month in the era5 dataset, I've shifted your example by 1 day:

```python
data_catalog = DataCatalog(data_libs=["artifact_data"])
region = dict(bbox=[12.4331, 46.4661, 12.5212, 46.5369])
region_crs = 4326
geom = parse_region_bbox(region, crs=region_crs)
tstart = "2010-02-03 12:00:00"
tstop = "2010-02-04 12:00:00"
meteo_data = data_catalog.get_rasterdataset(
    data_like="era5",
    geom=geom,
    buffer=2,
    time_range={"start": tstart, "end": tstop, "inclusive": False},
    variables=["temp"],
)

assert meteo_data.time.to_pandas().min() == pd.Timestamp("2010-02-04 00:00:00")
assert meteo_data.time.to_pandas().max() == pd.Timestamp("2010-02-04 00:00:00")

meteo_data = data_catalog.get_rasterdataset(
    data_like="era5",
    geom=geom,
    buffer=1,
    time_range={"start": tstart, "end": tstop, "inclusive": True},
    variables=["temp"],
)

assert meteo_data.time.to_pandas().min() == pd.Timestamp("2010-02-03 00:00:00")
assert meteo_data.time.to_pandas().max() == pd.Timestamp("2010-02-05 00:00:00")
```

PR #1461, opened 2026-05-12
feat: test using lowest allowed deps

LuukBlom:
## Summary

Add CI testing with `solve-strategy: lowest-direct` from pixi to verify the lower bounds of our dependency requirements.

## Changes

- Add `[tool.pixi.feature.lowest-deps]` with `solve-strategy = "lowest-direct"` and `no-build = true`
- Add minimum version bounds to all dependencies (core + optional)
- Add `low-py3xx` pixi environments
- Add `low` to the CI test matrix
- Fix test isolation issue in test_plugins.py where a mock leaked across plugin groups
- Suppress warnings-as-errors in `low-pyXXX` envs (older deps emit deprecation warnings)

## Approach for choosing minimum versions

I followed [Scientific Python SPEC 0](https://scientific-python.org/specs/spec-0000/) where possible.

For deps without existing constraints, I picked a stable release from ~2024 as a starting point. This keeps the supported range reasonable without being overly broad and a headache to maintain backwards compatibility. 
Open to discuss alternatives.

If the `low` env didn't resolve: bumped the minimum version.
If it resolved but tests failed: investigated and either fixed the test or bumped the minimum if the issue was a true incompatibility.

## Why no upper bounds

It was intentional to not add any upper bounds. Some of our deps use calendar versioning (xarray, fsspec, dask), where each yearly release is effectively a new "major" version. Upper bounds on CalVer packages cause dependency conflicts for users without actually solving anything. Our regular CI with `highest` solve already catches that. 
Also open to discuss how we want to handle this better. (this was a problem for pandas 3.0 for example)

For semver packages, we could add upper bounds, but it would require actively tracking new releases. 
The `lowest-direct` + `highest` (default) CI combination covers both ends of this range without becoming a maintenance headache.


hydromt/data_catalog/uri_resolvers/uri_resolver.py
```diff
@@ -25,42 +24,42 @@ class URIResolver(AbstractBaseModel, ABC):
 
     model_config = ConfigDict(extra="forbid")
     filesystem: FSSpecFileSystem = Field(default_factory=FSSpecFileSystem)
-    options: dict[str, Any] = Field(default_factory=dict)
```
deltamarnix, line 28: Does mypy accept this? I remember that mypy wants `dict[type, type]`

tests/data_catalog/drivers/dataset/test_dataset_xarray_driver.py
```diff
@@ -47,7 +47,7 @@ def test_zarr_read(self, example_zarr_file: Path):
         res: xr.Dataset = DatasetXarrayDriver().read([str(example_zarr_file)])
         assert list(res.data_vars.keys()) == ["variable"]
         assert res["variable"].shape == (10, 10)
-        assert list(res.coords.keys()) == ["xc", "yc"]
+        assert all(key in res.coords for key in ["xc", "yc"])
```
deltamarnix:
```suggestion
        assert set(res.coords) == {"xc", "yc"}
```

pyproject.toml
```diff
@@ -366,8 +368,18 @@ full-py314 = { features = [
 	"doc",
 	"examples",
 ], solve-group = "py314" }
+low-py314 = { features = [
+	"lowest-deps",
+	"py314",
+	"io",
+	"vrt",
```
deltamarnix: I have an idea that I want to check with you. Would it be possible to set the dependency resolver of `vrt` to `lowest`, so that also the transient dependencies are solved to their lowest point? Or would that still give problems because it will try to install numpy 1.0?

pyproject.toml
```diff
@@ -237,12 +237,25 @@ hydromt = { path = ".", editable = true }
...
 [tool.pixi.feature.vrt.dependencies]
 gdal = ">=3.5.0"
 
+[tool.pixi.feature.lowest-deps]
+solve-strategy = "lowest-direct"
+
+# Conda-side constraints to match the lower bounds declared in [project.dependencies].
+# Without these, lowest-direct can resolve packages below the PyPI floor if another dependency also uses it.
+[tool.pixi.feature.lowest-deps.dependencies]
+numpy = ">=2.0"
```
deltamarnix: I understand that this is required. numpy is resolved via a conda dependency first, and only then resolved for a second time in the pip installation step. It gives a wrong image of the situation, because it won't resolve to the lowest version that the pip dependencies depend on as there might be dependencies that make the numpy installation higher than indicated in this toml file.

Would it be possible to remove this and think a little further in our dependency tree? Are we okay with using pip only? Are there some dependencies that can go? Or can we resolve a package like gdal by installing the package only, and skip its dependencies, because we solve those with pip?

Can we move other feature dependencies towards pypi dependencies? Or make those features optional-dependencies in pyproject?

deltamarnix (commented):
I really like this changeset, but since there are still some quirks I wonder if we should include it the way it is now. It seems like we need to solve some problems in our dependency tree. Maybe also test if some of these direct dependencies are still required and if they now have a good version on pypi with wheels.

I am doubting for example if `netcdf4` is still a direct dependency. I don't see it used anywhere in the code. But maybe I'm missing some fix for a transient dependency.

LuukBlom:
closing in favour of https://github.com/Deltares/hydromt/pull/1485

PR #1464, opened 2026-05-18
fix: era5 processing script

### LINKED ISSUE #931: error with ERA5 zarr conversion (?)
### HydroMT version checks

### Reproducible Example

`dc = DataCatalog(data_libs="deltares_data")`

`era5_daily = dc.get_rasterdataset("era5_daily", bbox=[0.259, 6.234, 5.542, 10.378], time_tuple=(datetime.strptime("2022-01-01", '%Y-%m-%d'), datetime.strptime("2022-05-01", '%Y-%m-%d')))`

`era5_daily_zarr = dc.get_rasterdataset("era5_daily_zarr", bbox=[0.259, 6.234, 5.542, 10.378], time_tuple=(datetime.strptime("2022-01-01", '%Y-%m-%d'), datetime.strptime("2022-05-01", '%Y-%m-%d')))`

`era5_daily.sel(longitude=2.25, latitude= 8.75)["precip"].plot()`
`era5_daily_zarr.sel(longitude=2.25, latitude= 8.75)["precip"].plot()`

### Current behaviour

It appears there is an issue with converting ERA5 data to zarr format. The ERA5 zarr data show values exceeding 900 mm/day precipitation for an extended period (see the differences between era5_daily and era5_daily_zarr in the provided code example between mid Feb and March).

### Desired behaviour

make them identical

### Additional context

_No response_

LuukBlom:
## Issue addressed

Fixes #931

## Explanation

There are differences between the era5 data stored on the P drive when comparing `.nc` and `.zarr`.
They are likely produced from running a buggy script, permission errors while writing the enourmous files, chunks not aligning.
Either way, running the `update_zarr` function, and reading from the created datastore now produces the same data as reading from the netcdf sources. 

## Changes 

While waiting on the tests, i resolved some sonarcube errors/warnings, so not all changes are significant.
The main changes are:

- Update glob pattern for source netcdf files. 
- Chunk before writing: `ds.chunk(chunks_out).drop_vars(ds0[v].dims).to_zarr(`
- Zarr v3 updates: encoding / codecs

### DIFF
--- a/data/src/era5_download_resample_convert.py
+++ b/data/src/era5_download_resample_convert.py
@@ -8,11 +8,15 @@ import time
 from os.path import basename, isfile, join
 from typing import Any, Dict, List, Optional
 
+import cdsapi
 import dask
 import numpy as np
 import pandas as pd
 import xarray as xr
+import zarr
+from cftime import date2num
 from dask.diagnostics import ProgressBar
+from zarr.codecs import BloscCodec
 
 _JOULE_PER_M_SQ = "J m**-2"
 
@@ -104,8 +108,6 @@ def download_era5_year(
     days: (list)
         List of days to download data for. Use None to downloads data for all days.
     """
-    import cdsapi
-
     if not isfile(write_path):
         # Set default months, days and hours options
         _months = [f"{m:02d}" for m in range(1, 13)]
@@ -158,7 +160,7 @@ def flatten_era5_temp(
             # while expver1 might contain large odd values
             # https://confluence.ecmwf.int/display/CUSF/ERA5+CDS+requests+which+return+a+mixture+of+ERA5+and+ERA5T+data
             ds_out = ds.sel(expver=5).combine_first(ds.sel(expver=1)).fillna(nodata)
-            chunksizes = tuple([s[0] for s in ds_out.chunks.values()])
+            chunksizes = tuple(s[0] for s in ds_out.chunks.values())
             e0 = {
                 "zlib": True,
                 "dtype": "float32",
@@ -221,7 +223,7 @@ def resample_year(
     dask_kwargs = dask_kwargs or {}
 
     # nc out settings
-    chunksizes = tuple([s for s in chunks.values()])
+    chunksizes = tuple(s for _, s in chunks.items())
     e0 = {
         "zlib": True,
         "dtype": "float32",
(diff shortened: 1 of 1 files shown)


data/src/era5_download_resample_convert.py
```diff
@@ -221,7 +223,7 @@ def resample_year(
     dask_kwargs = dask_kwargs or {}
 
     # nc out settings
-    chunksizes = tuple([s for s in chunks.values()])
+    chunksizes = tuple(s for _, s in chunks.items())
```
deltamarnix, line 226: Looping over the values was fine, right?

data/src/era5_download_resample_convert.py
```diff
@@ -232,14 +234,14 @@ def resample_year(
     paths = []
     # read hourly data for year and year-1!
     for _year in [year - 1, year]:
-        path = join(ddir, var, f"era5_{var:s}_{year:d}_hourly.nc")
+        path = join(ddir, var, f"era5_{var:s}_{_year:d}_hourly.nc")
```
deltamarnix, line 237: This looks like a significant change?

data/src/era5_download_resample_convert.py
```diff
@@ -350,18 +352,18 @@ def append_zarr(
...
-        if compressor is not None:
-            encoding[v] = {"compressor": compressor}
+        encoding[v] = {
+            "compressors": [BloscCodec(cname="zstd", clevel=3, shuffle="shuffle")],
+            "_FillValue": float("nan"),
+        }
+
+    # default chunks: use dataset shape per dimension
+    if not chunks_out:
+        chunks_out = dict(ds.sizes.items())
```
deltamarnix, line 366: Would `chunks_out = ds.sizes.copy()` do the same?

data/src/era5_download_resample_convert.py
```diff
@@ -376,7 +378,7 @@ def append_zarr(
     for v in ds.data_vars:
         # pre-allocate new zarr variables
         if v not in ds0:
-            _da0 = ds0[list(ds0.data_vars.keys())[0]]
+            _da0 = ds0[next(iter(ds0.data_vars.keys()))]
```
deltamarnix, line 381: I don't know what this used to do, but keys are never sorted in a dict, so taking the first item from keys is not always the same. Can you make this more explicit?

deltamarnix (commented):
I can imagine that the chunking is going to do the trick. Some final questions.

data/src/era5_download_resample_convert.py
LuukBlom: Yes this is a bug indeed. It only affects 1st of jan for each year so its not the root cause, but good to fix either way.

data/src/era5_download_resample_convert.py
LuukBlom: True, I couldnt get the sonarcube warnings to disappear. It was complaining about redundant construction of a list and then immediately pass into a tuple. But just removing the list comprehension somehow also produced a different sonarcube warning.

Im okay with reverting tbh

deltamarnix:
Superseded by #1470. We're using earthdatahub for the zarr version of ERA5

PR #1466, opened 2026-05-18
chore: add retry logic to `_fetch_sas_token` to not have to manually retrigger tests

LuukBlom:
see https://github.com/Deltares/hydromt/actions/runs/25658300334/job/75312075790


hydromt/data_catalog/uri_resolvers/azure_blob_resolver.py
```diff
@@ -390,15 +392,17 @@ def _fetch_sas_token(url: str) -> str:
...
-    try:
-        with urllib.request.urlopen(url, timeout=10) as resp:  # noqa: S310
-            data = json.loads(resp.read())
-        return data["token"]
-    except Exception as exc:
-        raise PermissionError(
-            f"AzureBlobResolver: failed to fetch SAS token from '{url}'. "
-            "Check that the URL is correct and reachable."
-        ) from exc
+    for i in range(retries):
```
deltamarnix: We have some retry mechanisms in place. Maybe use pooch instead of requests. We already have that as one of our dependencies.

deltamarnix (changes requested):

hydromt/data_catalog/uri_resolvers/azure_blob_resolver.py
LuukBlom: urllib is a python builtin so we dont add to our dependencies here.
Using `pooch` would be overkill for this use case imo.
It is meant for downloading / caching scientific data files.
I'd say that `urllib` is the perfect fit here, quick and lightweight

deltamarnix (approved):

PR #1479, opened 2026-06-01
Skip longitude normalisation for projected datasets in harmonise_dims (fixes #1476)

### LINKED ISSUE #1476: `harmonise_dims` corrupts projected-CRS coordinates by unconditionally applying longitude normalisation
### HydroMT version checks

### Reproducible Example

```python
import numpy as np
import pandas
import xarray as xr
import hydromt 

from hydromt.data_catalog.drivers.preprocessing import harmonise_dims

x = np.arange(10490, 10490 + 30 * 30, 30, dtype=float)   # 30 m resolution
y = np.arange(39010, 39010 + 30 * 30, 30, dtype=float)[::-1] # all greater than 180 !
t = pd.date_range("2020-01-01", periods=3, freq="D")
data = np.random.rand(len(t), len(y), len(x))

ds = xr.Dataset(
    {"precip": (["time", "y", "x"], data)},
    coords={"x": x, "y": y, "time": t},
)
# No embedded CRS wflow output.nc does not carry one

print("x[0] before:", ds.x.values[0])   # 10490.0

ds_out = harmonise_dims(ds.copy())
print("x[0] after: ", ds_out.x.values[0])  # 10130.0  (shifted −360 m)
```

Output:
```
x[0] before: 10490.0
x[0] after:  10130.0
```

### Current behaviour

[`preprocessing.py` L30](https://github.com/Deltares/hydromt/blob/efec71c99f10eaae0c4d54ac7abe0672d97f501b/hydromt/data_catalog/drivers/preprocessing.py#L30)

applies longitude normalisation unconditionally:

```python
if np.any(lons > 180):
    ds[x_dim] = xr.Variable(x_dim, np.where(lons > 180, lons - 360, lons))
    ds = ds.sortby(x_dim)
```

For a projected-CRS dataset (e.g. SVY21 / EPSG:3414) where x is in metres, all coordinates are > 180. The function subtracts 360 from every x value, shifting the grid 360 m to the west. No warning is raised.

When the shifted dataset is later compared to the model grid x origin 360 m too far west

`identical_grid()` returns `False`, so HydroMT reprojects the forcing incorrectly, because the source CRS is still `None` producing negative or spatially displaced output values.

### Desired behaviour

`harmonise_dims` should skip longitude normalisation when the dataset CRS is projected. The minimal fix is a geographic-CRS check:

```python
_crs = ds.raster.crs
_is_geographic = _crs is None or _crs.is_geographic
if _is_geographic and np.any(lons > 180):
    ds[x_dim] = xr.Variable(x_dim, np.where(lons > 180, lons - 360, lons))
    ds = ds.sortby(x_dim)
```

`_crs is None` is kept in the guard so files without an embedded CRS (like wflow `output.nc`) are not silently corrupted the caller is then responsible for setting `crs` in the catalog entry.

gaoflow:
## Problem

`harmonise_dims` unconditionally wraps 0-360 longitudes to -180-180 for any dataset whose x-coordinates exceed 180:

```python
if np.any(lons > 180):
    ds[x_dim] = xr.Variable(x_dim, np.where(lons > 180, lons - 360, lons))
    ds = ds.sortby(x_dim)
```

For a **projected** dataset whose x is in **metres** (e.g. SVY21 / EPSG:3414, where every x value is > 180), this subtracts 360 from every coordinate, silently shifting the grid 360 m to the west. As the source CRS is still unset, the displaced grid is then reprojected incorrectly, producing spatially shifted output. No warning is raised. (Reported in #1476.)

## Fix

Normalise longitudes only when the dataset has a **geographic** CRS:

- **geographic CRS** → wrap to -180-180 (unchanged behaviour);
- **projected CRS** → leave untouched (x is metres, not longitude);
- **no CRS** → leave unchanged and emit a warning, rather than assuming geographic — the caller should set `crs` in the data catalog entry to opt in.

This follows the direction in the issue discussion (thanks @alimeshgi, @michaelohanrahan): an unknown CRS should not be assumed geographic, since that is the exact scenario that corrupts projected data.

## Reproduction (from #1476)

```python
import numpy as np, pandas as pd, xarray as xr
from hydromt.data_catalog.drivers.preprocessing import harmonise_dims

x = np.arange(10490, 10490 + 30*30, 30, dtype=float)   # 30 m resolution, all > 180
y = np.arange(39010, 39010 + 30*30, 30, dtype=float)[::-1]
t = pd.date_range("2020-01-01", periods=3, freq="D")
ds = xr.Dataset({"precip": (["time","y","x"], np.random.rand(3, len(y), len(x)))},
                coords={"x": x, "y": y, "time": t})

print(ds.x.values[0])                    # 10490.0
print(harmonise_dims(ds.copy()).x.values[0])
# before fix: 10130.0  (shifted -360)
# after fix : 10490.0  (unchanged, with a warning that no CRS is set)
```

## Verification

- New regression tests in `tests/data_catalog/drivers/test_preprocessing.py` cover all three cases (geographic normalises, no-CRS left unchanged + warns, projected-CRS left unchanged). The two bug-targeting tests fail on `main` and pass with this change; the geographic case is unchanged.
- Full `tests/data_catalog/drivers/` suite passes (112 passed, 15 skipped).
- `ruff check` and `ruff format --check` clean.


michaelohanrahan:
Thanks for your attention on this issue @gaoflow

LuukBlom (approved):

PR #1481, opened 2026-06-01
Respect handle_nodata for missing variables in RasterDatasetAdapter (fixes #1407)

gaoflow:
## Problem

`DataCatalog.get_rasterdataset` (via `RasterDatasetAdapter._slice_data`) raised a `NoDataException` whenever a requested variable was missing from the source — regardless of `handle_nodata`. So calling it with `NoDataStrategy.WARN` or `NoDataStrategy.IGNORE` still raised (#1407):

```python
mvars = [var not in ds.data_vars for var in variables]
if any(mvars):
    raise NoDataException(f"RasterDataset: variables not found {mvars}")
```

The other no-data path in the same method (no data left after bbox clipping) already routes through `exec_nodata_strat`, so the two were inconsistent.

## Fix

Route the missing-variables case through `exec_nodata_strat` (the same helper used for bbox clipping), so the behaviour matches `handle_nodata`:

- `RAISE` (the default) → raise `NoDataException` (unchanged);
- `WARN` → log a warning and return `None`;
- `IGNORE` → return `None` silently.

The caller (`RasterDatasetAdapter.transform`) already returns `None` when `_slice_data` returns `None`, consistent with the temporal/spatial slicing branches, so no further propagation changes are needed. The message now lists the missing variable **names** rather than a boolean mask.

## Reproduction (from #1407)

```python
import numpy as np, pandas as pd, xarray as xr
from hydromt.data_catalog.adapters import RasterDatasetAdapter
from hydromt.error import NoDataStrategy

ds = xr.Dataset(
    {"temperature": (["y", "x", "time"], np.random.rand(10, 10, 3)),
     "precipitation": (["y", "x", "time"], np.random.rand(10, 10, 3))},
    coords={"x": np.linspace(0, 1, 10), "y": np.linspace(0, 1, 10),
            "time": pd.date_range("2020-01-01", periods=3)},
)

RasterDatasetAdapter._slice_data(ds, variables=["nonexistent_var"],
                                 handle_nodata=NoDataStrategy.WARN)
# before: raises NoDataException
# after : returns None (and logs a warning)
```

## Verification

- New regression test `test_transform_missing_variables_respects_nodata_strategy` covers all three strategies (RAISE raises; WARN/IGNORE return `None`). It fails on `main` and passes with this change.
- Full `tests/data_catalog/adapters/` suite passes (58 passed).
- `ruff check` and `ruff format --check` clean.


binggao1230:
I checked the remaining red CI job for this branch.

All pytest matrix jobs are passing. In the red `run test coverage and SonarQube scan` job, the test suite itself also passed (`1015 passed, 9 skipped, 77 deselected`). The failure happens afterward during the SonarQube upload because `SONAR_TOKEN` is unavailable/unauthorized in this PR context:

`Not authorized or project not found. Please check the SONAR_TOKEN environment variable...`

I did not push code changes because the remaining failure is CI credentials/configuration rather than this PR's code path.

LuukBlom (approved):

PR #1485, opened 2026-06-02
pixi solve strategy lowest and highest

LuukBlom:
## Explanation

Add lowest dependency solving to the test matrix.
Only use pypi, and no conda packages anymore. This works because we removed GDAL and instead use rio-vrt as dependency.

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist


.github/workflows/docs.yml
```diff
@@ -37,7 +37,10 @@ jobs:
         uses: actions/checkout@v6
       - uses: prefix-dev/setup-pixi@v0.9.6
         with:
-          pixi-version: "v0.59.0"
+          pixi-version: "v0.67.0"
+
+      - name: Install pandoc
```
deltamarnix: We can also add this to `[tool.pixi.feature.docs.dependencies]`, because @LuukBlom said it was a standalone dependency without any other dependencies that will be installed alongside.

LuukBlom: True, that is where it lived before this PR, see the `pyproject.toml` changes.

I removed it since we install it from conda, which we were trying to move away from.
Might as well use pixi global to install it then. 

But can also add it back to not have his manual step, I'm okay with either

pyproject.toml
```diff
@@ -333,72 +330,78 @@ clean-docs-examples = { cmd = ["rm", "-rf", "docs/examples"] }
...
+low-py314 = { features = [
+	"lowest-deps",
+	"py314",
+	"io",
+	"extra",
+	"dev",
+	"test",
+	"doc",
+	"examples",
+] }
```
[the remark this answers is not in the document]
LuukBlom: That is intended. using `solve-group` can make env resolution much faster.
But, essentially pixi pretends environments in the same solve group, ARE the same environment and have the exact same deps, which is obviously not the case here.

If we use the same solve group for the `low` and `full/min`, pixi will force the same package versions across all (or error when trying to resolve).

We must use different solve groups for these envs.

LuukBlom:
`rio-vrt` has only 1 maintainer, and the last commit and release was done 2 years ago. 

I had to introduce a workaround for our tests since `rio-vrt.build_vrt` has a bug, resulting in TypeErrors when the tiles list has only 1 element.
I dont think this will happen often outside of tests, but its a cryptic error that we should handle imo.

There are 2 things we an easily do:

1. have this workaround solution that deduplicates the xml (this passes our tests)
```python
    # Pass duplicated list, then strip the duplicate source from the written XML.
    duplicated = False
    if len(new) == 1:
        new = new + new
        duplicated = True

    build_vrt(vrt_destination_path, new)

    # de-duplicate
    if duplicated:
        tree = ET.parse(vrt_destination_path)
        for band in tree.getroot().findall("VRTRasterBand"):
            source_els = [s for s in band if s.tag.endswith("Source")]
            if len(source_els) > 1:
                band.remove(source_els[-1])
        tree.write(vrt_destination_path, xml_declaration=True, encoding="UTF-8")
```

OR 

2. raise an informative error when there is only 0 or 1 files.
```python
    if len(new) <= 1:
        raise ValueError("Need at least 2 files when creating a vrt")
    build_vrt(vrt_destination_path, new)
```

1 is a bit ugly, but 2 breaks usage when users query a small bbox/region that only intersects 1 tile.

Any opinions @deltamarnix ?

LuukBlom:
I created an issue for the bug in the rio-vrt repo: https://github.com/12rambau/rio-vrt/issues/26

It seems to not be maintained actively anymore, so good to manage expectations.

hydromt/_utils/caching.py
```diff
@@ -171,15 +162,25 @@ def cache_vrt_tiles(
...
+    build_vrt(vrt_destination_path, new)
+
+    # de-duplicate
+    if duplicated:
+        tree = ET.parse(vrt_destination_path)
+        for band in tree.getroot().findall("VRTRasterBand"):
+            source_els = [s for s in band if s.tag.endswith("Source")]
+            if len(source_els) > 1:
+                band.remove(source_els[-1])
+        tree.write(vrt_destination_path, xml_declaration=True, encoding="UTF-8")
```
evetion, line 183: That's terribly... pragmatic.

evetion (approved):

PR #1490, opened 2026-06-02
fix version

LuukBlom:
introduced by the grep not being anchored.
Will not happen in the future


dalmijn (approved):
FInd the manual dispatch a bit dodgy, but its needed for 1.4.0.
Maybe delete it afterwards.
LGTM.

LuukBlom:
sure, valid. 
good to note that it only works for existing tags (which not everyone can create) and pypi will error if the release already exists

PR #1494, opened 2026-06-04
added hydromt_delft3dfm as downstream plugin

### LINKED ISSUE #1493: Add hydromt_delft3dfm as downstream plugin

veenstrajelmer:
## Issue addressed

Fixes #1493

## Explanation

## General Checklist

- [ ] Updated documentation
- [ ] Updated changelog.rst

## Data/Catalog checklist

@LuukBlom I temporarily added main as a trigger to the downstream test workflow, to see if all tests pass (they don't, but they also don't in main). This addition should be removed before merging.

Furthermore, one of the hydromt_delft3dfm tests is failing because of a conflicting hydromt version, because we included a git link in the pyproject.toml as suggested in [the hydromt docs](https://deltares.github.io/hydromt/stable/dev/plugin_dev/testing.html#setting-up-a-github-actions-workflow): 
image
I'd like to keep this git link in our toml, is there a way to work around this?

### DIFF
--- a/.github/workflows/downstream-plugins.yml
+++ b/.github/workflows/downstream-plugins.yml
@@ -43,6 +43,13 @@ jobs:
             pixi_version: v0.65.0
             branch: main
 
+          - name: hydromt-delft3dfm
+            repo: Deltares/hydromt_delft3dfm
+            environment: default
+            pixi_task: test
+            pixi_version: v0.67.1
+            branch: main
+
     steps:
       - name: Checkout hydromt
         uses: actions/checkout@v6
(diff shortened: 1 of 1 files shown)


LuukBlom:
Hi Jelmer, thanks for making the PR.

You are right its strange the docs recommend something that doesnt work.

But, you are essentially running compatibility test with `deps=true` for hydromt_delft3dfm on every PR.
So adding it to the compatibility test is a bit redundant, but better to do twice than not at all.

The problem is that the workflow runs `pixi add "hydromt@file:$wheel_file" --pypi` which adds the wheel to the default feature.
I will try some things to see if we can easily resolve this

LuukBlom (approved):

PR #1500, opened 2026-06-11
fix: explicitly convert integer columns to float

### LINKED ISSUE #723: missing values in reservoir parameters
### HydroMT-Wflow version checks

- [x] I have checked that this bug exists on the latest version of HydroMT-Wflow.

### Reproducible Example

n:\Deltabox\Postbox\Tsiokanos, Athanasios\hydromt_wflow_error_reservoir3\

### Current behaviour

Reservoir parameters such as reservoir_demand, reservoir_max_release, reservoir_initial_depth etc can contain NULL values. It might be worth investigating why this occurs. I ve seen it several times recently. It would also be good to check whether there is actually missing data in our grand data.

Good to note that last year the models for these reservoirs were built with proper values.

### Desired behaviour

Build the model with all parameters or throw a warning/error in case something is missing

### Additional context

_No response_

LuukBlom:
## Issue addressed

Fixes [#723](https://github.com/Deltares/hydromt_wflow/issues/723)

Reproducer, investigation and context: https://github.com/LuukBlom/reservoir-issue-723

## Explanation

The problematic line was: 
`gdf.loc[:, c] = np.where(is_nodata, np.nan, gdf.loc[:, c])`

np.where in this case will always return a float ndarray since np.nan is a float.

In pandas v2, this function used to silently convert to float dtype when setting np.nan.
In pandas v3, that is no longer allowed and they raise on it now.

So now we explicitly set the numeric column dtype to float if required.

The relevant changes are in `_set_nodata`, all others are type hint updates.


hydromt/data_catalog/adapters/geodataframe.py
```diff
@@ -27,11 +28,11 @@ def transform(
         gdf: gpd.GeoDataFrame,
         metadata: "SourceMetadata",
         *,
-        mask: Optional[gpd.GeoDataFrame] = None,
-        variables: Optional[List[str]] = None,
+        mask: gpd.GeoDataFrame | None = None,
```
ClaireDons, line 31: Out of interest for me, why are we changing the Optional arguments?

ClaireDons (approved):
Looks good to me, I ran the issue locally and this seems to fix it.

hydromt/data_catalog/adapters/geodataframe.py
LuukBlom: This is the new recommended python syntax introduced in python 3.11.
Its more readable, and requires fewer imports.
In principle it does the same.

The typing imports are all over the codebase. So, we are slowly updating the whole code base when we touch a file, we also update the type hints.

ClaireDons: Makes sense :)

tests/data_catalog/adapters/test_geodataframe_adapter.py
```diff
@@ -50,3 +52,20 @@ def test_set_crs(
             "GeoDataFrame : CRS from data catalog does not match CRS of"
             " data. The original CRS will be used. Please check your data catalog."
         ) in caplog.text
+
+    def test__set_nodata(self):
```
ClaireDons:
```suggestion
    def test_set_nodata(self):
```

evetion:
For reference, note there's a NA (experimental) for integers as well in Pandas: https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html

LuukBlom:
> For reference, note there's a NA (experimental) for integers as well in Pandas: https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html

Yes I saw this as well, but we do a lot of comparisons where we expect to compare against `np.nan`:

```python
if np.isnan(ds.raster.nodata):
  ...
``` 

That would try to convert pd.NA into a bool, which errors:

```bash
>>> bool(np.isnan(pd.NA))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
    bool(np.isnan(pd.NA))
    ~~~~^^^^^^^^^^^^^^^^^
  File "pandas/_libs/missing.pyx", line 392, in pandas._libs.missing.NAType.__bool__
TypeError: boolean value of NA is ambiguous
```
To not introduce this new behaviour and have to fix it all across the codebase, I decided to make the previous behaviour explicit and just convert to float.

I agree it would be nicer to use this nullable int type, good for an issue!

evetion:
> > For reference, note there's a NA (experimental) for integers as well in Pandas: https://pandas.pydata.org/pandas-docs/stable/user_guide/integer_na.html
> 
> Yes I saw this as well, but we do a lot of comparisons where we expect to compare against `np.nan`:
> 
> ```python
> if np.isnan(ds.raster.nodata):
>   ...
> ```
> 
> That would try to convert pd.NA into a bool, which errors:
> 
> ```shell
> >>> bool(np.isnan(pd.NA))
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>     bool(np.isnan(pd.NA))
>     ~~~~^^^^^^^^^^^^^^^^^
>   File "pandas/_libs/missing.pyx", line 392, in pandas._libs.missing.NAType.__bool__
> TypeError: boolean value of NA is ambiguous
> ```
> 
> To not introduce this new behaviour and have to fix it all across the codebase, I decided to make the previous behaviour explicit and just convert to float.
> 
> I agree it would be nicer to use this nullable int type, good for an issue!

Yeah so that would be (for the future issue):

```python
>>> bool(pd.isna(pd.NA))
True
>>> bool(pd.isna(np.nan))
True
```

LuukBlom:
https://github.com/Deltares/hydromt/issues/1504

PR #1507, opened 2026-06-20
Restrict possible multi-file reading errors.

### LINKED ISSUE #741: Different hydromt-generated precipitation forcings on Linux vs Windows
### HydroMT-Wflow version checks

- [x] I have checked that this bug exists on the latest version of HydroMT-Wflow.

### Reproducible Example

I will provide a model example with the used datacatalogs/config to @DirkEilander and @evetion 

### Current behaviour

For our yearly WMO contribution (wflow_sbm models for the entire Europe + ~3200 models for reservoirs globally) we encountered some differences in the hydromt generated precipitation, in the majority of the modes, when generating them in linux (our cluster and Snellius) and windows. In the linux setup constant block values can appear but also, and most concerning, differences in the values between the two setups. See a figure below for a pixel within a basin.

Image

It is a bit difficult to check so many models but the windows generated appears correct when compared to the raw dataset.

Please investigate the root of this issue. 

### Desired behaviour

Have identical values irrespective of the machine you use (assuming always same input data, software versions, and configuration)

### Additional context

_No response_

evetion:
## Issue addressed

Fixes https://github.com/Deltares/hydromt_wflow/issues/741

## Explanation

While we have not been able to reproduce the issue, we know it is caused by dropped chunks of data (whole years). We prevent this from happening again by introducing the `require_monotonic=True` keyword to resample time. This will raise an error instead of trying to resample a daily timeseries with whole years missing.

Since we don't understand how this corruption is introduced, we code defensively to list all multifile urls in debug mode, hopefully catching missing files (an easy explanation, could be filesystem issues under heavy load). We also introduce a check based on the  multifile resolver that a multifile dataset needs to have the years (or months, or a combination) found from the placeholders urls. So dataset based on `{year}.nc` with `[2021.nc, 2022.nc]` requires a `2021` and `2022` timestamp in the time dimension (if any).

Finally, we introduce an error when users pass `parallel=True` and `lock=False` for a dataset with the default or threaded dask options, as it is guaranteed to cause corruption, as the netcdf library is not thread safe.

## General Checklist

- [x] Updated documentation
- [x] Updated changelog.rst

## Additional Notes

Note that this issue could still occur (just not in `resample_time`), when the bug is in the dask computation. Here we exclude other (easier) possible causes. We could make the time dimension check delayed as well.

Since this is potentially breaking if your datasets contain non-monotonic time periods (unexpectedly), I'm pinging @JoostBuitink @hboisgon @atsiokanos @michaelohanrahan @veenstrajelmer @DirkEilander. This is an FYI, but let me know if this breaks something for you.

PS. The `p:/wflow_global/forcing/MSWEP_V316_test/Past/Yearly/mswep_v316_{year}.nc` dataset contains two missing days in August 1993.
PPS @deltamarnix you made https://github.com/Deltares/hydromt_wflow/pull/698/changes/0d462a7c5b337b144843b27462ffcc35520a3817, do you have a link to the original issue?

### DIFF
--- a/hydromt/data_catalog/drivers/xarray_options.py
+++ b/hydromt/data_catalog/drivers/xarray_options.py
@@ -113,9 +113,7 @@ class XarrayDriverOptions(DriverOptions):
 
     def get_kwargs(self) -> dict[str, Any]:
         """Return attributes set that are not explicitly declared fields."""
-        kwargs = super().model_dump(
-            exclude=self.get_excluded_kwarg_names_for_open(), exclude_unset=True
-        )
+        kwargs = super().get_kwargs()
         if kwargs.get("parallel") is True and kwargs.get("lock") is False:
             scheduler = get_scheduler(collections=[delayed(lambda: None)()])
             if getattr(scheduler, "__module__", None) == "dask.threaded":
--- a/hydromt/data_catalog/drivers/xarray_options.py
+++ b/hydromt/data_catalog/drivers/xarray_options.py
@@ -3,7 +3,10 @@
 import enum
 import logging
 from os.path import splitext
+from typing import Any
 
+from dask import delayed
+from dask.base import get_scheduler
 from pydantic import Field
 
 from hydromt.data_catalog.drivers.base_driver import DriverOptions
@@ -107,3 +110,19 @@ class XarrayDriverOptions(DriverOptions):
             else:
                 filtered_uris.append(_uri)
         return filtered_uris, io_format
+
+    def get_kwargs(self) -> dict[str, Any]:
+        """Return attributes set that are not explicitly declared fields."""
+        kwargs = super().model_dump(
+            exclude=self.get_excluded_kwarg_names_for_open(), exclude_unset=True
+        )
+        if kwargs.get("parallel") is True and kwargs.get("lock") is False:
+            scheduler = get_scheduler(collections=[delayed(lambda: None)()])
+            if getattr(scheduler, "__module__", None) == "dask.threaded":
+                raise ValueError(
+                    "Cannot use xarray open_mfdataset parallel=True with lock=False "
+                    "with a threaded dask scheduler. Use lock=True or a non-threaded "
+                    "scheduler such as distributed."
+                )
+
+        return kwargs
--- a/hydromt/data_catalog/sources/data_source.py
+++ b/hydromt/data_catalog/sources/data_source.py
@@ -9,7 +9,6 @@ from os.path import abspath, join, splitext
 from pathlib import Path, PurePath
 from typing import Any, ClassVar, Optional, TypeVar, Union, cast
 
-import pandas as pd
 from pydantic import (
     BaseModel,
     ConfigDict,
@@ -26,7 +25,7 @@ from hydromt._utils.uris import _is_valid_url
 from hydromt.data_catalog.adapters.data_adapter_base import DataAdapterBase
 from hydromt.data_catalog.drivers import BaseDriver
 from hydromt.data_catalog.uri_resolvers import ConventionResolver, URIResolver
-from hydromt.error import NoDataStrategy, exec_nodata_strat
+from hydromt.error import NoDataStrategy
 from hydromt.typing import DataType, SourceMetadata
 from hydromt.typing.type_def import TimeRange, TotalBounds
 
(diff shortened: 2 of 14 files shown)


hydromt/data_catalog/drivers/xarray_options.py
```diff
@@ -107,3 +110,19 @@ def filter_uris_by_format(
             else:
                 filtered_uris.append(_uri)
         return filtered_uris, io_format
+
+    def get_kwargs(self) -> dict[str, Any]:
+        """Return attributes set that are not explicitly declared fields."""
+        kwargs = super().model_dump(
+            exclude=self.get_excluded_kwarg_names_for_open(), exclude_unset=True
+        )
```
deltamarnix:
```suggestion
        kwargs = super().get_kwargs()
```

Instead of copying what is inside super.get_kwargs, I would propose to call it.

hydromt/data_catalog/sources/data_source.py
```diff
@@ -84,6 +85,59 @@ def _log_start_read_data(self):
...
+                    month=int(d["month"]),
+                    freq="M",
+                )
+                for d in found
+            }
+            # A time coord with 2020-12-15 and 2021-01-15 becomes monthly
+            # periods "2020-12" and "2021-01", matching `expected`.
+            got = set(time.to_period("M"))
+            if missing_months := sorted(expected - got):
+                _handle_missing(f"missing months: {[str(m) for m in missing_months]}")
```
deltamarnix: Maybe also nice to log the year where those months are missing.

hydromt/data_catalog/sources/dataset.py
```diff
@@ -83,6 +84,9 @@ def read_data(
         if ds is None:  # handle_nodata == ignore
             return None
 
+        if isinstance(self.uri_resolver, ConventionResolver):
```
deltamarnix, line 87: Would it be an option to move `_check_multifile_is_complete` into the `ConventionResolver`? Since it is only used for that resolver. I'm not a fan of the `isinstance` checks in these files.

hydromt/data_catalog/uri_resolvers/convention_resolver.py
```diff
@@ -29,9 +32,28 @@ def __missing__(self, key):
...
+    protocol, _ = split_protocol(uri)
+    if protocol is None and os.name == "nt":
+        return PureWindowsPath(uri).as_posix()
+    return uri
+
+
 class ConventionResolver(URIResolver):
     """URIDataResolver using HydroMT naming conventions."""
 
+    _resolved_uri_placeholders: list[dict[str, str]] = PrivateAttr(default_factory=list)
```
deltamarnix, line 56: `_resolved_uri_placeholders` is not a private attribute, it is being used in `DataSource`.

deltamarnix (changes requested):
I would like to see some changes in the way that these functions are called, but I agree with the overall construct of the erroring.

hydromt/data_catalog/sources/dataset.py
evetion: Sure, but you would keep the `isinstance` check, unless we make it a dummy method on all resolvers. The check is nasty that it requires two specialized versions of both the dataset and resolver.

evetion: I've moved it to ConventionResolver but kept the ugly isinstance check.

hydromt/data_catalog/sources/data_source.py
evetion: It does, the logging will be like `missing months: 2020-12`.

hydromt/data_catalog/sources/dataset.py
LuukBlom: I agree with disliking the isinstance check, but I realize its getting nitpicky, so do with this what you want.

Alternative to get rid of it and make the DataSource agnostic of uri resolver: Add a more generic `_post_read_check` function on the `URIResolver` baseclass that does nothing and call it here in all cases. 
In `ConventionResolver`  we then overwrite that function to call `check_multifile_is_complete`.

LuukBlom (approved):
LGTM! nice job

I have 1 nit that you can decide to implement or not.

hydromt/data_catalog/sources/dataset.py
evetion: Yeah, a generic dummy is the way forward. But let's go there once we have other post-process methods.

PR #1509, opened 2026-06-23
update `SONARCLOUD_TOKEN_2025` to the new `SONAR_TOKEN` gh secret

LuukBlom:
The previous secret doesnt exist anymore.

https://github.com/Deltares/hydromt/settings/secrets/actions


LuukBlom:
perhaps the token needs to be regenerated?

deltamarnix:
I checked with @Sonny-Deltares and he says that this PR is doing the reverse. SONARCLOUD_TOKEN_2025 is the correct token. If it is not working, we have to check with FMSST, not change it to the old token.

PR #1511, opened 2026-06-25
skip sonar scan for dependabot and forks

LuukBlom:
## Explanation

SonarScan was broken sometimes due to a missing token.
I think I traced it to the secrets not being available for dependabot or forks.

So I added a guard to conditionally run the scan.

A more generic guard could be this

```yml
if: ${{ secrets.SONAR_TOKEN != '' }}
```

But I also like being explicit, since this will silently skip the scan if the secret name is changed.

### DIFF
--- a/.github/workflows/sonar.yml
+++ b/.github/workflows/sonar.yml
@@ -33,6 +33,9 @@ defaults:
 
 jobs:
   scan:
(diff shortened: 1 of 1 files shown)


LuukBlom:
Closing for cleanup, we decided to not do this

PR #1513, opened 2026-06-25
Sort urls in convention resolver log statements

evetion:
I could add extra code to make sorting perfect (in case of weird keys like `1`,`10`,`2`), but given our usage is mostly years, this is good enough.

### DIFF
--- a/hydromt/data_catalog/uri_resolvers/convention_resolver.py
+++ b/hydromt/data_catalog/uri_resolvers/convention_resolver.py
@@ -205,7 +205,7 @@ class ConventionResolver(URIResolver):
                 self.name,
                 len(uris),
                 uri_expanded,
-                uris,
+                sorted(uris),
             )
 
         return uris


hydromt/data_catalog/uri_resolvers/convention_resolver.py
```diff
@@ -205,7 +205,7 @@ def resolve(
                 self.name,
                 len(uris),
                 uri_expanded,
-                uris,
+                sorted(uris),
             )
```
deltamarnix, line 209: Would be useful to have a better string representation for the uris. The default is `["a.csv", "b.csv"]`, but maybe it's better to have `\n` in between each item?

evetion: I had that, but it blew up the log for larger multi file datasets (you can have up to 40+ newlines from it).

PR #1529, opened 2026-08-11
fix latest docs and version switcher

### LINKED ISSUE #1528: latest docs are not being published
### HydroMT version checks

- [x] I have checked that the issue still exists on the latest versions of the docs on `main` [here](https://github.com/Deltares/hydromt)

### Kind of issue

Docs are missing

_No response_

### Documentation problem

they are being published, but not present in the switcher

_No response_

LuukBlom:
Problem was that some workflows were using `dev` to denote the docs built from main, and others were using `latest`.

## Issue addressed

Fixes #1528

## Explanation

add 'preferred' to switcher.json. 
add concurrency to prevent conflicting gh-pages writes.
fix usage of latest/dev to always use latest 
fix bug in changelog & switcher insertion where location of inserted section would be incorrect.


.github/scripts/record-release-on-main.sh
```diff
@@ -149,32 +156,46 @@ if [ ! -s /tmp/section.rst ]; then
...
-  !inserted && $0 ~ "^"family_base"( |$)" {
-    printf "%s\n", section
-    inserted = 1
-  }
-  !inserted && /^v[0-9]+\.[0-9]+\.[0-9]+/ {
-    # family_base not in changelog; insert before the first version heading.
-    printf "%s\n", section
-    inserted = 1
+  !inserted && $0 ~ /^v[0-9]+\.[0-9]+\.[0-9]+/ {
+    match($0, /^v[0-9]+\.[0-9]+\.[0-9]+/)
```
JoerivanEngelen: There's quite some bash/shell magic going on here. I think this is more readable if you assign $0 to a variable, so I understand what is checked here. Furthermore a comment what is checked in this boolean?

.github/scripts/record-release-on-main.sh
```diff
@@ -149,32 +156,46 @@ if [ ! -s /tmp/section.rst ]; then
...
   { print }
   END {
     if (!inserted) {
-      # No version headings at all; append.
+      # Every existing heading is >= NEW_VERSION, or there are no headings
+      # at all — append at the end.
       printf "\n%s", section
     }
   }
 ' /tmp/main-changelog.rst > /tmp/changelog-merged.rst
```
JoerivanEngelen: It took me some time to figure out this was one awk command running a script on main-changelog.rst ... Wouldn't this be more readable if you would separate this script to a file with a clear name?

JoerivanEngelen (commented):
My knowledge of bash script is a bit limited, so it was hard to follow. I have some suggestions to improve readability somewhat

.github/scripts/record-release-on-main.sh
LuukBlom: I extracted the update changelog and update switcher logic into separate scripts with more usage info, context and comments to make this better. 

awk commands take some getting used to though.

What this does is 
1. loop over all lines ($0 is line)
2. start capturing when the exact version is matched
3. stop capturing when another version is matched
4. echo what was captured (e.g. the new changelog section)

JoerivanEngelen (approved):
This makes things more understandable.

PR #1536, opened 2026-09-03
fix(data_catalog): reading from private bucket

### LINKED ISSUE #1530: HydroMT DataCatalog does not read from private cloud buckets
### HydroMT version checks

### Reproducible Example

I cannot read from the hydromt-data minio bucket. 

Note that a credentials and profile need to be set up as described here: https://github.com/Deltares-research/hydromt_data_pipelines

```python

dc = DataCatalog().from_dict(
    {
        "meta": {
            "version": "v1",
            "hydromt_version": ">1.0a,<2",
        },
        "merit_hydro": {
            "data_type": "RasterDataset",
            "uri": "s3://hydromt-data/merit_hydro/merit_hydro.zarr",
            "driver": {
                "name": "raster_xarray",
                "filesystem": {
                    "protocol": "s3",
                    # see https://s3fs.readthedocs.io/en/latest/ for options
                    "anon": False,
                    "profile": "hydromt-data",
                },
                "options": {"zarr_format": 3},
            },
        },
    }
)
ds = dc.get_rasterdataset("merit_hydro")
```

### Current behaviour

The code above returns a permission error.

This is caused by the current implementation which reduces the filesystem to a uri str at the time it is passed to, in this case, xarray.open_zarr. The profile and credentials information is no longer available and hence xarray tries to read from s3 with default credentials returning the error.

### Desired behaviour

Reads the zarr file.

A `fsspec.mapping.FsMap` object instead of a flat string should be passed to the xarray methods. 
This combines the fs protocal, profile and credentials with the uri.

### Additional context

A prototype implementation using `FsMap` has been made in the `fix/persmission_private_bucket` branch https://github.com/Deltares/hydromt/compare/main...fix/permission_private_bucket

This approach needs to be tested if it also works with other filesystems and implemented in other drivers.
We should also investigate if the `AzureBlobResolver` can be replaced with this approach.

LuukBlom:
Update zarr/xarray reading logic to use fsspec.FSMap for s3 private bucket access, and reduce duplication by extracting it into a single function.

## Issue addressed

Fixes #1530

## Explanation

Took Dirks changes as a starting point.

- Sonarcloud raised an issue that all Xarray-based drivers had a bunch of duplication, so I extracted the duplicated reading logic into a single function: `hydromt.data_catalog.drivers.xarray_options._read_xarray`
- Add handling for private bucket access in readers.py and xarray/zarr drivers.
- Add aws profile configuration to tests.yml
- Add gh secrets for the Access/Secret keys
- Add boto3 and dask[distributed] to deps
- Add `tests/data_catalog/test_private_bucket.py` that reads from the (now still) private hydromt-data minio bucket
- Added `open_zarrs` & `open_mfdataset` to `hydromt.readers`


DirkEilander (changes requested):
thanks for this PR @LuukBlom!

I comfired it is working for zarr and tif, but not yet for nc files. See code snippet below for a test.

It would be good to also add some examples to the docs as there are some specificities that are good for users to know. Like nc only works with the "h5netcdf" engine, not with "netcdf4".

I also tested for vector files where I think another issue (#1542) related to MinIO specific (or different endpoint) seems to be the issue. I made a seperate issue for that one as I can imagine it requires a new PR. 

```py
import s3fs
import xarray as xr

s3_path = "s3://hydromt-data/test/chirps-v2.0.1981.days_p05.nc"

## test with Hydromt DataCatalog
dc = DataCatalog().from_dict(
    {
        "test": {
            "data_type": "RasterDataset",
            "uri": s3_path,
            "driver": {
                "name": "raster_xarray",
                "filesystem": {
                    "protocol": "s3",
                    "anon": False,
                    "profile": "hydromt-data",
                },
                # NOTE: requires h5netcdf engine
                "options": {"engine": "h5netcdf"},
            },
        },
    }
)
try:
    ds = dc.get_rasterdataset("test")
    print(ds)
except Exception as e:
    print(e)
    # > PermissionError: Forbidden


## test with s3fs and xarray (this works)
fs = s3fs.S3FileSystem(anon=False, profile="hydromt-data")
with fs.open(s3_path, mode="rb") as s3_file:
    ds = xr.open_dataset(s3_file, engine="h5netcdf")
    print(ds)
```

docs/user_guide/data_catalog/data_azure_blob_storage.rst
```diff
@@ -0,0 +1,380 @@
...
+     - Azure Blob Resolver
+   * - Public ``abfs://`` container (anonymous)
+     - Yes
+     - Yes
+   * - Private ``abfs://`` container (key/SAS via env vars)
+     - Yes
+     - Yes
+   * - HTTPS blob URLs (``https://<acct>.blob.core.windows.net/…``)
+     - No
+     - **Yes**
```
ClaireDons: Weird nitpick but why are some of the "yes"s bold?

hydromt/data_catalog/drivers/geodataset/vector_driver.py
```diff
@@ -91,6 +92,9 @@ def read(
 
         preprocessor = self.options.get_preprocessor()
         crs: CRS | None = metadata.crs if metadata else None
+        # open_geodataset resolves the data file relative to the location file
+        # on the local filesystem, so this driver is local-only.
+        assert_local(self.filesystem.get_fs(), reader=f"Driver '{self.name}'", uri=uri)
```
ClaireDons: Is this necessary?

